PR #237, opened 2022-05-19
Re-enable GPU tests (in docker)

kwen2501:
Move GPU tests into a CUDA docker container.

Previously the CI failed because the network repo from which we install the CUDA library can be upgrading versions, which could have conflict with the driver installation provided by AWS. 

Moving to docker stabilizes this and save the CUDA library installation.


jamesr66a (approved):
image

PR #239, opened 2022-05-21
T5 DDP: not needing to resolve_pg_per_stage

kwen2501:
Only create c10d PG around the workers (excluding masters).
Then we would be able to rely on PiPPy's internal DDP initialization.


kwen2501:
Stale. Closing.

PR #252, opened 2022-06-08
Add PyTorch version check >= 1.12

### LINKED ISSUE #250: problem with split_module method
hi，
When I use your example code, I found that the split_module has inconsistent number of parameters. How to solve this problem?
This is your example code:
```
import torch
from typing import Any

class MyNetworkBlock(torch.nn.Module):
    def __init__(self, in_dim, out_dim):
        super().__init__()
        self.lin = torch.nn.Linear(in_dim, out_dim)

    def forward(self, x):
        x = self.lin(x)
        x = torch.relu(x)
        return x
   `class MyNetwork(torch.nn.Module):
    def __init__(self, in_dim, layer_dims):
        super().__init__()

        prev_dim = in_dim
        for i, dim in enumerate(layer_dims):
            setattr(self, f'layer{i}', MyNetworkBlock(prev_dim, dim))
            prev_dim = dim

        self.num_layers = len(layer_dims)
        # 10 output classes
        self.output_proj = torch.nn.Linear(layer_dims[-1], 10)

    def forward(self, x):
        for i in range(self.num_layers):
            x = getattr(self, f'layer{i}')(x)

        return self.output_proj(x)

mn = MyNetwork(512, [512, 1024, 256])

from pippy.IR import Pipe

pipe = Pipe.from_tracing(mn)
print(pipe)
print(pipe.split_gm.submod_0)
```
and this error is :
```
Traceback (most recent call last):
  File "test1.py", line 38, in <module>
    pipe = Pipe.from_tracing(mn)
  File "/var/haishin/lab/PiPPy-main/pippy/IR.py", line 833, in from_tracing
    return Pipe._from_traced(mod, traced, multi_use_param_spec, output_loss_value_spec=output_loss_value_spec)
  File "/var/haishin/lab/PiPPy-main/pippy/IR.py", line 604, in _from_traced
    split = split_module(traced, mod, split_callback, qualname_map)
TypeError: split_module() takes 3 positional arguments but 4 were given
```

pbelevich:
Fixes #250

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-torch >= 0.0.0.dev0
+torch >= 1.12.0.dev0
--- a/pippy/IR.py
+++ b/pippy/IR.py
@@ -6,7 +6,17 @@ from typing import Dict, List, Optional, Tuple, Union, cast
 
 import torch
 import torch.fx
+from packaging import version
 from torch.fx.passes.split_module import split_module
+
+# Because split_module with 4 arguments is available only in PT 1.12+
+TORCH_FX_REQUIRED_VERSION = version.parse("1.12")
+
+torch_version = version.parse(torch.__version__)
+assert (torch_version.major, torch_version.minor) >= (
+    TORCH_FX_REQUIRED_VERSION.major,
+    TORCH_FX_REQUIRED_VERSION.minor,
+), "PiPPy requires PyTorch >= 1.12"
 
 # TODO:
 # 1. investigate gradient sync for shared parameters. how does DDP do it?


requirements.txt
```diff
@@ -1 +1 @@
-torch >= 0.0.0.dev0
+torch >= 1.12.0.dev0
```
jamesr66a, line 1: Are we requiring 1.12 or master? `>= 1.12` implies that `1.12` will work, but if we depend on master that wouldn't be true

pbelevich: good question, but this check is better than nothing :) PyPPy for sure doesn't work with < 1.12

kwen2501 (approved):

PR #265, opened 2022-06-20
Replace local_rank with rank

pbelevich:


kwen2501 (approved):
LGTM! Thanks for making this clearer!

PR #266, opened 2022-06-21
Support ViT models in HF test

makslevental:
Support Vision Transformer models, fixes https://github.com/pytorch/PiPPy/issues/244

cc @desertfire

### DIFF
--- a/test/hf_test.py
+++ b/test/hf_test.py
@@ -138,6 +138,18 @@ def roberta_splitter(model) -> int:
             return model.config.num_hidden_layers + 1
 
 
+def vit_splitter(model) -> int:
+    if isinstance(model, ViTModel):
+        for i in range(model.config.num_hidden_layers):
+            annotate_split_points(model, {f'encoder.layer.{i}': PipeSplitWrapper.SplitPoint.BEGINNING})
+        annotate_split_points(model, {'pooler': PipeSplitWrapper.SplitPoint.BEGINNING})
+        return model.config.num_hidden_layers + 2
+    else:
+        for i in range(model.config.num_hidden_layers):
+            annotate_split_points(model, {f'vit.encoder.layer.{i}': PipeSplitWrapper.SplitPoint.BEGINNING})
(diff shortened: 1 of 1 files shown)


test/hf_test.py
```diff
@@ -391,9 +417,8 @@ def test_case(self):
             if model_cls in [SwinForMaskedImageModeling, SwinForImageClassification, SwinModel]:
                 self.skipTest('Need to support SWIN models')
 
-            # TODO: support ViT models https://github.com/pytorch/PiPPy/issues/244
-            if model_cls in [ViTForMaskedImageModeling, ViTForImageClassification, ViTModel]:
-                self.skipTest('Need to support ViT models')
+            if model_cls in [ViTModel]:
```
pbelevich, line 420: Instead of this please add `ViTModel` [here](https://github.com/pytorch/PiPPy/blob/ffa34589bc5e2981b1230d5a4b4625035eca1fb5/test/hf_test.py#L466)

test/hf_test.py
```diff
@@ -369,6 +392,9 @@ def get_output_loss_value_spec_for_model(model_cls):
                      PLBartForSequenceClassification, PegasusForConditionalGeneration, BartForConditionalGeneration]:
         return {'loss': True, 'logits': False, 'encoder_last_hidden_state': False}
 
+    if model_cls in [ViTForMaskedImageModeling, ViTForImageClassification, ViTModel]:
+        return {'loss': True, 'logits': True}
```
pbelevich, line 396: Why 'logits': True? True should be specified for the variable that is supposed to be "loss"
```suggestion
        return {'loss': True, 'logits': False}
```

pbelevich (changes requested):
thanks for working on this!

test/hf_test.py
makslevental: sorry that's a holdover from before i figured out that the right API to get the models to return loss at all. thanks for the catch

pbelevich (approved):
LGTM

PR #274, opened 2022-07-07
Auto-parallelization using DP

makslevental:
WIP implementation of auto-parallelization using Alpa's DP formulation for inter-op latencies.

The high-level algorithm for this implementation of auto-parallelization is

1. Divide the cluster of $N$ hosts with $M$ devices per host into a mesh of *submeshes*, i.e. tuples $(n_s, m_s)$ which correspond to a subset of hosts and devices.
2. Compute intra-op latencies on a per submesh basis i.e., the latency of executing a subset of the ops on each submesh (note this isn't implemented in this PR).
3. Compute the number of pipeline stages and total latency for the optimal mapping of subsets of ops to submesh using dynamic programming.

`#3` is primarily what's implemented in this PR. The DP is defeind by the following recurrence relation

image

When $t_{intra}$ is the intra-op latency of ops $o_k, \dots, o_i$ on submesh $Mesh(n_s, m_s)$.

### DIFF
--- a/pippy/IR.py
+++ b/pippy/IR.py
@@ -2,7 +2,7 @@
 import copy
 import operator
 from enum import Enum
-from typing import Dict, List, Optional, Tuple, Union, cast
+from typing import Callable, cast, Dict, List, Optional, Tuple, Union
 
 import torch
 import torch.fx
@@ -837,7 +837,9 @@ class Pipe(torch.nn.Module):
 
     @staticmethod
     def from_tracing(mod : torch.nn.Module, multi_use_param_spec : Optional[MultiUseParamSpec] = None,
-                     tracer=None, output_loss_value_spec=None, deep_copy_module=False, **kwargs):
+                     tracer=None, output_loss_value_spec=None, deep_copy_module=False,
+                     auto_parallel_strategy : Optional[Callable[[torch.fx.GraphModule], torch.fx.GraphModule]] = None,
+                     **kwargs):
         # TODO: abstract partitioning policy
 
         global _pipeline_tracer
@@ -852,8 +854,10 @@ class Pipe(torch.nn.Module):
         finally:
             _pipeline_tracer = old__pipeline_tracer
 
-        return Pipe._from_traced(mod, traced, multi_use_param_spec, output_loss_value_spec=output_loss_value_spec)
+        if auto_parallel_strategy is not None:
+            traced = auto_parallel_strategy(traced)
 
+        return Pipe._from_traced(mod, traced, multi_use_param_spec, output_loss_value_spec=output_loss_value_spec)
 
     def __str__(self):
         return self.split_gm.__str__()
--- a/pippy/auto_parallelization.py
+++ b/pippy/auto_parallelization.py
@@ -1,0 +1,290 @@
+# Adaptation of https://github.com/alpa-projects/alpa/blob/a88992ce3b46024c0a4ee4aa8cb069a62830cec2/alpa/pipeline_parallel/stage_construction.py
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


jamesr66a (commented):
I know it's a draft but I just scanned through it and left a couple comments

pippy/auto_parallelization.py
```diff
@@ -0,0 +1,190 @@
...
+from enum import Enum
+
+
+class SubmeshSpace(Enum):
+    ALL = "ALL"
+    POWER_OF_TWO = "POWER_OF_TWO"
+    SMALL_POWER_OF_TWO = "SMALL_POWER_OF_TWO"
+
+
+def get_submesh_choices(num_hosts: int, num_devices_per_host: int, space: SubmeshSpace):
```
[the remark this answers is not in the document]
makslevental: Yea I plan to add thorough docs/descriptions/explanations (ie brain dump of what I learned by performing the adaptation).

kwen2501:
Great work!

Some general comments:
- In the PR's description box, can you put a short description of the auto parallelism algorithm?
- Can you add more comments next to the codes?
Thanks!

pippy/IR.py
```diff
@@ -2,7 +2,7 @@
 import copy
 import operator
 from enum import Enum
-from typing import Dict, List, Optional, Tuple, Union, cast
+from typing import Dict, List, Optional, Tuple, Union, cast, Callable, Any
```
jamesr66a, line 5: nit: Do you mind putting these new imports in the proper lexicographical order?

pippy/auto_parallelization.py
```diff
@@ -0,0 +1,257 @@
...
+        while i <= min(n_compute_nodes, 4):
+            submeshes.append((i, n_devices_per_node))
+            i *= 2
+    else:
+        raise ValueError(f"Invalid submesh space: {submesh_space}")
+
+    return submeshes
+
+
+NUMPY_RANDOM_SEED = 42
```
jamesr66a, line 62: I just meant fix the seed in the test but this works too :D

requirements.txt
```diff
@@ -1,2 +1,3 @@
 torch >= 1.12.0.dev0
-packaging >= 21.3
\ No newline at end of file
+packaging >= 21.3
+numba >= 0.55.2
```
jamesr66a, line 3: Hmm I'm not sure if we want to add a top-line dependency to PiPPy for this experimental feature. How does the performance compare without numba?

jamesr66a (commented):
Looks good so far. Please add docs and comments!

requirements.txt
makslevental: Base implementation (i.e. their impl) sees a [250x difference](https://github.com/alpa-projects/alpa/blob/9a215e9d49a5b0fbb6e8f2f4f7cc4ab44ff33009/tests/pipeline_parallel/test_dynamic_programming.py#L49). For mine probably the gap is narrower but still.

jamesr66a: How about we make the dependency optional:

1) Remove it from requirements.txt
2) in `auto_parallelization.py` do something like this:

```
try:
  from numba import njit
  from numba.typed import List as NumbaList
except ImportError:
  def njit(*args, **kwargs):
    pass
    
  NumbaList = list
```

(I haven't run the above code so might be wrong but you should get the idea)

makslevental: this pretty close to what they have ([maybe_numba_jit](https://github.com/alpa-projects/alpa/blob/2c421ee9677d1487253bc056339fdd673faa969d/alpa/util.py#L1135)) so it occurred to me,  but I was thinking we wanted to be perf conscious. anyway i'll add it.

pippy/auto_parallelization.py
```diff
@@ -0,0 +1,265 @@
...
+import torch.fx
+from enum import Enum
+
+from pippy import pipe_split
+
+try:
+    from numba import njit
+    from numba.typed import List as NumbaList
+except ImportError:
+    def njit(*args, **kwargs):
```
makslevental, line 27: decorators that take args have to return a function that [actually does the wrapping](https://stackoverflow.com/a/5929165)

jamesr66a (approved):
LGTM assuming licensing issue is OK

makslevental:
> LGTM assuming licensing issue is OK

All of the ported code is in one file and the attribution/license repro is there as well so we're GTG on that.

PR #285, opened 2022-07-15
Use torch-mlir and MLIR to estimate the latencies of fx graph ops

makslevental:
This PR extends the auto-parallelization API to take an "intra-op latency estimator" and implements one such estimator. The implemented estimator uses MLIR (and potentially LLVM) to count the number of floating point operations in each op and then combines them with a heuristic model of network traffic to compute total intra-op costs. The heurstic model is ingress plus egress bytes (into and out of a subgraph comprising a pipeline stage).

High-level:

1. Compile the model (using its fx graph) through MLIR
2. Unroll all the loops in the model (using either [-affine-loop-unroll](https://mlir.llvm.org/docs/Passes/#-affine-loop-unroll-unroll-affine-loops) or further lowering to LLVM and using [--loop-unroll --unroll-count](https://llvm.org/doxygen/LoopUnrollPass_8cpp.html#a235a94f9bb6716e557601149ca0fb534)).
3. Count up all of the floating point ops and multiply them by their respective latencies (and sum up).
4. Use "prefix sums" to compute both the total latencies for any subgraph and total network traffic.
5. Feed these latencies to the inter-op DP

Running on `resnet18` for a 4x4 configuration you get

```
INFO:root:DP optimal layer submesh assignments [
 ((0, 10), (1, 1), 0),
 ((10, 17), (1, 1), 0),
 ((17, 26), (1, 1), 0),
 ((26, 42), (1, 1), 0),
 ((42, 56), (1, 1), 0),
 ((56, 66), (1, 1), 0),
 ((66, 67), (1, 1), 0),
 ((67, 68), (2, 4), 0),
 ((68, 69), (1, 1), 0)
]
```
which represents the `pipeline stages -> submesh` assignments as `[ith node, jth node), (num hosts, num devices), autosharding config idx` (i.e. non-trivial pipeline stage assignments 🙌✊🥳🎉👏)

This produces the following split for the network:

```
GraphModule(
  (submod_0): GraphModule(
    (conv1): Conv2d(3, 64, kernel_size=(7, 7), stride=(2, 2), padding=(3, 3), bias=False)
    (bn1): BatchNorm2d(64, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (relu): ReLU(inplace=True)
    (maxpool): MaxPool2d(kernel_size=3, stride=2, padding=1, dilation=1, ceil_mode=False)
    (layer1_0_conv1): Conv2d(64, 64, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer1_0_bn1): BatchNorm2d(64, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer1_0_relu): ReLU(inplace=True)
    (layer1_0_conv2): Conv2d(64, 64, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer1_0_bn2): BatchNorm2d(64, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
  )
  (submod_1): GraphModule(
    (layer1_0_relu): ReLU(inplace=True)
    (layer1_1_conv1): Conv2d(64, 64, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer1_1_bn1): BatchNorm2d(64, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer1_1_relu): ReLU(inplace=True)
    (layer1_1_conv2): Conv2d(64, 64, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer1_1_bn2): BatchNorm2d(64, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
  )
  (submod_2): GraphModule(
    (layer1_1_relu): ReLU(inplace=True)
    (layer2_0_conv1): Conv2d(64, 128, kernel_size=(3, 3), stride=(2, 2), padding=(1, 1), bias=False)
    (layer2_0_bn1): BatchNorm2d(128, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer2_0_relu): ReLU(inplace=True)
    (layer2_0_conv2): Conv2d(128, 128, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer2_0_bn2): BatchNorm2d(128, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer2_0_downsample_0): Conv2d(64, 128, kernel_size=(1, 1), stride=(2, 2), bias=False)
    (layer2_0_downsample_1): BatchNorm2d(128, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
  )
  (submod_3): GraphModule(
    (layer2_0_relu): ReLU(inplace=True)
    (layer2_1_conv1): Conv2d(128, 128, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer2_1_bn1): BatchNorm2d(128, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer2_1_relu): ReLU(inplace=True)
    (layer2_1_conv2): Conv2d(128, 128, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer2_1_bn2): BatchNorm2d(128, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer3_0_conv1): Conv2d(128, 256, kernel_size=(3, 3), stride=(2, 2), padding=(1, 1), bias=False)
    (layer3_0_bn1): BatchNorm2d(256, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer3_0_relu): ReLU(inplace=True)
    (layer3_0_conv2): Conv2d(256, 256, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer3_0_bn2): BatchNorm2d(256, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer3_0_downsample_0): Conv2d(128, 256, kernel_size=(1, 1), stride=(2, 2), bias=False)
    (layer3_0_downsample_1): BatchNorm2d(256, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
  )
  (submod_4): GraphModule(
    (layer3_0_relu): ReLU(inplace=True)
    (layer3_1_conv1): Conv2d(256, 256, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer3_1_bn1): BatchNorm2d(256, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer3_1_relu): ReLU(inplace=True)
    (layer3_1_conv2): Conv2d(256, 256, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer3_1_bn2): BatchNorm2d(256, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer4_0_conv1): Conv2d(256, 512, kernel_size=(3, 3), stride=(2, 2), padding=(1, 1), bias=False)
    (layer4_0_bn1): BatchNorm2d(512, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer4_0_relu): ReLU(inplace=True)
    (layer4_0_conv2): Conv2d(512, 512, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer4_0_bn2): BatchNorm2d(512, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer4_0_downsample_0): Conv2d(256, 512, kernel_size=(1, 1), stride=(2, 2), bias=False)
  )
  (submod_5): GraphModule(
    (layer4_0_downsample_1): BatchNorm2d(512, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer4_0_relu): ReLU(inplace=True)
    (layer4_1_conv1): Conv2d(512, 512, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer4_1_bn1): BatchNorm2d(512, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
    (layer4_1_relu): ReLU(inplace=True)
    (layer4_1_conv2): Conv2d(512, 512, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    (layer4_1_bn2): BatchNorm2d(512, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
  )
  (submod_6): GraphModule(
    (avgpool): AdaptiveAvgPool2d(output_size=(1, 1))
  )
  (submod_7): GraphModule()
  (submod_8): GraphModule(
    (fc): Linear(in_features=512, out_features=1000, bias=True)
  )
)
```

One thing to call out: `submod_7` (which is just a `flatten`) gets the largest submesh assignment `(67, 68), (2, 4), 0)`. Counterintuitive but it is as a result of the key "fix": 

```python
                    intra_costs[i, j, k, :] = (
                        lat + ingress_traffic[i, j] + egress_traffic[i, j]
                    ) * np.prod(shape)  # <-------------- the "fix"
```
i.e. there needs to be some penalty for using all of the devices. Thus, the reason `submod_7` gets the most devices is because the floating point latency of that op is lowest (0). Obviously, `np.prod(shape)` (i.e. simply scaling by the number of devices in the submesh) is not the correct model but at least it's proof positive that given some penalty, the algorithm optimizes for *something*. Naturally, we need a better cost model for multi-device.

This PR has gone through considerable revision since last reviewed. Other changes:

1. Further optimization of the DP (sorting valid DP table indices in descending order).
2. Working/correct insertion of `pipe_split` into the fx graph.
3. Refactoring most of the code into a `util` directory such that dependencies can be optional.
4. Parallelization of the compiler using multiprocessing (compile nodes through MLIR in parallel - big speedup over sequential).

I have put inline review comments in the place I would appreciate some more eyes on my implementation of small pieces.


pippy/count_flops.py
```diff
@@ -0,0 +1,160 @@
...
+def find_op_in_submodule(submodule):
+    for node in submodule.graph.nodes:
+        if node.op in {"call_function", "call_module"}:
+            return node.name
+    return None
+
+
+def compile_model_op_by_op(model, example_input):
+    """
+    High-level this
```
jamesr66a, line 67: Maybe add a colon here so it's clear this is starting the numbered list? (I assume that's what this is here for)

pippy/count_flops.py
```diff
@@ -0,0 +1,160 @@
...
+    6. Map the mlir module back to the relevant identifiers (submodule name and aten op name)
+    
+    Returns a list of 3 tuples of (submodule name, aten op name, mlir module);  the mlir module
+    can then be passed to count_flop_latency_in_mlir_module to get a latency number, in terms of floating point ops.
+    """
+    tracer = torch.fx.Tracer()
+    graph = tracer.trace(model)
+    traced = torch.fx.GraphModule(model, graph)
+    module_with_submodules = split_module(traced, model, mod_partition)
+    global PARTITION_COUNTER
```
jamesr66a, line 84: Does this really need to be global? can it just be a closed-over function-local var? Otherwise this will have side effects when invoked multiple times

jamesr66a (commented):
This is really cool. I had some tiny comments. And then also transcribing our discussion from chat:

1) It would be good to have this model both memory and bandwidth-bound operations by calculating arithmetic intensity and applying a different model depending on which regime it lies in
2) It would be really cool if we can generalize this to include communication, i.e. create an automated "generalized roofline" model that can also model scenarios that are network-bandwidth-bound. I believe the fundamentals should be the same

pippy/count_flops.py
makslevental: good call

util/flop_latency.py
```diff
@@ -0,0 +1,440 @@
...
+
+def compute_graph_slice_traffic(nodes, node_shapes):
+    # Collect incoming edges for every slice of the graph in order to
+    # compute ingress and egress data (bytes) further down.
+    cumu_inputs = []
+    for node in nodes:
+        if cumu_inputs:
+            cumu_inputs.append(
+                set((n.name, node.name) for n in node.all_input_nodes) | cumu_inputs[-1]
+            )
```
makslevental, line 331: "prefix sums" for edges, i.e. an increasing collection of in-edges

util/flop_latency.py
```diff
@@ -0,0 +1,440 @@
...
+            cumu_inputs.append(set((n.name, node.name) for n in node.all_input_nodes))
+
+    # Collect outgoing edges for every slice of the graph in order to
+    # compute ingress and egress data (bytes) further down.
+    cumu_outputs = []
+    for node in reversed(nodes):
+        if cumu_outputs:
+            cumu_outputs.append(
+                set((node.name, n.name) for n in node.users) | cumu_outputs[-1]
+            )
```
makslevental, line 342: "prefix sums" for edges, i.e. an increasing collection of out-edges - **note that this loop traverses the node list in reverse**.

util/flop_latency.py
```diff
@@ -0,0 +1,440 @@
...
+            cumu_outputs.append(set((node.name, n.name) for n in node.users))
+    cumu_outputs = cumu_outputs[::-1]
+
+    ingress_traffic = np.zeros((len(nodes), len(nodes)))
+    egress_traffic = np.zeros((len(nodes), len(nodes)))
+    for i in range(len(nodes)):
+        for j in range(i, len(nodes)):
+            input_edges = (cumu_inputs[j] - cumu_inputs[i]) | set(
+                [(n.name, nodes[i].name) for n in nodes[i].all_input_nodes]
+            )
```
makslevental, line 353: net in edges

util/flop_latency.py
```diff
@@ -0,0 +1,440 @@
...
+                [(nodes[j].name, n.name) for n in nodes[j].users]
+            )
+            ingress_edges = input_edges - output_edges
+            egress_edges = output_edges - input_edges
+            ingress_traffic[i, j] = sum(
+                np.prod(node_shapes.get(n, [0])[0]) for n, _ in ingress_edges
+            )
+            egress_traffic[i, j] = sum(
+                np.prod(node_shapes.get(n, [0])[0]) for n, _ in egress_edges
+            )
```
makslevental, line 364: I think this make sense but would appreciate someone double checking my work.

jamesr66a (commented):
just realized i didn't submit my last round of review a long time ago

PR #286, opened 2022-07-15
[spmd] fix distributed_tensor to convert to device type

wanchaol:
device mesh initalized with corresponding device, distribute_tensor
should convert the tensor to device before calling collective


yhcharles (approved):
LGTM

PR #287, opened 2022-07-15
Fix run_until function

kwen2501:
Previously, the `run_until()` function will actually run just one node due to the return command.

### DIFF
--- a/pippy/PipelineDriver.py
+++ b/pippy/PipelineDriver.py
@@ -1184,7 +1184,7 @@ class RemoteInterpreter(torch.fx.Interpreter, EventRecorder):
             if predicate(node):
                 return node
 
-            return self.run_one(node)
+            self.run_one(node)
 
     def run_one(self, node):
         # TODO: hoist run() implementation
--- a/pippy/PipelineDriver.py
+++ b/pippy/PipelineDriver.py
@@ -1177,7 +1177,7 @@ class RemoteInterpreter(torch.fx.Interpreter, EventRecorder):
         else:
             raise AssertionError(f'Unknown operator {torch.typename(target)}')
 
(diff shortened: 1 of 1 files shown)


pippy/PipelineDriver.py
```diff
@@ -1184,7 +1184,7 @@ def run_until(self, predicate: Callable[[torch.fx.Node], bool]):
             if predicate(node):
                 return node
 
-            return self.run_one(node)
+            self.run_one(node)
```
jamesr66a, line 1187: Nit, also add `-> Optional[Node]` return annotation to the `run_until` signature. Maybe also add an `assert False, 'No node returned True for predicate'` after the while loop here

jamesr66a (approved):

pippy/PipelineDriver.py
kwen2501: Actually no call site of `run_until` requires any return -- shall we make this function just return nothing?

jamesr66a: It's theoretically useful for `run_until` to report when it stopped, so I would elect to keep the return

kwen2501: Got it, will add `-> Optional[Node]`.
For the `assert False` call, wouldn't it make things fail here?
https://github.com/pytorch/PiPPy/blob/main/pippy/PipelineDriver.py#L1038

kwen2501: I will probably just add `return None` after the while loop, meaning no more node to run :)

jamesr66a: Sure, I think the semantics here were not actually that clear since the function didn't actually work properly, but it's a fine decision to make it so that the FN runs until the end regardless of whether the predicate is True

PR #288, opened 2022-07-16
[spmd] rewrite redistribute with autograd function

wanchaol:
This PR:
  1. rewrite `redistribute` with autograd function
  2. add _partial to shard resharding
  3. add test cases for all case


spmd/tensor/redistribute.py
```diff
@@ -10,9 +10,9 @@
 
 def redistribute_spmd_tensor(input, device_mesh, placements):
     current_placements = input.placements
+    assert len(placements) == 1, "Only support 1-d placement for now"
+    assert input.device_mesh.mesh.equal(device_mesh.mesh), "cross mesh comm not support yet"
     if isinstance(current_placements[0], Shard) and isinstance(placements[0], Replicate):
```
yhcharles, line 15: Do we want to support `_Partial` -> `Shard` as well?

wanchaol: Just added this :)

yhcharles (approved):
LGTM

PR #289, opened 2022-07-19
[SPMD] Enable black lint for code in `spmd`

yhcharles:
This PR changes the CI config to enable the black linter for `spmd` and `test/spmd` folders only. It also fixes the lint errors.


wanchaol (commented):
Thanks for adding the linter! Can we fix the lint errors in this PR together? Ideally we shouldn't break the CI

yhcharles:
@wanchaol I was planning to fix the lint errors in another PR, as there will be many .py file changes, which might bury this config file change. What do you think?

PR #291, opened 2022-07-20
[spmd] clean up some code

wanchaol:
as titled, also add more detailed instruction to run pytest on specific
test case

### DIFF
--- a/spmd/tensor/ops/math_ops.py
+++ b/spmd/tensor/ops/math_ops.py
@@ -1,7 +1,13 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
 from torch.distributed.distributed_c10d import ReduceOp
 from spmd.tensor.api import Tensor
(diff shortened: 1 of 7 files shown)


spmd/tensor/ops/math_ops.py
```diff
@@ -13,16 +19,14 @@ def dist_sum(self: Tensor) -> Tensor:
 
     local_sum = self_local.sum()
 
-    if isinstance(self_placement, Shard) or isinstance(
-        self_placement, _Partial
-    ):
+    if is_shard(self_placement) or is_partial(self_placement):
```
yhcharles, line 22: Does it make sense to move `is_shard`, `is_partial` to methods of `Placement`?

wanchaol: Yeah that's a good idea, I thought `Placement` is the base type, was wondering if there might be class undefined issue, but I guess I can try it.

yhcharles (approved):

PR #311, opened 2022-07-27
[TP] Enable Dropout for DT and add test for activations

fduwjj:
Enable dropout for DT.


test/spmd/tensor/test_ops.py
```diff
@@ -102,6 +103,43 @@ def test_ones_like(self):
...
+        self._run_sharded_elementwise_ops(device_mesh, [Shard(0)], (8, 5),
+                                          torch.nn.functional.dropout, p=0.4, training=False)
+        self._run_sharded_elementwise_ops(device_mesh, [Shard(1)], (3, 14),
+                                          torch.nn.functional.dropout, p=0.5, training=True)
+
+    @with_comms
+    def test_dropout_errors(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        with self.assertRaisesRegex(RuntimeError, 'Not supported!'):
+            self._run_sharded_elementwise_ops(device_mesh, [_Partial(ReduceOp.SUM)], (8, 5),
```
fduwjj, line 140: @wanchaol Looks like partial tensor does not trigger the override. So no errors are thrown here.

spmd/tensor/ops/dropout.py
```diff
@@ -0,0 +1,20 @@
...
+from typing import Tuple
+from spmd.tensor.ops.utils import register_impl
+from spmd.tensor.api import Tensor
+
+
+@register_impl("aten.native_dropout.default")
+def _dist_dropout(self: Tensor, p: float, train: bool) -> Tuple[Tensor, Tensor]:
+    print(self.placements)
+    self_placement = self.placements[0]
+    if self_placement.is_partial():
```
wanchaol, line 12: let's error out on both partial and replicated for now, and maybe add an TODO here to explain the CPU/GPU difference and fix later

test/spmd/tensor/test_ops.py
```diff
@@ -102,6 +103,36 @@ def test_ones_like(self):
...
+    @with_comms
+    def test_activations(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        self._run_sharded_elementwise_ops(device_mesh, [Shard(0)], (8, 5), torch.nn.functional.gelu)
+        self._run_sharded_elementwise_ops(device_mesh, [Replicate()], (8, 5), torch.nn.functional.gelu)
+        self._run_sharded_elementwise_ops(device_mesh, [Shard(1)], (3, 14), torch.nn.functional.relu)
+        self._run_sharded_elementwise_ops(device_mesh, [Replicate()], (8, 5), torch.nn.functional.relu)
+
+    @with_comms
+    def test_dropout(self):
```
wanchaol, line 129: can we only test this for GPU? also add a TODO here to enable CPU case so we don't forget haha

spmd/tensor/ops/dropout.py
fduwjj: We can still keep this logic since we need this logic for GPU right?

test/spmd/tensor/test_ops.py
fduwjj: Sure.

spmd/tensor/ops/dropout.py
wanchaol: Yep, in GPU it will trigger this function

spmd/tensor/ops/dropout.py
```diff
@@ -0,0 +1,21 @@
...
+from spmd.tensor.ops.utils import register_impl
+from spmd.tensor.api import Tensor
+
+
+@register_impl("aten.native_dropout.default")
+def _dist_dropout(self: Tensor, p: float, train: bool) -> Tuple[Tensor, Tensor]:
+    self_placement = self.placements[0]
+    # TODO: To figure out why partial tensor does not dispatch here when in CPU.
+    # and with kwargs.
+    if self_placement.is_partial():
```
wanchaol, line 13: I think we should error out now for the replicated case, as it's non-deterministic for replicated output. We can enforce the seed generation to be the same for replicated tensor in later PRs

test/spmd/tensor/test_elementwise_ops.py
```diff
@@ -0,0 +1,57 @@
...
+        reset_seed() if reset_seed else None
+        expected = op(input_tensor, **kwargs)
+        self.assertEqual(input_tensor, dist_tensor.local_tensor())
+        self.assertEqual(expected, dt.local_tensor())
+
+    @with_comms
+    @skip_if_lt_x_gpu(TEST_GPU_NUM)
+    def test_activations(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        self._run_sharded_elementwise_ops_gpu(device_mesh, [Shard(0)], (8, 5), torch.nn.functional.gelu)
```
wanchaol, line 31: did we figure out why gelu/relu works magically

wanchaol (approved):
lgtm, two comments.

wanchaol:
nit: linting seems failing on CI

test/spmd/tensor/test_elementwise_ops.py
fduwjj: No idea tbh..

spmd/tensor/ops/dropout.py
fduwjj: Got it, sure.

PR #313, opened 2022-07-28
[spmd] Support N-D DeviceMesh

wanchaol:
This PR adding N-D DeviceMesh support, concretely:

* It underlying creates subgroups according to the n-d mesh array.
* It accepts a user constructed dim based process_groups, so that
  DeviceMesh could fit into existing codebase where subgroups are
  created by hand
* basic support for creating a dist tensor from local torch.Tensor
  with multi-dim device mesh

This should make the initial step towards https://github.com/pytorch/PiPPy/issues/283

The support for operator and redistribute will be done in next few PRs.

### DIFF
--- a/spmd/tensor/api.py
+++ b/spmd/tensor/api.py
@@ -2,13 +2,16 @@
 import copy
 import torch
 from torch.utils._pytree import tree_map
-from typing import Dict, List, Callable, Optional
+from typing import Dict, Callable, Optional, Sequence
 from spmd.tensor.device_mesh import get_global_device_mesh, DeviceMesh
 from spmd.tensor.placement_types import Placement, Shard, Replicate, _Partial
 from spmd.tensor.redistribute import Redistribute
 
 
-class Tensor(torch.Tensor):
+class Tensor(torch.Tensor):  # pyre-ignore[13]: pyre is bad at __new__
+    _local_tensor: torch.Tensor
+    _device_mesh: DeviceMesh
+    _placements: Sequence[Placement]
     __slots__ = ["_local_tensor", "_device_mesh", "_placements"]
 
     # class attribute that handles ops, all handled
@@ -16,21 +19,18 @@ class Tensor(torch.Tensor):
     # pyre-fixme[24]: Generic type `Callable` expects 2 type parameters.
     _dist_tensor_dispatch_ops: Dict[str, Callable] = {}
 
-    # context = contextlib.nullcontext
-
     # pyre-fixme[4]: Attribute must be annotated.
     __torch_function__ = torch._C._disabled_torch_function_impl
 
     @staticmethod
-    # pyre-fixme[3]: Return type must be annotated.
     def __new__(
         cls,
         device_mesh: DeviceMesh,
-        placements: List[Placement],
+        placements: Sequence[Placement],
         size: torch.Size,
         # pyre-fixme[2]: Parameter must be annotated.
         **kwargs,
-    ):
+    ) -> "Tensor":
         # new method instruct wrapper tensor and add placement spec
         # it does not do actual distribution, __init__ should do it instead.
         # TODO: implement __init__ for tensor constructors
@@ -65,14 +65,14 @@ class Tensor(torch.Tensor):
     # pyre-fixme[3]: Return type must be annotated.
     # pyre-fixme[2]: Parameter must be annotated.
     def __torch_dispatch__(cls, func, types, args=(), kwargs=None):
-        # pyre-fixme[3]: Return type must be annotated.
-        # pyre-fixme[2]: Parameter must be annotated.
-        def unwrap_mesh(e):
+        def unwrap_mesh(e: Tensor) -> DeviceMesh:
             # if this tensor is not Distributed, then return none. We will reinterpret it as replicated
             if not isinstance(e, Tensor):
                 return None
             mesh = e.device_mesh
-            assert mesh.ndim == 1, "DistributedTensor ops not supporting multi-dim mesh yet"
+            assert (
+                mesh.ndim == 1
+            ), "DistributedTensor ops not supporting multi-dim mesh yet"
             return mesh
 
         # pyre-fixme[3]: Return type must be annotated.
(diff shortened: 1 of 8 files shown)


fduwjj (commented):
Thanks for sending out the PR so quickly!

test/spmd/tensor/test_device_mesh.py
```diff
@@ -0,0 +1,196 @@
...
+        dim_groups = []
+        dim_expected_ranks = [[[0, 2], [1, 3]], [[0, 1], [2, 3]]]
+        for dim_group_ranks in dim_expected_ranks:
+            per_dim_groups = []
+            for subgroup_ranks in dim_group_ranks:
+                per_dim_groups.append(new_group(ranks=subgroup_ranks))
+            dim_groups.append(per_dim_groups)
+
+        # construct a device mesh from the subgroups
+        mesh = DeviceMesh(self.device_type, [[0, 1], [2, 3]], dim_groups=dim_groups)
```
fduwjj, line 103: Great!!! I think this is what TP needs.

spmd/tensor/device_mesh.py
```diff
@@ -69,61 +95,114 @@ def __init__(self, device_type, mesh):
...
+            # TODO: add more checks to check the correctness
+            # of user passed in dim groups
+            self._dim_to_groups = dim_groups
+            return
+
+        self._dim_to_groups: List[List[ProcessGroup]] = [
+            [] for _ in range(self.mesh.ndim)
+        ]
+        # create sub pgs base on the mesh argument
+        for dim in range(self.mesh.ndim):
```
fduwjj, line 123: So if I understand this correctly, if we have 2D mesh, then we will have 2 sub-process groups right?

spmd/tensor/device_mesh.py
```diff
@@ -69,61 +95,114 @@ def __init__(self, device_type, mesh):
...
+                else:
+                    # smaller than world, create new group
+                    subgroups_by_dim.append(
+                        new_group(ranks=self.mesh, backend=backend_name)
+                    )
+            else:
+                # multi-dim mesh, create subgroups by
+                # looping over the transposed stacked_mesh
+                # and append the groups
+                for dim_mesh in stacked_mesh.T:
```
fduwjj, line 145: Why is T needed here?

For your test example, if the mesh is like:
torch.arange(8).reshape(2, 2, 2)
tensor([[[0, 1],
         [2, 3]],
        [[4, 5],
         [6, 7]]])
So stacked_mesh will be
tensor([[0, 1, 2, 3],
        [4, 5, 6, 7]])

And stacked_mesh.T becomes:
tensor([[0, 4],
        [1, 5],
        [2, 6],
        [3, 7]])

Any reason why we use T not the original tensor to create the sub-process group? I personally think using the original tensor makes more sense. But this is not a blocker just out of curiosity.

fduwjj (approved):
Unblock. For the n-d subprocess part, I don't have a strong opinion.

spmd/tensor/api.py
```diff
@@ -72,7 +71,9 @@ def unwrap_mesh(e):
             # if this tensor is not Distributed, then return none. We will reinterpret it as replicated
```
yhcharles, line 71: How about adding type annotation and fix pyre warnings when we update existing code?

spmd/tensor/device_mesh.py
wanchaol: The reason we have this transpose is that this works for 1d and 2d to generate corresponding submeshes, and I am relying on induction rules to ensure n-d works. 

For 3d, it's a bit hard to visualize, we can think of 3d as a cube that have the 3rd dimension, given the 3rd dim size = 2, it should be something like 

([[0, 4],
[1, 5],
[2, 6],
[3, 7]])

If we only look over through the 3rd dimension. 

For more than 3d, it's hard and I am purely relying on induction.

spmd/tensor/api.py
```diff
@@ -114,25 +115,19 @@ def wrap(e, mesh, placements):
...
-        device_mesh=None,
-        # pyre-fixme[2]: Parameter must be annotated.
-        placements=None,
-        # pyre-fixme[2]: Parameter must be annotated.
-        run_check=True,
-    ):
+        local_tensor: torch.Tensor,
+        device_mesh: DeviceMesh = None,
+        placements: List[Placement] = None,
+        run_check: bool = True,
```
aazzolini, line 123: run_check is still not implemented, right?

spmd/tensor/device_mesh.py
```diff
@@ -16,48 +20,70 @@
...
+    DeviceMesh represents a mesh of devices, where layout of
+    devices could be represented as a n-d dimension array, and
+    each value of the n-d dimensional array is the global id
+    of the default process group ranks.
+
+    DeviceMesh could be used to describe the layout of devices
+    across the cluster, and serves as a proxy for communication
+    among the device lists within the cluster.
 
     We use the default ProcessGroup in this DeviceMesh class
```
aazzolini, line 47: should we allow to pass a "parent" process group?

test/spmd/tensor/test_device_mesh.py
```diff
@@ -0,0 +1,196 @@
...
+    @with_comms
+    def test_device_mesh_context_manager(self):
+        with DeviceMesh(self.device_type, list(range(self.world_size))) as mesh:
+            shard_spec = [Shard(0)]
+            local_tensor = torch.randn(3, 3)
+            sharded_tensor = Tensor.from_local(
+                local_tensor, device_mesh=mesh, placements=shard_spec
+            )
+
+        with DeviceMesh(self.device_type, list(range(self.world_size))):
```
aazzolini, line 41: could we add a test for when a tensor is sharded on the same mesh dimension on 2 tensor dimensions [Shard(0), Shard(0)] and check that the size is correctly computed?

spmd/tensor/device_mesh.py
aazzolini: Would the following work? 
ranks_per_mesh = mesh.swapdims(-1, dim).reshape(-1, mesh.shape(dim))
for dim_mesh in ranks_per_mesh:
    new_group(..)

swapdims would place the current dim as the inner-most dimension of the tensor, and then reshape would flatten all other dimensions out.

Also, I think that new_group will only return a valid process group for one of the calls in this loop, since each rank will belong to only one process group per mesh dimension, so maybe we can check for that and make subgroups_by_dim a List[PG] instead of List[List[PG]] (maybe call them local_subgroup_by_dim)

test/spmd/tensor/test_device_mesh.py
aazzolini: In case we have a [Shard(0), Shard(0)] tensor, how are we going to e.g. perform things like all_gather? Would we have to do all_gather on one dimension of the mesh first, then all_gather on the other dimension? Or would it be possible to do a single all_gather on both? In that case I think we'd have a sub process group that covers 2 dimensions of the mesh ?

spmd/tensor/device_mesh.py
wanchaol: Yeah that looks better, especially the single pg per dim could help us simplify things, will update, thanks for suggestion!

spmd/tensor/device_mesh.py
wanchaol: We should probably always use default pg as parent group, there's bunch of reasons:
1. DeviceMesh is using logical device id in meshes, which is the default pg rank number, if we allow passing a parent group and use the parent group as id assignment, the ids between device meshes gonna be different and it's hard to tell the mapping
2. Currently in c10d now we are relying on some global id from default pg to do some communication, i.e. rank related c10d op like scatter, are using global rank as the src_rank, so we better keep the default pg with the global rank to do such operations.

We should allow people to slice the device mesh and get a sub device_mesh, which basically user could get the smaller sized group as they want, or user could construct a device mesh with only subset of ranks, to keep the logical id meaning in the mesh

spmd/tensor/api.py
wanchaol: Yeah sure let me add a docstr, `run_check` is actually partially implemented by broadcasting for replicated tensor, we still need to check additional attributes like tensor sizes, dtype, etc.

PR #314, opened 2022-07-29
[SPMD] Implement operator admm

yhcharles:


spmd/tensor/ops/matrix_ops.py
```diff
@@ -16,17 +16,18 @@ def dist_addmm(
...
     mat2: Tensor,
-    # pyre-fixme[2]: Parameter must be annotated.
-    beta=1,
-    # pyre-fixme[2]: Parameter must be annotated.
-    alpha=1,
+    beta: float = 1,
+    alpha: float = 1,
 ) -> Tensor:
     # dist addmm:
+    # input:replicate   mat1: replicate,  mat2: replicate
```
wanchaol, line 23: this should not goes into this function, and instead handled by the default handler inside torch_dispatch?

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -1,31 +1,84 @@
...
+)
 
 
 class DistMatrixOpsTest(DistTensorTestBase):
     @with_comms
     def test_addmm(self):
-        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
-        shard_spec = [Shard(0)]
-        replica_spec = [Replicate()]
+        from torch.distributed import get_rank
```
wanchaol, line 19: we can use `self.rank` to directly get the rank

wanchaol (approved):
looks good to me! Some minor comments

spmd/tensor/ops/matrix_ops.py
```diff
@@ -35,31 +36,37 @@ def dist_addmm(
...
+            and mat2_placement.is_shard(dim=0)
         )
-    else:
-        raise RuntimeError(
-            f"addmm operator supported for inputs: {mat1}, {mat2}"
+    ):
+        local_res = local_input.addmm(
+            local_mat1, local_mat2, beta=beta, alpha=alpha
         )
+        return Tensor.from_local(local_res, device_mesh, input.placements)
```
wanchaol, line 64: Took a closer look, there're some issues here, we cannot simply assume the output placement is the same in every case, in fact, in multiple cases, they are different, we need to specify the output placement case by case.

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -1,31 +1,84 @@
...
+        device_mesh = DeviceMesh(self.device_type, list(range(world_size)))
+        placement_rep = [Replicate()]
+        placement_par = [_Partial(ReduceOp.SUM)]
+        placement_s0 = [Shard(0)]
+        placement_s1 = [Shard(1)]
+
+        # Ground truth
+        torch.manual_seed(0)
+        input = torch.randn((world_size, world_size))
+        mat1 = torch.randn((world_size, world_size * 2))
```
wanchaol, line 33: also as Junjie suggested, we need to directly test on `nn.Linear` as well in additional to addmm

spmd/tensor/ops/matrix_ops.py
wanchaol: as a reference, please see P519128154 for a correct output placement specification

wanchaol (changes requested):
seems having some issue after testing with linear, requesting changes for output placement and need more test in nn.Linear.

yhcharles:
Drop this PR as a new solution is merged in main branch.

PR #319, opened 2022-08-03
[spmd] enable distributed tensor work with torch.Tensor in autograd

wanchaol:
This PR enables distributed tensor work together with torch.Tensor
in both forward and backward, by:
* make `from_local` and `to_local` to always call autograd function, this hooks up the graph of torch.Tensor and tensor subclasses in a way that's not intrusive to the autograd engine itself
* make `DTensor.__new__` accept local_tensor and allow construction like `DTensor(local_tensor, ..)` directly, this is never differentiable and we always get a leaf tensor.  
* with this change,now dist tensor could generate grads, tensor could generate grads when appropriate after the backward pass
* bunch of bug fixes coming together with this PR.

This enable us to plug in DistributedTensor as part of the module hierarchy, instead of urging all tensors converting to DistributedTensor

### DIFF
--- a/spmd/__init__.py
+++ b/spmd/__init__.py
@@ -60,7 +60,7 @@ def distribute_tensor(
             tensor_list = list(tensor.chunk(num_chunks, dim=shard_dim))
             scatter_shape = list(tensor.size())
             scatter_shape[shard_dim] = chunk_size
-            local_tensor = device_mesh.scatter(tensor_list)
+            local_tensor = device_mesh.scatter(tensor_list, mesh_dim=idx)
             dist_tensor = DTensor(
                 local_tensor,
                 device_mesh,
--- a/spmd/__init__.py
+++ b/spmd/__init__.py
@@ -46,13 +46,13 @@ def distribute_tensor(
             scatter_shape = list(tensor.size())
             scatter_shape[shard_dim] = chunk_size
             local_tensor = device_mesh.scatter(tensor_list)
-            dist_tensor = Tensor.from_local(
+            dist_tensor = Tensor(
                 local_tensor, device_mesh, placements
             )
         elif isinstance(placement, Replicate) or isinstance(
             placement, _Partial
         ):
-            dist_tensor = Tensor.from_local(tensor, device_mesh, placements)
+            dist_tensor = Tensor(tensor, device_mesh, placements)
         else:
             raise RuntimeError("Not supported!")
 
--- a/spmd/__init__.py
+++ b/spmd/__init__.py
@@ -1,5 +1,5 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
-from typing import List
+from typing import Sequence, Optional
 import torch
 import torch.nn as nn
 from spmd.tensor import Tensor, Placement, Shard, Replicate, _Partial
@@ -8,14 +8,11 @@ from spmd.tensor.device_mesh import get_global_device_mesh, DeviceMesh
 torch.__future__.set_overwrite_module_params_on_conversion(True)
 
 
-# pyre-fixme[3]: Return type must be annotated.
 def distribute_tensor(
     tensor: torch.Tensor,
-    # pyre-fixme[9]: device_mesh has type `DeviceMesh`; used as `None`.
-    device_mesh: DeviceMesh = None,
-    # pyre-fixme[9]: placements has type `List[Placement]`; used as `None`.
-    placements: List[Placement] = None,
-):
+    device_mesh: Optional[DeviceMesh] = None,
+    placements: Optional[Sequence[Placement]] = None,
+) -> Tensor:
     # get default device mesh if there's nothing specified
     device_mesh = (
         get_global_device_mesh() if device_mesh is None else device_mesh
@@ -49,9 +46,7 @@ def distribute_tensor(
             dist_tensor = Tensor(
                 local_tensor, device_mesh, placements
             )
-        elif isinstance(placement, Replicate) or isinstance(
-            placement, _Partial
-        ):
+        elif isinstance(placement, Replicate):
             dist_tensor = Tensor(tensor, device_mesh, placements)
         else:
             raise RuntimeError("Not supported!")
@@ -63,10 +58,8 @@ def distribute_tensor(
 # pyre-fixme[3]: Return type must be annotated.
 def distribute_module(
     mod: nn.Module,
-    # pyre-fixme[9]: device_mesh has type `DeviceMesh`; used as `None`.
-    device_mesh: DeviceMesh = None,
-    # pyre-fixme[9]: spec has type `List[Placement]`; used as `None`.
-    spec: List[Placement] = None,
+    device_mesh: Optional[DeviceMesh] = None,
+    spec: Optional[Sequence[Placement]] = None,
 ):
     """
     this function coverts all module parameters
--- a/spmd/__init__.py
+++ b/spmd/__init__.py
@@ -43,9 +43,19 @@ def distribute_tensor(
             scatter_shape = list(tensor.size())
             scatter_shape[shard_dim] = chunk_size
             local_tensor = device_mesh.scatter(tensor_list)
-            dist_tensor = Tensor(local_tensor, device_mesh, placements)
+            dist_tensor = Tensor(
+                local_tensor,
+                device_mesh,
+                placements,
+                requires_grad=local_tensor.requires_grad,
+            )
         elif isinstance(placement, Replicate):
-            dist_tensor = Tensor(tensor, device_mesh, placements)
+            dist_tensor = Tensor(
+                tensor,
+                device_mesh,
+                placements,
+                requires_grad=tensor.requires_grad,
+            )
         else:
             raise RuntimeError("Not supported!")
 
--- a/spmd/__init__.py
+++ b/spmd/__init__.py
@@ -13,6 +13,24 @@ def distribute_tensor(
     device_mesh: Optional[DeviceMesh] = None,
     placements: Optional[Sequence[Placement]] = None,
 ) -> Tensor:
+    """
+    Distribute a torch.Tensor to the `device_mesh` according to the `placements`
+    specified. The rank of `device_mesh` and `placements` must be the same.
+
+    Args:
+        tensor (torch.Tensor): torch.Tensor to be distributed
+        device_mesh (:class:`DeviceMesh`, optional): DeviceMesh to distribute the
+            tensor, if not specified, must be called under a DeviceMesh context
+            manager, default: None
+        placements (List[:class:`Placement`], optional): the placements that
+            describes how to place the tensor on DeviceMesh, must have the same
+            number of elements as `device_mesh.ndim`. If not specified, we will
+            by default replicate the tensor across the `device_mesh` from the
+            first rank of each dimension of the `device_mesh`.
+
+    Returns:
+        A :class:`spmd.Tensor` object
+    """
     # get default device mesh if there's nothing specified
     device_mesh = (
         get_global_device_mesh() if device_mesh is None else device_mesh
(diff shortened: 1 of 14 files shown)


spmd/tensor/api.py
```diff
@@ -12,6 +12,62 @@
...
+        # so that the gradient layout matches, and we could return
+        # local gradients directly
+        if grad_output.placements != previous_placement:
+            grad_output = grad_output.redistribute(
+                previous_device_mesh, previous_placement
+            )
+        return grad_output._local_tensor, None, None
+
 
 class Tensor(torch.Tensor):  # pyre-ignore[13]: pyre is bad at __new__
```
albanD, line 72: Why not `DistributedTensor`? That makes the code pretty hard to read in this file :/
Also it makes the typing very hard to follow as everything is typed "Tensor" !

spmd/tensor/api.py
```diff
@@ -12,6 +12,62 @@
...
+# As an example, it enables backward computation of the following:
+#   Module A (with all torch.Tensor params)
+#       -> Module B (with all DistributedTensor params)
+#           -> Module C (with all torch.Tensor params)
+class ToTorchTensor(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, input: "Tensor"):  # type: ignore
+        ctx.previous_placement = input.placements
+        ctx.previous_device_mesh = input.device_mesh
+        return input._local_tensor
```
albanD, line 32: The returned object here should be a "fresh" Tensor. You should add an op (.detach or .view_as) here to prevent any weird stuff.

spmd/tensor/api.py
```diff
@@ -12,6 +12,62 @@
...
+        previous_device_mesh = ctx.previous_device_mesh
+
+        # reshard to the placement when creating DistributedTensor
+        # so that the gradient layout matches, and we could return
+        # local gradients directly
+        if grad_output.placements != previous_placement:
+            grad_output = grad_output.redistribute(
+                previous_device_mesh, previous_placement
+            )
+        return grad_output._local_tensor, None, None
```
albanD, line 69: This should be `grad_output.local_tensor()`. Keep in mind that the backward should be differentiable!

spmd/tensor/api.py
```diff
@@ -30,41 +86,53 @@ class Tensor(torch.Tensor):  # pyre-ignore[13]: pyre is bad at __new__
...
-            dtype=dtype,
-            device=device,
-            layout=layout,
-            requires_grad=requires_grad,
+            torch.Size(tensor_shape),
+            strides=local_tensor.stride(),
+            dtype=local_tensor.dtype,
+            device=local_tensor.device,
+            layout=local_tensor.layout,
+            requires_grad=local_tensor.requires_grad,
```
albanD, line 119: This should be `kwargs.get("requires_grad")`
```python
>>> a = torch.rand(2, requires_grad=True)
>>> torch.tensor(a)
<stdin>:1: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
tensor([0.1073, 0.9690])
>>> torch.tensor(a, requires_grad=True)
tensor([0.1073, 0.9690], requires_grad=True)
>>> torch.tensor(a.detach(), requires_grad=True)
tensor([0.1073, 0.9690], requires_grad=True)

```

spmd/tensor/api.py
```diff
@@ -30,41 +86,53 @@ class Tensor(torch.Tensor):  # pyre-ignore[13]: pyre is bad at __new__
...
+        # the wrapper tensor directly instead of local torch.Tensor
+        r._local_tensor = local_tensor.detach()
         return r
 
     # pyre-fixme[14]: `__repr__` overrides method defined in `Tensor` inconsistently.
     # pyre-fixme[3]: Return type must be annotated.
     def __repr__(self):
         # TODO: consider all_gather the local tensors for better debugging
-        return f"DistributedTensor({self._local_tensor}, placements={self._placements})"
+        return f"DistributedTensor(local_tensor={self._local_tensor}, placements={self._placements})"
```
albanD, line 135: No printing of the device mesh?

spmd/tensor/api.py
```diff
@@ -176,44 +247,22 @@ def from_local(
...
-            elif isinstance(placement, Replicate):
-                if run_check:
+        if run_check:
+            # TODO: by default check tensor metas across rank, and see if we need
+            # to make the run_check inside autograd function
+            for idx, placement in enumerate(placements):
+                if placement.is_replicate():
                     # broadcast rank 0 tensor to all ranks
                     # only broadcast if run_check is True
                     local_tensor = local_tensor.contiguous()
```
albanD, line 257: Is this contiguous call ok?
Note that this means that you will have a `clone` in the autograd graph if the Tensor was not differentiable.

If you don't want that clone in the graph, then yes, you can move this logic to the custom Function

spmd/tensor/api.py
```diff
@@ -30,41 +86,53 @@ class Tensor(torch.Tensor):  # pyre-ignore[13]: pyre is bad at __new__
     @staticmethod
     def __new__(
         cls,
+        local_tensor: torch.Tensor,
```
fduwjj, line 89: What if users want to pass in a global tenor and convert it to a sharded DT. So the only way is to use distribute_tensor?

spmd/tensor/api.py
```diff
@@ -236,7 +285,7 @@ def redistribute(
         return Redistribute.apply(self, device_mesh, placements)
 
     def local_tensor(self) -> torch.Tensor:
-        return self._local_tensor  # type: ignore
+        return ToTorchTensor.apply(self)  # pyre-ignore[16]: autograd func
```
fduwjj, line 288: Thanks for making this happening. Can we move the dt.grad.local_tensor to dt.local_tensor().grad? So that FSDP + TP can work. Or we need to create an ad-hoc backward hook?

spmd/tensor/api.py
wanchaol: At some point, we probably move the whole DistributedTensor implementation to core torch.distributed, and there was a discussion on whether we should call it a `DistributedTensor` or other names. One argument is that we don't want the naming ended be `torch.distributed.DistributedTensor`, as there're duplicate "distributed" there, `torch.distributed.Tensor` looks better, and that's the reason we are having it as "Tensor" now.  Although I agree it's a bit confusing..

Any suggestions on naming? Also cc @aazzolini @yhcharles @wangkuiyi @jamesr66a @mrshenli on bettering naming of DistributedTensor. 

Or we can do as follows: This tensor subclass definition we switch to use `DistributedTensor`. But when we expose it in package init (i.e. in torch.distributed.__init__.py), we can do things like following:

`from spmd.tensor.api import DistributedTensor as Tensor`

This will make an name alias for DistributedTensor, and still appear from the package level as `torch.distributed.Tensor`

fduwjj (commented):
Thanks so much for making this happening so fast.

spmd/tensor/api.py
wanchaol: yeah that's what the `distributed_tensor` API helps, because it needs additional logic on chunking it to shards and sending it over to other ranks.

spmd/tensor/api.py
wanchaol: good catch, adding

spmd/tensor/api.py
aazzolini: Strongly agree re. code readability issues. I am supportive of renaming to DistributedTensor and if it's clear from the context user can always alias it.

spmd/tensor/api.py
fduwjj: I see. It would be nice if we can mention this in the comment or at least add a TODO for it.

spmd/tensor/api.py
albanD: If you do `dt = DistributedTensor.from_local(foo)` then the gradients will be populated on `foo.grad`.
Neither `dt.grad` nor `dt.local_tensor()` will have a .grad field.

albanD (commented):
Sounds quite good! Only a small detail on stride there that could be a follow up

fduwjj (commented):
Btw I found more places where from_local is called. https://github.com/pytorch/PiPPy/blob/main/test/spmd/tensor/test_tensor_ops.py#L26.

[Edit]: NVM, looks like you are not removing the API, sorry for the confusion.

spmd/tensor/api.py
fduwjj: I see. Sounds like this is what we want for TP + FSDP. Let me try it out to verify that. Thanks!

wanchaol: Yeah, that means we probably need to invent a API in TP + FSDP integration, called `distributed_fsdp_param`, where it use `from_local` instead of the plain `Tensor` constructor. 

The `distribute_tensor` API likely won't use from_local for this case as module parameters should be dist param

spmd/tensor/api.py
soumith: `DTensor` maybe?

spmd/__init__.py
```diff
@@ -8,14 +8,11 @@
...
-# pyre-fixme[3]: Return type must be annotated.
 def distribute_tensor(
     tensor: torch.Tensor,
-    # pyre-fixme[9]: device_mesh has type `DeviceMesh`; used as `None`.
-    device_mesh: DeviceMesh = None,
-    # pyre-fixme[9]: placements has type `List[Placement]`; used as `None`.
-    placements: List[Placement] = None,
-):
+    device_mesh: Optional[DeviceMesh] = None,
+    placements: Optional[Sequence[Placement]] = None,
```
aazzolini, line 14: could you add a docstring mentioning the default behavior for placements?

spmd/tensor/api.py
```diff
@@ -14,6 +15,79 @@
...
+# NOTE [Autograd interaction between torch.Tensor]
+#
+# The autograd functions defined below are being used by the public
+# facing APIs (i.e. from_local, local_tensor) to ensure our Tensor
+# works together with torch.Tensor within autograd engine. This
+# allows DistributedTensor to exist on part of the module hierarchy
+# and still able to calculate gradients across the torch.Tensor and
+# DistributedTensor boundary.
+# As an example, it enables backward computation of the following:
+#   Module A (with all torch.Tensor params)
```
aazzolini, line 27: Would it be possible to use more concrete examples here? In the general case I feel like the conversion from local to distributed could lead to surprising behavior in the backward pass depending on the use case. 

For example --
1) if the user wraps a DDP _activation_ with DistributedTensor  , they should pass [Shard(0)] into the constructor. Then in the backward pass, the inverse operation would guarantee that we get a gradient that is also [Shard(0)]
2) however, if the user wraps a DDP _parameter_ with DistributedTensor, then they should pass [Replicate] since all ranks should have the same parameter. In the backward pass, we would potentially end up with a [Partial_] gradient, which we would then convert to Replicate -- but this is wrong, since DDP will later do an all-reduce again on an already replicated parameter.

spmd/tensor/api.py
wanchaol: > `DTensor` maybe?

Yeah that's actually a good one! Short enough to distinguish with torch.Tensor, and even if we have `torch.distributed.DTensor` it looks still reasonable! This also seems consistent with [TF's DTensor](https://www.tensorflow.org/guide/dtensor_overview) and easy to map between each other. 

I am down with `DTensor` :)

wanchaol: It's happening here https://github.com/pytorch/PiPPy/pull/330

spmd/tensor/api.py
```diff
@@ -29,43 +102,72 @@ class DTensor(torch.Tensor):  # pyre-ignore[13]: pyre is bad at __new__
...
+                shard_dim = placement.dim
+                local_dim_size = tensor_shape[shard_dim]
+                # recover tensor shape on the shard dim
+                tensor_shape[shard_dim] = local_dim_size * device_mesh.size(idx)
+
+                # recover tensor stride by modifying the stride that larger than
+                # the current stride on the shard_dim
+                for i in range(len(tensor_stride)):
+                    if (
+                        i < shard_dim
```
albanD, line 128: You don't want this condition.
A larger dimension might have been transposed after the shard_dim and you still want to update it!

fduwjj (approved):

PR #320, opened 2022-08-03
Add strict mode for DT and more checks and error …

aazzolini:
…messages

- Allow DT to fail if no implementation for specific op is present.
- Fail with good error message when mixing distributed and vanilla tensors.
- Improve error message for when sharding would result in un-even sizes.


wanchaol (approved):
looks great! Thanks for adding

PR #321, opened 2022-08-04
[spmd] separate sharding propagation with compute

wanchaol:
This PR separates the computation and sharding propagation, so that we could enable easier/quick operator implementation. 

TODO:
add more point-wise op tests, and find a way to map between the original op impl with the aten op impl (i.e. map `torch.add` to `torch.ops.aten.add.default`, so that we could do testing easily.


spmd/tensor/ops/pointwise_ops.py
```diff
@@ -0,0 +1,116 @@
+from typing import List, Optional
+from spmd.tensor.placement_types import Shard, Replicate, _Partial, PlacementSpec
+
+[
```
fduwjj, line 4: QQ, how do we come up with this list and ensure its coverage?

wanchaol: It's coming from https://github.com/pytorch/pytorch/blob/fd84c458f49df2ec2801a230f332ea3730d352a9/docs/source/torch.rst#L275

It seems that it's still missing other pointwise ops as it seems not a complete list. To ensure test coverage, I guess we need to loop through the list and compare them with local tensor operations

spmd/tensor/api.py
```diff
@@ -1,30 +1,41 @@
...
+    # class attribute that handles operator placements propagation
+    # rules, keyed by aten op name, value is propagation func
+    _op_to_rules: Dict[str, Callable] = {}
 
-    # class attribute that handles ops, all handled
-    # ops should appear in this table
+    # class attribute that handles custom registered ops, all handled
+    # custom ops should appear in this table, and overriding the default
+    # operators
     # pyre-fixme[24]: Generic type `Callable` expects 2 type parameters.
```
aazzolini, line 37: Could you actually fix this -- it would help document the callable's semantic via its signatrue.

spmd/tensor/api.py
```diff
@@ -1,30 +1,41 @@
...
+
+    # class attribute that handles operator placements propagation
+    # rules, keyed by aten op name, value is propagation func
+    _op_to_rules: Dict[str, Callable] = {}
 
-    # class attribute that handles ops, all handled
-    # ops should appear in this table
+    # class attribute that handles custom registered ops, all handled
+    # custom ops should appear in this table, and overriding the default
+    # operators
```
aazzolini, line 36: Could you clarify what "default operators" means here, is it operators that are not covered by the _op_to_rules above?

spmd/tensor/api.py
```diff
@@ -74,12 +86,10 @@ def __torch_function__(cls, func, types, args=(), kwargs={}):
         # if we find nn.functional name in dispatch op, dispatch to it instead,
         # this allow us to override some python level behaviors that wouldn't be
         # possible in __torch_dispatch__ level.
-        if func.__name__ in DTensor._dist_tensor_dispatch_ops:
+        if func.__name__ in DTensor._custom_dispatch_ops:
```
aazzolini, line 89: Could it happen that there's a torch.function with the same name as a torch dispatch function? IN that case which would take priority? also is func.__name__ the same as str(func) (in the dispatch below?)

spmd/tensor/dispatch.py
```diff
@@ -0,0 +1,73 @@
...
+        # placement base on the redistribute cost
+        # TODO: implement auto distribute with simple
+        # cost estimation
+        if output_placements is None:
+            # do auto distributed/boxing here
+            raise NotImplementedError("auto redistribute not implemented yet!")
+
+        # run local op computation
+        local_results = op_info.op_call(
+            *op_info.args_with_local_tensor, **op_info.kwargs_with_local_tensor
```
aazzolini, line 48: @wanchaol I remember in our discussion we had mentioned that kwargs may need to be updated by the sharding propagation, is this something that we expect to happen in-place inside of the rules? Maybe we should clarify that , or even return a new set of kwargs from the prop. rule to make it explicit?

aazzolini: @wanchaol Another thing -- maybe a suggestion -- there's a tradeoff of generality vs. how much you can do on the sharding propagation function depending on how much info you pass into it. 

Specifically, in order for us to be able to support ahead of time propagation, would it be a good thing if we made surre to only pass "meta" tensors or "fake" tensors into the sharidn gpropagation function , so to make sure its implementaiton would never rely on .data?

spmd/tensor/ops/prop_rules.py
```diff
@@ -0,0 +1,120 @@
...
+    return PlacementSpec.from_dim_map(
+        input_specs[0].mesh,
+        [dim_to_sharding[dim] for dim in output_dim],
+        pending_sums,
+    )
+
+
+def mm_prop(
+    mat1_spec: PlacementSpec, mat2_spec: PlacementSpec
+) -> Optional[PlacementSpec]:
```
aazzolini, line 72: I think it's not hard to generalize this one , maybe you can do it here:

1) group sharding by dimension of the tensor.
2) check that the sharding-mesh-dim of dimension=1 of mat1 is the same as sharding of dimension=0 of mat2 (or that they are both replicated).
3) check that the sharding for dim=0 of mat1 and dim=1 of mat2 are _not_ on the same mesh dimension as the other ones above.

aazzolini (approved):

spmd/tensor/api.py
wanchaol: torch_function and torch_dispatch func.__name__ should be different for every operator, bc torch_dispatch will append a "default" to every op while torch_function won't have it. Example difference for op `mm`:

torch function `func.__name__`: `mm`
torch function `str(func)`: `<built-in method mm of type object at 0x10a5eb380>`
torch dispatch `func.__name__`: `mm.default`
torch dispatch `str(func)`: `aten.mm.default`

we are using the first and last one. I think the last one should be proper to distinguish clearly with torch_function, where it always work on `aten` ops

spmd/tensor/api.py
wanchaol: it's actually the operator of `_op_to_rules` + fallback, in another word: `_custom_dispatch_ops` take the highest priority. i'll update the comment.

spmd/tensor/ops/matrix_ops.py
```diff
@@ -1,109 +1,31 @@
...
-    transposed_local_mat = local_mat.t()
-    device_mesh = self.device_mesh
 
-    new_shard_dim = 1 if mat_placement.is_shard(dim=0) else 0
-    return DTensor.from_local(
-        transposed_local_mat, device_mesh, [Shard(new_shard_dim)]
-    )
+@register_prop_rule("aten.t.default")
+def transpose_rule(op_info: OpInfo) -> Optional[PlacementSpec]:
+    return einop_prop("ij->ji", op_info.args_spec)
```
fduwjj, line 31: N00b question, what does "eion" here stands for?

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -50,7 +50,7 @@ def test_mm(self):
         grad_res = torch.ones(12, 16)
         grad_dist_res = distribute_tensor(grad_res, device_mesh, shard_spec)
         dist_res.backward(grad_dist_res)
-        print(mat1.grad)
+        # print(mat1.grad)
```
fduwjj, line 53: Nit: remove? or make it assertTrue(mat1.grad is not None)?

test/spmd/tensor/test_tp_sharding_ops.py
```diff
@@ -127,7 +127,6 @@ def test_replicated_permute(self):
         new_dt = dist_tensor.permute(1, 0, 2)
         self.assertTrue(new_dt.placements[0].is_replicate())
         self.assertEqual(new_dt.to_local(), tensor.permute(1, 0, 2))
-        self.assertEqual(new_dt.stride(), tensor.permute(1, 0, 2).stride())
```
fduwjj, line 130: Why do we remove this?

spmd/tensor/placement_types.py
```diff
@@ -38,4 +39,46 @@ class Replicate(Placement):
...
+        # and int >=0 represent shard on that device mesh dim
+        r = [-1] * self.ndim
+        for i, placement in enumerate(self.placements):
+            if placement.is_shard():
+                shard_dim = cast(Shard, placement).dim
+                r[shard_dim] = i
+        return r
+
+    @classmethod
+    def from_dim_map(
```
fduwjj, line 66: It would be nice that we can add more documentations on how to use this API and `dim_map`. Or if you want to merge this one first, we can add a TODO here for a detailed document which covers different scenarios on how users can use this API to define specs for examples like [Replicate(), Shard(0), Shard(1)], or [Shard(0), Shard(0)].

spmd/tensor/ops/prop_rules.py
```diff
@@ -0,0 +1,120 @@
...
+        ij,j->ij - broadcasted addition
+    """
+    alphabet = "abcdefghijklmnopqrstuvwxyz"
+    max_dim = max(input.ndim for input in input_specs)
+    dimchars = []
+    for input in input_specs:
+        start_dim = max_dim - input.ndim
+        p = alphabet[start_dim:max_dim]
+        dimchars.append(p)
+    out_dimchars = alphabet[:max_dim]
```
fduwjj, line 118: Why this one does not come with start_dim?

spmd/tensor/dispatch.py
wanchaol: chatted offline: We are actually only updating the “non-tensor” args or “non-tensor” kwargs, which means we don’t even need the “tensor” args/kwargs metadata (i.e. `tensor.view(*shape)`, no need for tensor metadata like size, we only need to update the *shape tuple to update the local tensor shape). The only reason we keep “tensor” args in `args_with_local_tensor`, is that after the modification on “non-tensor” args, we could directly pass `args_with_local_tensor` to the local op computation `op_info.op_call(*op_info.args_with_local_tensor, **op_info.kwargs_with_local_tensor)`. We will need to figure out a way to disallow modification of "tensors" data on `args_with_local_tensor` with each sharding propagation rules.

aazzolini: Makes sense. Just note that we should if possible disallow _read_ access to .data as well, otherwise we won't be able to do ahead-of-time sharding propagation.

spmd/tensor/placement_types.py
wanchaol: added :)

spmd/tensor/ops/prop_rules.py
wanchaol: this is the logic to deal with implicit broadcasting, where the out_dimchars would always be the one that have the max_dim

test/spmd/tensor/test_tp_sharding_ops.py
wanchaol: sorry that was a accidental delete, added back

spmd/tensor/ops/prop_rules.py
wanchaol: yeah this make sense, this PR is getting too long, let me create a follow up issue and we can update incrementally :)

wanchaol: https://github.com/pytorch/PiPPy/issues/335

spmd/tensor/dispatch.py
wanchaol: I just updated the PR with OpSchema, where I explicitly document "what can be used" and "what can be changed" as propagation rule author. "tensor" argument should be explicitly disallowed, but we didn't have a way to check if it's getting modified or not. We can take this as a follow up and make sure there's no way to modify/read tensor data in propagation rules https://github.com/pytorch/PiPPy/issues/336

regarding on disallow read access of .data, could you elaborate a bit more on it, is it because ahead of time sharding propagation likely won't have any .data materialized?

PR #322, opened 2022-08-04
[spmd] switch device mesh collective to use mesh dim groups

wanchaol:
This PR:
* switches device mesh to use dim groups on all collectives
by specifying the `mesh_dim` argument in the collective wrapper calls.
* It switches `scatter` to use c10d collective directly, as the `nn.functional.scatter` have some bugs, we will need to follow up separately. 
* It also deal with the logic that src rank based collective need to
pass in global rank instead of local rank
* It also add some simple example tests on 1d mesh, testing 2d/multi-dim mesh
will come as separate PRs

### DIFF
--- a/spmd/tensor/device_mesh.py
+++ b/spmd/tensor/device_mesh.py
@@ -6,7 +6,10 @@ from torch.distributed.distributed_c10d import (
     get_rank,
     get_world_size,
     ReduceOp,
+    GroupMember,
+    scatter,
     _get_default_group,
+    _get_global_rank,
     _reduce_scatter_base,
     new_group,
     ProcessGroup,
@@ -18,7 +21,6 @@ from torch.distributed.nn.functional import (
     _all_gather_base,
     all_reduce,
     broadcast,
-    scatter,
 )
 
 _global_device_mesh: Optional["DeviceMesh"] = None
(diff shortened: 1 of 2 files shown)


spmd/tensor/device_mesh.py
```diff
@@ -211,44 +217,78 @@ def get_rank(self) -> int:
         return get_rank()
 
     def scatter(
-        self, tensors: List[torch.Tensor], src: int = 0
+        self, scatter_list: List[torch.Tensor], mesh_dim: int = 0
     ) -> torch.Tensor:
-        return scatter(tensors, src=src)
+        to_scatter = [tensor.contiguous() for tensor in scatter_list]
```
fduwjj, line 222: So does this solve the scatter errors observed in rowwise sharding of Linear module?

fduwjj (approved):
Nice!!! Thanks for fixing that. This is indeed needed right now.

spmd/tensor/device_mesh.py
wanchaol: Yep I manually checked it, this should solve it, the `nn.functional.scatter` have two bugs:
1. no contiguous call
2. checking local rank instead of global rank
we gonna need to fix it in the core

spmd/tensor/device_mesh.py
```diff
@@ -211,44 +217,78 @@ def get_rank(self) -> int:
         return get_rank()
 
     def scatter(
```
aazzolini, line 219: Maybe write a docstring to clarify semantic.

aazzolini (approved):
Looks great!

spmd/tensor/device_mesh.py
wanchaol: added a docstr for `scatter`, will add more docstr for other collectives in following PRs.

PR #325, opened 2022-08-04
[spmd] add multi-d tests for all_reduce, broadcast, scatter

ShawnYLuo:
add simple example tests on mult-d (3d) mesh for all_reduce, broadcast, scatter

### DIFF
--- a/test/spmd/tensor/test_device_mesh.py
+++ b/test/spmd/tensor/test_device_mesh.py
@@ -186,5 +186,78 @@ class DeviceMeshTest(DistTensorTestBase):
         )
 
 
+class DeviceMeshCollectiveTest(DistTensorTestBase):
+    @property
+    def world_size(self):
+        return 8
+
+    @with_comms
+    def test_all_reduce_1d(self):
+        mesh = DeviceMesh(self.device_type, torch.arange(self.world_size))
+        local_tensor = torch.ones(3, 3) * self.rank
+        reduced_tensor = mesh.all_reduce(local_tensor, mesh_dim=0)
+        res_num = ((0 + self.world_size - 1) * self.world_size) / 2
+        self.assertEqual(reduced_tensor, torch.ones(3, 3) * res_num)
+
+    @with_comms
+    def test_broadcast_1d(self):
+        mesh = DeviceMesh(self.device_type, torch.arange(self.world_size))
+        local_tensor = torch.ones(3, 3) * self.rank
+        broadcasted_tensor = mesh.broadcast(local_tensor, mesh_dim=0)
+        self.assertEqual(broadcasted_tensor, torch.zeros(3, 3))
+
+    @with_comms
+    def test_scatter_1d(self):
+        mesh = DeviceMesh(self.device_type, torch.arange(self.world_size))
+        tensor_list = [
+            torch.ones(3, 3) * rank for rank in range(self.world_size)
+        ]
+        scattered_tensor = mesh.scatter(tensor_list, mesh_dim=0)
+        self.assertEqual(scattered_tensor, torch.ones(3, 3) * self.rank)
+
+        # slight complicate use case
+        # fix the random seed
+        torch.manual_seed(0)
+        global_tensor = torch.randn(3, 3 * self.world_size)
+        # chunk by dim 1 will make it non-contiguous
+        chunked_list = global_tensor.chunk(self.world_size, dim=1)
+        # test if scatter still works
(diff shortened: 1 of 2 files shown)


test/spmd/tensor/test_device_mesh.py
```diff
@@ -186,5 +186,136 @@ def test_device_mesh_nd(self):
...
+        local_tensor = torch.ones(3, 3) * self.rank
+        expected_ranks_by_dim = [
+            [[0, 4], [2, 6], [1, 5], [3, 7]],
+            [[0, 2], [1, 3], [4, 6], [5, 7]],
+            [[0, 1], [2, 3], [4, 5], [6, 7]],
+        ]
+        for dim, dim_ranks in enumerate(expected_ranks_by_dim):
+            for ranks in dim_ranks:
+                if self.rank in ranks:
+                    reduced_tensor = mesh.all_reduce(local_tensor, mesh_dim=dim)
```
wanchaol, line 274: could you explain why we need to loop through `expected_ranks_by_dim` and then call `all_reduce`?

My impression is that for n-d mesh, we should try to run all_reduce on each dim (i.e. 3d mesh, we should try to run all_reduce on mesh dim "0, 1, 2"), and then check if `local_tensor` get all_reduced on those ranks?

test/spmd/tensor/test_device_mesh.py
```diff
@@ -186,5 +186,136 @@ def test_device_mesh_nd(self):
...
+        )
+        self.assertEqual(reduce_scattered_tensor, torch.ones(3, 3) * res_num)
+
+    @with_comms
+    def test_all_reduce_nd(self):
+        mesh_tensor = torch.arange(8).reshape(2, 2, 2)
+        mesh = DeviceMesh(self.device_type, mesh_tensor)
+        local_tensor = torch.ones(3, 3) * self.rank
+        expected_ranks_by_dim = [
+            [[0, 4], [2, 6], [1, 5], [3, 7]],
```
wanchaol, line 267: I am curious how do we get the this list, are we just print/visualize the 3d tensor like we did here? https://github.com/pytorch/PiPPy/blob/1f9ed8cfa93fcbb65d465912d12ad118f35fb86f/spmd/tensor/device_mesh.py#L160

test/spmd/tensor/test_device_mesh.py
```diff
@@ -186,5 +186,136 @@ def test_device_mesh_nd(self):
...
+            [[0, 2], [1, 3], [4, 6], [5, 7]],
+            [[0, 1], [2, 3], [4, 5], [6, 7]],
+        ]
+        for dim, dim_ranks in enumerate(expected_ranks_by_dim):
```
wanchaol, line 271: I think ideally these ranks check should actually happen within the DeviceMesh's collective wrapper itself. We can do it as a follow up task :)

ShawnYLuo: "to run all_reduce on each dim (i.e. 3d mesh, we should try to run all_reduce on mesh dim "0, 1, 2")" is exactly what I'd like to do. This variable "expected_ranks_by_dim" is here to easily see the expected result so I can just sum it. Would it be more reasonable to let it be computed in mesh.get_dim_groups()?

test/spmd/tensor/test_device_mesh.py
ShawnYLuo: Yes

> I am curious how do we get the this list, are we just print/visualize the 3d tensor like we did here?
> 
> https://github.com/pytorch/PiPPy/blob/1f9ed8cfa93fcbb65d465912d12ad118f35fb86f/spmd/tensor/device_mesh.py#L160

test/spmd/tensor/test_device_mesh.py
wanchaol: Looking into this a bit more, I think we probably don't need to do this check logic, as when we are creating sub process groups, it's only being added when the current rank in the rank of that group, and that's why we can directly call `device_mesh.scatter(mesh_dim=1)` without checking my_rank (if my_rank not in the group, it's not calling the collective on that group so wouldn't be a issue). https://github.com/pytorch/PiPPy/blob/1608285749307c30f65dd85f4110810076e7f3cb/spmd/tensor/device_mesh.py#L175

wanchaol:
btw https://github.com/pytorch/PiPPy/pull/322 is landed, so you can get rid of my commit and rebase to the main branch in order to get this PR merged :)

PR #328, opened 2022-08-04
[spmd] add __torch_function__ to allow us override on python level

wanchaol:
as titled, some ops that not going through dispatch, we need to override
it in torch_function level, we also need to add some gradient hook in the
future to ensure gradient layout is the same, this should also happen
in the torch_function level.


fduwjj (approved):
lgtm.

PR #330, opened 2022-08-06
[spmd] Rename Tensor to DTensor and change APIs

wanchaol:
API rename:
1. rename Tensor to DTensor for us to use internally, we could choose later if we still need name alias on package level.
2. rename `local_tensor()` to `to_local()` as it aligns more with `from_local`


soumith (approved):
stamped!

PR #332, opened 2022-08-09
Fix #331: improve Pipeline Driver issuance efficiency

### LINKED ISSUE #331: Pipeline Driver slowing down pipeline execution
### Problem Statement:
As Figure 1 shows, the backward stages were not executed as soon as the previous chunk was completed. Instead, they waited for the driver to issue the `invoke` command.

Screen Shot 2022-08-08 at 2 09 52 PM

### Analysis:
Zooming in, there are two factors that cause the delay of invoke calls.
1. Current Pipeline Interpreter uses sync RPC calls. Thus, it blocks the driver till the remote executor finishes the invoke call and returns. See "IB_7,0" for example.
2. Each `getitem` invocation is a separate RPC call. If a stage has many output indices, then the driver would have to issue many RPC calls. See "G7,0" for example.
Screen Shot 2022-07-25 at 2 01 05 PM

Figures are from T5 3B model, with 8 chunks (8 PP ranks), and small chunk size.

kwen2501:
### Changes
1. Change invoke calls from RPC sync calls to RPC async calls
2. Coalesce `getitem` calls
3. Correct stage issuance order

### Result
Tested on T5 3B config, 8 GPUs, 8 stages, 8 chunks, small chunk size
1. Before PR:
One iteration takes 1.05 second.
Screen Shot 2022-08-05 at 4 37 13 PM

2. After PR:
One iteration takes 671 ms.
Screen Shot 2022-08-05 at 4 36 59 PM

Fix issue #331


pippy/PipelineDriver.py
```diff
@@ -593,32 +592,63 @@ def extract_value_ref_args(arg):
...
-    def index_value(self, output_unique_key : str, output_refcount : int, value_ref, idx):
-        # For the purposes of refcounting, decrement this use
-        with self.value_store_lock:
-            refcounted_future = self.value_store[value_ref.unique_key]
-            if refcounted_future.release():
-                self.value_store.pop(value_ref.unique_key)
+        return ValueReference(self.stage_id, output_unique_key)
 
-            indexed = refcounted_future.future.then(lambda f: f.value()[idx])
+    def coalesce_index_value(self, indices : List[Tuple]):
```
jamesr66a, line 606: nit: `coalesce` sounds like the verb here, but i think that should be an adjective, like "coalesced index value". Maybe change to `coalesced`

jamesr66a: nit: also maybe have a more precise type annotation for `indices` by typing the contained values, e.g. `List[Tuple[str, int, ValueRef, int]]` (not sure if that's right but you get the idea)

pippy/PipelineDriver.py
```diff
@@ -593,32 +592,63 @@ def extract_value_ref_args(arg):
...
+        for index_tuple in indices:
+            (output_unique_key, output_refcount, value_ref, idx) = index_tuple
+            logging.info(f'[{self.stage_id}] Received getitem call: {(output_unique_key, output_refcount, value_ref, idx)}')
+            with self.value_store_cv:
+                if output_unique_key in self.value_store:
+                    # TODO: investigate why there are repeated getitem calls with 0 users
```
jamesr66a, line 612: Make an issue

Also this looks to be guarded by an assert. Are there situations the user would run into where this assert is false and this tears down the program? would that be a user-facing bug/failure?

pippy/PipelineDriver.py
```diff
@@ -1094,6 +1126,8 @@ def __init__(self, remote_stage_executor_rrefs, stage_to_executor, module, cur_m
         self.args_iter = iter(self.args)
         self.batch_id = batch_id
         self.num_microbatches = num_microbatches
+        # Dict from stage id to a list holding the coalesced getitem indices
+        self.stage_indices : Dict[int, List[Tuple]] = {}
```
jamesr66a, line 1130: nit: maybe make this more specific like `stage_output_indices`

Also same comment as above about more precise type annotation

pippy/PipelineDriver.py
```diff
@@ -1217,6 +1281,28 @@ def wait_for_confirmation(n):
         return node
 
 
+class run_until_criteria:
```
jamesr66a, line 1284: Make this private i.e. `_run_until_criteria`

jamesr66a (approved):
image

pippy/PipelineDriver.py
kwen2501: Thanks, fixed.

pippy/PipelineDriver.py
kwen2501: First of all, I guess there should not be repeated index calls. Should there be?
Second, the repeated calls I have seen (with T5 only) all have 0 users, so I am just letting it pass **for now** as no other stage is using it.
But, if there is indeed a repeated call with non-zero users, then I would like it to error out and let user report to us.
Thus, I added an assertion error message in a new commit:
"PiPPy internal error: repeated index value call with non-zero users"

pippy/PipelineDriver.py
kwen2501: Thanks, fixed.

pippy/PipelineDriver.py
kwen2501: Thanks, fixed.

pbelevich (approved):

PR #333, opened 2022-08-09
[SPMD] Enable GPU CI for Distributed Tensor

fduwjj:
We have not enabled distributed tensor test for GPUs. Let's do that.

Also add an E2E test for a Megatron-LM style TP.


test/spmd/tensor/test_megatron_example.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+
+
+class DistTensorMegatronTest(DistTensorTestBase):
+    @with_comms
+    def test_simple_megatron_e2e(self):
+        LR = 0.5
+        inp_size = [5, 10]
+        torch.manual_seed(0)
+        inp = torch.rand(*inp_size).cuda(self.rank)
```
wanchaol, line 93: nit: this will fail on CPU right? can we try to make the whole test be device agnostic? i.e. don't set the `cuda()` call, as `distributed_tensor` could automatically convert to proper type specified by `device_mesh`? as long as we set `device_mesh` to be sth like `DeviceMesh(self.device_type, [1, 2, 3, 4]`, then it should work for both CPU and GPU

fduwjj: I see. Sure.

.github/workflows/spmd_gpu_tests.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Install dependencies
+# Turn off progress bar to save logs
+pip3 install --upgrade pip
+pip3 config set global.progress_bar off
+pip3 install flake8 pytest pytest-cov pytest-shard numpy expecttest
+if [ -f requirements.txt ]; then pip3 install -r requirements.txt --find-links https://download.pytorch.org/whl/nightly/cu102/torch_nightly.html; fi
+
+# Install pippy
+python3 smpd/setup.py install
```
wanchaol, line 27: -> spmd

.github/workflows/spmd_tests.yaml
```diff
@@ -29,3 +29,69 @@ jobs:
...
+            -e VERBOSE \
+            -e OMP_NUM_THREADS \
+            --tty \
+            --detach \
+            -v "$(pwd):${PIPPY_ROOT}" \
+            -w "${PIPPY_ROOT}" \
+            "${DOCKER_IMAGE}"
+          )
+          # Run GPU tests and return error signal from docker
+          docker exec -t -w "${PIPPY_ROOT}" "${container_name}" bash -c "bash .github/workflows/spmd_gpu_tests.sh; exit $?"
```
wanchaol, line 84: so currently this command does not work as docker won't return error signal from test?

fduwjj: At least from CI it is not working as expected.

spmd/tensor/api.py
```diff
@@ -380,8 +381,12 @@ def _view_with_sharding_dim_change(self, sharding_dim, shape):
 
             # Infer the dim which is specified with -1.
             if infer_idx is not None:
-                st_size = math.prod(self.size())  # type: ignore[attr-defined]
-                shape_size = -1 * math.prod(shape)  # type: ignore[attr-defined]
+                st_size = reduce(
```
yhcharles, line 384: Curious, why do we need to replace `math.prod` with `reduce(operator.mul, ...)`?

yhcharles (approved):
LGTM

spmd/tensor/api.py
fduwjj: Because this is only supported since Python 3.8 however the python version in the docker we used for gpu test is just 3.7...

yhcharles: I see 👍🏻

.github/workflows/spmd_tests.yaml
```diff
@@ -39,3 +39,71 @@ jobs:
...
         run: |
           pytest --shard-id=${{ matrix.shard }} --num-shards=4 --cov=spmd test/spmd/
+
+  pytest_tests_gpu:
+    runs-on: linux.16xlarge.nvidia.gpu
+    strategy:
+      matrix:
+        shard: ["0", "1", "2", "3"]
+    env:
+      DOCKER_IMAGE: gingerhugo/cuda-11.3-python-3.9:v1.0.1
```
wanchaol, line 49: Should we point this to the dockerfile you added in this PR?

fduwjj: This is indeed the docker image built from the Dockerfile in this PR. This is the repo I created on Docker hub, lol.. It was a wired name because I just created them a while back when I was still at school.

fduwjj:
Turns out the root cause of CI failure is due to this one: https://github.com/NVIDIA/nccl/issues/290

fduwjj:
This is reference for installing Python from source: https://hackersandslackers.com/multiple-versions-python-ubuntu/

fduwjj:
CI all passed and linter fixed. Let's merge it.

PR #334, opened 2022-08-09
Move stage_backward etc to a separate file

### LINKED ISSUE #10: Move stage_backward to a separate file
It's kinda weird for this to be in `IR.py` considering it also implements the runtime semantics of running backward: https://github.com/jamesr66a/PiPPy/blob/527af1fd8123d35bd81b9fe304a8d0ed29c9fd8d/pippy/IR.py#L18

Move it to a separate common file that both IR and PipelineDriver import

kwen2501:
Fix #10


pbelevich (approved):
LGTM!

PR #337, opened 2022-08-09
[spmd] fix some cuda tests

wanchaol:
ProcessGroupNCCL does not behave exactly the same as ProcessGroupGloo
which result in some test breakage in CI, this PR fixes it


fduwjj (approved):
Thanks for the fix!

test/spmd/tensor/test_pointwise_ops.py
```diff
@@ -50,6 +52,7 @@ def _reset_random_seed():
             [Shard(0)],
             (8, 5),
             torch.nn.functional.dropout,
+            reset_seed=_reset_random_seed,
```
fduwjj, line 55: Wait we don't need this right? because training is False

PR #340, opened 2022-08-10
Finish all lint tests even if previous one fails

kwen2501:
So that we get all lint error in one CI, rather than fix and rerun, fix and rerun, ...


wanchaol (approved):
looks good, is there a way to merge without using github web page?

kwen2501:
> 

You can do `git merge <branch>` on the `main` branch, if you are confident that the PR will not break main.

PR #344, opened 2022-08-10
[SPMD] Add one Megatron simple e2e test and remove relative import

fduwjj:
We need a E2E Megatron style test to ensure that any further change down the road does not break it.

### DIFF
--- a/spmd/tensor/ops/pointwise_ops.py
+++ b/spmd/tensor/ops/pointwise_ops.py
@@ -1,9 +1,7 @@
 from typing import Optional
 from spmd.tensor.api import DTensor
 from spmd.tensor.dispatch import OpSchema
-from spmd.tensor.placement_types import (
-    PlacementSpec,
-)
+from spmd.tensor.placement_types import PlacementSpec
 from spmd.tensor.ops.prop_rules import pointwise_prop
 
 # leave the pointwise_ops list here for convenience,
@@ -120,10 +118,7 @@ from spmd.tensor.ops.prop_rules import pointwise_prop
 #     "xlogy",
 # ]
 
-pointwise_ops = [
-    "aten.relu.default",
-    "aten.gelu.default",
-]
+pointwise_ops = ["aten.relu.default", "aten.gelu.default"]
 
 
 def pointwise_rules(op_schema: OpSchema) -> Optional[PlacementSpec]:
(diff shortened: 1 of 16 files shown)


spmd/tensor/ops/pointwise_ops.py
```diff
@@ -120,10 +118,7 @@
 #     "xlogy",
 # ]
 
-pointwise_ops = [
-    "aten.relu.default",
-    "aten.gelu.default",
-]
+pointwise_ops = ["aten.relu.default", "aten.gelu.default"]
```
wanchaol, line 121: are we using the same formatter command? `black --line-length 80 spmd test/spmd`? it seems I use this command on my end but it never reformatted like this?

test/spmd/tensor/test_megatron_example.py
```diff
@@ -0,0 +1,115 @@
...
+        distribute_tensor(m.net2.bias, device_mesh, replicate)
+    )
+    m = _replicate_input_tensor(m, device_mesh, replicate)
+    m.net2 = _aggregate_local_tensor(m.net2)
+    m.net1.weight.register_hook(
+        functools.partial(_gradient_hook, m.net1.weight)
+    )
+
+
+class DistTensorMegatronTest(DistTensorTestBase):
```
wanchaol, line 80: wouldn't this not working yet? I think CI will fail as the local patches not submitted yet

test/spmd/tensor/test_megatron_example.py
```diff
@@ -0,0 +1,115 @@
...
+        optim.step()
+
+        torch.manual_seed(3)
+        inp = torch.rand(*inp_size).cuda(self.rank)
+        output = model(inp)
+        output_tp = model_tp(inp)
+        self.assertEqual(output, output_tp)
+
+
+if __name__ == "__main__":
```
wanchaol, line 114: maybe try removing this `main`? but probably pytest would still scan this test case lol

test/spmd/tensor/test_megatron_example.py
fduwjj: Yes, I can merge this after your fix got landed?

test/spmd/tensor/test_megatron_example.py
fduwjj: Hmmm you know what, maybe I can comment it out. And you can open it when your fix is ready. Let me do that.

spmd/tensor/ops/pointwise_ops.py
fduwjj: Yes I use "black --line-length 80 spmd test/spmd" and it just reformat like this...

wanchaol (approved):
thanks!!

PR #345, opened 2022-08-10
[DTensor] Add tests for sigmoid ops

anj-s:
Added the following for checking sigmoid op coverage for DT:
1. Registered `aten.sigmoid.default` in list of supported pointwise ops
2. Added tests for 1-D mesh for replicate and shard specs.

Open questions:
1. Should we add tests for 2D since that is the more complex case? 
2. Do we need to add 1 shard and replicate for all pointwise ops or is there something more generic we can add?

### DIFF
--- a/spmd/tensor/dispatch.py
+++ b/spmd/tensor/dispatch.py
@@ -7,7 +7,7 @@ from spmd.tensor.placement_types import PlacementSpec
 """
 If set to true, __DEBUG_STRICT will fail when an op doesn't have a sharding rule registered.
 """
-_DEBUG_STRICT = True
+_DEBUG_STRICT = False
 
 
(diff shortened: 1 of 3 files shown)


wanchaol (approved):
looks great!! Thanks!!

regarding on the open question, yep I think for pointwise it should support 1d -> nd placement types, we can have tests for 1d -> 3d mesh to test all working well, sth like:
1d: `[[Shard(0)], [Replicate()]]`
2d: `[[Shard(0), Replicate()], [Replicate(), Shard(1)]`, ..]
3d: `[[Shard(0), Shard(1), Replicate()], [....]]`

Maybe coverage up to 2d + one 3d placement is enough to test the correctness. Feel free to add these in follow up PRs!

PR #346, opened 2022-08-10
Fix code returned from docker to GHA

### LINKED ISSUE #353: [CI] Docker run does not return error code correctly
When there is a test failure inside docker, the GitHub Action is not able to catch it.

kwen2501:
Previously, our GHA command has:
```
docker exec -t -w "${PIPPY_ROOT}" "${container_name}" bash -c "bash .github/workflows/gpu_tests.sh; exit $?"
```

`$?` was expanded at the GHA level rather than inside docker. Hence docker always do `exit 0`, hiding the error.

Adding a `\` before `$` to fix this issue.

Fixed #353

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -208,7 +208,7 @@ jobs:
             "${DOCKER_IMAGE}"
           )
           # Run GPU tests and return error signal from docker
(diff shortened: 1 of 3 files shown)


fduwjj (approved):
Thanks!!

PR #354, opened 2022-08-11
[spmd] Implement basic auto redistribute for DTensor

wanchaol:
Discussed with @aazzolini offline and we should implement some basic auto redistribute to enable both eager/compiler work in certain cases. This PR:
* Implement a very barebone basic of auto redistribute feature for DTensor, where the operator sharding propagation rule itself could give suggestions to the operator dispatch in the case that sharding propagation failed.
* refactors a bit on the dispatching logic to allow auto redistribute when the input specs aren't matching a sharding
propagation rule
* now the sharding propagation only relies on the OpSchema, instead of passing local tensors into sharding propagation, which might result in some risky behavior that modifies the local tensors directly. This also fixes https://github.com/pytorch/PiPPy/issues/336
* Add a very basic example how the auto redistribute work on `addmm`, essentially when `input(bias)` is replicate and output of `mm` is partial, we fail the sharding propagation rule and return suggestion to convert `input(bias)` to partial. 
* Add the redistribute functionality from replicate -> partial, note that this is disabled from user facing API, but enabled for our internal operator implementations.


spmd/tensor/ops/matrix_ops.py
```diff
@@ -1,27 +1,60 @@
...
 
 @register_prop_rule("aten.addmm.default")
-def addmm_rules(op_schema: OpSchema) -> Optional[PlacementSpec]:
+def addmm_rules(op_schema: OpSchema) -> OutputSharding:
     input_spec, mat1_spec, mat2_spec = op_schema.args_spec
     mm_out_spec = mm_prop(mat1_spec, mat2_spec)
     if mm_out_spec is None:
-        return None
-    return pointwise_prop((input_spec, mm_out_spec))
+        # non-eligible input, suggest addmm input specs
```
aazzolini, line 20: Maybe you can do instead something like?


  return pointwise_rules([mm_out_spec.placements[0], input_spec], linear=True)

spmd/tensor/redistribute.py
```diff
@@ -76,20 +73,50 @@ def redistribute_spmd_tensor(
...
+                # get global_rank and in the case of multidim mesh it's not correct!
                 new_local_tensor = local_tensor.narrow(
                     shard_dim, my_rank * chunk_size, chunk_size
                 )
             else:
                 # diff shard dim on new placement, record in attempted transforms
                 attempted_transforms.append(current)
+        elif target.is_partial():
+            if current.is_replicate():
+                if is_backward:
```
aazzolini, line 86: I'm  trying to understand why this would be called in the first place in the backward. It thought this would only be called if it was necessary for the next operation to succeed, otherwise i thought we'd never do replicated to partial. Is there a case where this is going to hit even if it's not necessary?

aazzolini: Also is this the only case where we will silently ignore the desired target?

spmd/tensor/redistribute.py
```diff
@@ -111,7 +138,10 @@ def backward(ctx, grad_output: "spmd_tensor.DTensor"):  # type: ignore
         previous_device_mesh = ctx.previous_device_mesh
         return (
             redistribute_spmd_tensor(
-                grad_output, previous_device_mesh, previous_placement
+                grad_output,
+                previous_device_mesh,
+                previous_placement,
+                is_backward=True,
```
aazzolini, line 144: oh i see this is the backward for redistribute itself. 

I have the feeling that if user is calling redistributed manually, then backward makes sense.

But if we're calling redistribute internally as part of the operator dispatch, we should actually implement that _below_ autograd -- we are redistributing because that specific op implementation doesn't support a specific input sharding. It doesn't say anything about the "backward" implementation for that same op -- if the backward op happens no to support that sharding we should be calling redistribute again anyways. Does that make sense?

wanchaol: Yep make sense. This flag and the special treatment is purely for a manual call outside of operator dispatch. For operator dispatch itself, since it's already below autograd, so it will never trigger this `is_backward` flag, and in fact the way I call it in the operator_dispatch is to call the `redistribute_spmd_tensor` api directly instead of calling the autograd function version. https://github.com/pytorch/PiPPy/pull/354/files#diff-19bfa7cfb887150c3ec7d76b0dab34dafc0b97aeeebc2a3cd557d79cffc1d975R69

I tried to reuse the same logic here for manual call outside of operator dispatch and operator call inside torch_dispatch, let me think about how to better model this.

spmd/tensor/redistribute.py
wanchaol: Yeah I think this should be the only case we silently ignore the desired target, and it should only happen in the backward of a manual call from user, it should not happen in operator dispatch

spmd/tensor/redistribute.py
wanchaol: Ok I think what i'll do is to just make this special treatment exist in the backward call of the `Redistribute` autograd function, so that in operator_dispatch it will never hit this case.

spmd/tensor/ops/matrix_ops.py
wanchaol: could you elaborate more on what this mean? here we are trying to return a suggestion that contains a OpSchema of args/kwargs, if we return a pointwise_rules with `linear=True`, wouldn't this just propagate through without actually doing the reshard? or are you saying that we can directly run this, and don't set the output_spec of the output sharding, but rather setting the `op_schema` passed in to the `schema_suggestion` list?

aazzolini: oh i put my comment in the wrong line , see comment in the line below.

spmd/tensor/ops/matrix_ops.py
```diff
@@ -1,27 +1,60 @@
...
+                        suggested_mat2_spec,
+                    ),
+                    op_schema.kwargs_schema,
+                )
+            ],
+        )
+
+    return OutputSharding(
+        pointwise_prop((input_spec, mm_out_spec), linear=True)
+    )
```
aazzolini, line 55: What I meant is to replace this with the more general pointwis_rules implementation. THis is like calling add(Partial, Replicated), so it generalizes to a pointwise operation. The pointwise sharding propagation should take this input and then return a suggestion of Partial, Partial --- then we'd back-propagate the suggestion to the output.  (it's like chaining suggestions). This will only work if the pointwise rule reutrns a suggestion that keeps the "partial" input untouched though but we could make it in a way that ensures that's the case.

aazzolini: The reason for doing this is to reuse the code that will support multiple mesh dims.

wanchaol: Oh i see, got it, yeah this make sense, will update

spmd/tensor/redistribute.py
aazzolini: Makes sense, thanks!

aazzolini (approved):
Thanks for going through the comments, looks good!

PR #356, opened 2022-08-12
Cancel CI on previous commit when a new commit is pushed

kwen2501:
Concurrency ensures that only a single job or workflow using the same concurrency group will run at a time.

Reduce our CI load.

Reference:
https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre


fduwjj (approved):
This is sooooo great!! Let's do it!

PR #358, opened 2022-08-12
Support for multi-dim distribute and redistribute

aazzolini:
Previously we only supported single-mesh dimension for distribute and re-distribute; enabling this so we can write test cases for ops that support multi-dimensional sharding.


spmd/__init__.py
```diff
@@ -51,7 +50,7 @@ def distribute_tensor(
                 shard_dim <= tensor.ndim
             ), "Sharding dim {shard_dim} greater than tensor ndim {tensor.ndim}"
             # TODO: handle multi-dim device mesh and last shard
```
anj-s, line 52: should we update this TODO?

spmd/tensor/redistribute.py
```diff
@@ -40,38 +45,55 @@ def redistribute_spmd_tensor(
...
+            num_chunks = device_mesh.size(dim=i)
             assert (
                 input.size(shard_dim) % num_chunks == 0
             ), "Only support chunk sharding evenly now"
-            chunk_size = input.size(shard_dim) // num_chunks
-            my_rank = device_mesh.get_rank()
+            chunk_size = (
+                local_tensor.size(shard_dim) // num_chunks
+            )  # this may already be sharded on a differernt mesh dimension
+            my_rank = device_mesh.get_rank_for_dim(dim=i)
```
wanchaol, line 81: Is this rank coordinate API only used for checking if the current rank is within the mesh or not? I feel like there might be some other uses, could we document a bit with concrete TODO?

spmd/tensor/redistribute.py
```diff
@@ -18,11 +18,16 @@ def redistribute_spmd_tensor(
...
 
-    assert len(placements) == 1, "Only support 1-d placement for now"
-
     attempted_transforms = []
     new_local_tensor = None
-    for i, (current, target) in enumerate(zip(current_placements, placements)):
+
+    # we need to go backwards in the case of to respect the scatter order established in
+    # distribute_tensor(), for the cases where we're sharding multiple times on the same
+    # tensor dimension (e.g. [Shard(0), Shard(0)])
```
wanchaol, line 26: This is nice!

spmd/tensor/device_mesh.py
```diff
@@ -262,28 +276,44 @@ def broadcast(self, tensor: torch.Tensor, mesh_dim: int = 0):
         return broadcast(tensor.contiguous(), src=src_for_dim, group=dim_group)
 
     # pyre-fixme[3]: Return type must be annotated.
-    def all_gather(self, tensor: torch.Tensor, mesh_dim: int = 0):
+    def all_gather(
+        self, tensor: torch.Tensor, mesh_dim: int = 0, tensor_dim: int = 0
```
wanchaol, line 280: do we really need tensor_dim in `all_gather`? I can understand we need it in `all_gather_base` to make sure reshuffling the dim or concating it in non-0 dimension. But seems we don't need it here, or is it because we want to keep the all_gather API consistent? IMO we can probably have a follow up to delete `all_gather_base` API and use this `all_gather` API with the semantic that it will directly return a big tensor instead of a tensor list :)

wanchaol (commented):
This looks great! Left a few questions and comments :)

test/spmd/tensor/test_redistribute.py
```diff
@@ -151,5 +153,68 @@ def test_partial_to_shard_1(self):
...
+                # create a new DTensor reinterpreting some of the replicated entires as "Partial"
+                dt = DTensor(dt.to_local(), device_mesh, inputs)
+
+            if torch.distributed.get_rank() == 0:
+                print(inputs)
+
+            for outputs in all_outputs:
+                dt2 = dt.redistribute(device_mesh, outputs)
+                if torch.distributed.get_rank() == 0:
+                    print("  ", outputs)
```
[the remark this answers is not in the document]
aazzolini: what should we assert?

spmd/tensor/device_mesh.py
```diff
@@ -214,6 +221,13 @@ def backend(self) -> str:
     def get_rank(self) -> int:
         return get_rank()
 
+    def get_rank_for_dim(self, dim: int) -> Optional[int]:
```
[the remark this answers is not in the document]
aazzolini: Right this is needed in order to identify your rank within a process group for a given mesh dimension, or other things like when you wnat to do Distributed -> Sharded , where you need to pass the right offset to torch.narrow.

spmd/tensor/redistribute.py
aazzolini: It is used in order to calculate offset in the torch.narrow operation below, for replicate -> sharded  redistribution.

spmd/tensor/device_mesh.py
aazzolini: Oh makes sense -- it doesn't do anything anyways!

spmd/tensor/device_mesh.py
```diff
@@ -262,28 +276,44 @@ def broadcast(self, tensor: torch.Tensor, mesh_dim: int = 0):
...
 
     # pyre-fixme[3]: Return type must be annotated.
-    def all_gather(self, tensor: torch.Tensor, mesh_dim: int = 0):
+    def all_gather(
+        self, tensor: torch.Tensor, mesh_dim: int = 0, tensor_dim: int = 0
+    ):
         dim_group = self._dim_groups[mesh_dim]
-        return all_gather(tensor, group=dim_group)
+        if tensor_dim != 0:
+            tensor = tensor.movedim(tensor_dim, 0)
```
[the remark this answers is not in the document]
aazzolini: Ha interesting seems like transpose and movedim have same signature and same semantics as well! just replacing movedim with transpose.

spmd/tensor/device_mesh.py
```diff
@@ -130,6 +130,13 @@ def __init__(
                 f"DeviceMesh cannot have duplicate values, but found {self.mesh.tolist()}"
             )
 
+        # coordinates of this rank on the mesh
+        rank_coords = (self.mesh == get_rank()).nonzero()
+        assert rank_coords.size(0) in (0, 1)
```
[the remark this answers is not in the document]
aazzolini: Right, this is actually a case covered in test_device_mesh.py that i had to fix. 
I was going to ask you by the way whether we have a clear semantic / expectation in general for whether users are ok to call SPMD methods on non-participating ranks. 

For example, let's say we have 8 ranks on the global PG, but then we have a mesh with only 6 ranks. (e.g. 2x3) . For the two non-participating members, shouold they also call dt.redistribute() or not? Right now I believe the behavior is that redistribute will fail if called from a non-participating member. Let me know if we'd like to make it just no-op instead.

wanchaol (approved):
Looks good, left some comments regarding on testing and for the spmd device mesh behaviors.

spmd/__init__.py
```diff
@@ -42,16 +42,15 @@ def distribute_tensor(
...
 
     # distribute the tensor according to PlacementSpec
-    assert len(placements) == 1, "Only support 1-d placement now"
     for idx, placement in enumerate(placements):
         if placement.is_shard():
             placement = cast(Shard, placement)
             shard_dim = placement.dim
             assert (
                 shard_dim <= tensor.ndim
             ), "Sharding dim {shard_dim} greater than tensor ndim {tensor.ndim}"
```
yhcharles, line 51: Missing `f` for this f-string?

test/spmd/tensor/test_redistribute.py
```diff
@@ -151,5 +153,68 @@ def test_partial_to_shard_1(self):
...
+
+                local_full = dt.redistribute(
```
yhcharles, line 203: Do you mean `dt2.redistribute()` here?

aazzolini: yeah fixed now.

PR #363, opened 2022-08-13
Don't create _null_coalesce_accumulate node for placeholder node

pbelevich:


jamesr66a (approved):
LGTM but we should create an issue to systematically handle model inputs that require grad

PR #367, opened 2022-08-15
Separate pippy and spmd CI and speed up spmd CI

wanchaol:
This PR separates SPMD CI job so that each PR in subproject could
run its own CI separately without interfering each other. This help
us save some resources and shorter waiting time on changes

This also change spmd tests to use shard to speed up testing


aazzolini (approved):
Nice!

fduwjj:
Nice! Thanks for doing that!

PR #370, opened 2022-08-15
[checkpointing]Add support for 2D checkpointing with LOCAL_STATE_DICT

wz337:
We want to support 2d checkpointing with LOCAL_STATE_DICT.

Currently, each FSDP process group holds different data but will use the same FQN in the state_dict. 
To overcome that we'll use the new dist.cp extension API to rename FQNs for ShardedTensors based on the process groups that they belong to.

This requires writing subclasses of both DefaultSavePlanner and DefaultLoadPlanner.
On both subclasses, we'll identify ShardedTensor instances that are using the non-default process groups and change their FQNs to be prefixed by a PG-specific string.

@kumpera


spmd/distributed.py
```diff
@@ -0,0 +1,94 @@
...
+    return [c10d._get_global_rank(pg, i) for i in range(pg.size())]
+
+
+class ProcessGroupAwareSavePlanner(DefaultSavePlanner):
+    """
+    SavePlanner class defining the protocol used by save_state_dict() to plan
+    a process group aware save plan.
+    """
+
+    def create_local_plan(self) -> SavePlan:
```
wanchaol, line 33: Could you also add some tests for these planner changes? You can create a `test_checkpoint.py` in the folder of `/test/spmd/`

wanchaol (commented):
Please also fix the lint, you can look into the linter and follow the same instructions to run the linter locally `https://github.com/pytorch/PiPPy/blob/2d4eaebbc16a7d6f28e899e7698a8814e36182f4/.github/workflows/code-quality.yml#L21`

spmd/distributed.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        Each FSDP process group holds different data but use the same FQN in the state_dict.
+        This will modify the FQN in the save plan to be prefixed by a PG specific string.
+        """
+        save_plan = super().create_local_plan()
+
+        write_item = save_plan.items[0]
```
kumpera, line 40: We should perform the replacement on all ShardedTensor's in a state_dict that belong to a process group.

To see that behavior, you can use a model with multiple FSDP instances:
```
from torch.distributed.fsdp.wrap import (
    enable_wrap,
    wrap,
)
class SequentialModel(torch.nn.Module):
    def __init__(self, inner_module=None):
        super(SequentialModel, self).__init__()

        self.inner0 = wrap(NestedModule())
        self.inner1 = wrap(NestedModule())
        self.inner2 = wrap(NestedModule())
        self.inner3 = wrap(NestedModule())
        self.net1 = torch.nn.Linear(1000, 1000, device="meta")

    def reset_parameters(self):
        pass

with enable_wrap(wrapper_cls=FSDP, device_id=dist.get_rank(), process_group):
    model = wrap(SequentialModel())
```

spmd/distributed.py
```diff
@@ -0,0 +1,94 @@
...
+        write_item = replace(write_item, index=metadata_index)
+        save_plan.items[0] = write_item
+
+        return save_plan
+
+    def lookup_object(self, index: MetadataIndex) -> Any:
+        """
+        Remove the PG specific prefix for the fqn to allow object lookup in the state_dict.
+        """
+        if "flat_param" in index.fqn:
```
kumpera, line 55: Don't hardcode `flat_param`, record the original names when renaming and use them here. Something like the following:

```
def create_local_plan
  ...
  self.remapping[renamed_key] = original_key

def lookup_object
 ...
 index = replace(index, fqn=self.remapping.get(index.fqn, index.fqn)
```

spmd/distributed.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def init(
+        self, state_dict: STATE_DICT_TYPE, metadata: Metadata, is_coordinator: bool
+    ) -> None:
+        """
+        Initialize a instance with a fqn prefixed by a PG specific string
+        to load data into ``state_dict``
+        """
+        self.original_state_dict = state_dict
+        self.metadata = metadata
```
kumpera, line 74: self.metadata and self.is_coordinator are initialized by `DefaultLoadPlanner`, they don't need to be done here.

spmd/distributed.py
```diff
@@ -0,0 +1,94 @@
...
+        self.is_coordinator = is_coordinator
+
+        # initialize a state dict copy with a fqn prefixed by a PG specific string.
+        pg_global_ranks = get_ranks(state_dict["flat_param"]._process_group)
```
kumpera, line 78: Don't hardcode the name of the params, you need to test which keys need the renaming.

```
for key,value in state_dict:
  if insinstance(value, ShardedTensor):
   #rename
```

kumpera (changes requested):

spmd/distributed.py
```diff
@@ -0,0 +1,94 @@
+import io
+from dataclasses import replace
+from typing import Any
+
+import torch
+import torch.distributed.distributed_c10d as c10d
+from torch.distributed import ProcessGroup
+from torch.distributed._shard.checkpoint.default_planner import (
+    DefaultLoadPlanner,
+    DefaultSavePlanner,
```
yhcharles, line 10: I was trying to understand these planner classes, but failed to find them in https://github.com/pytorch/pytorch/blob/master/torch/distributed/_shard/checkpoint/__init__.py
Are these classes defined somewhere else?

spmd/distributed.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        Each FSDP process group holds different data but use the same FQN in the state_dict.
+        This will modify the FQN in the save plan to be prefixed by a PG specific string.
+        """
+        save_plan = super().create_local_plan()
+
+        write_item = save_plan.items[0]
+        metadata_index = write_item.index
+        pg_global_ranks = get_ranks(self.state_dict["flat_param"]._process_group)
+        fqn_prefix = "_".join([str(rank) for rank in pg_global_ranks])
```
yhcharles, line 43: noob question: what is `fqn`?

spmd/distributed.py
kumpera: They are in the process of being merged. This change requires a mashup of multiple PRs (dist.cp, fsdp+tp integration) plus fixes on top to them to work.

At this point point, this is the best place to land this change as all work around DistributedTensor is converging here.

spmd/distributed.py
kumpera: Fully Qualified Name, in this context it's the name of the parameter in a state dict.

For example:

```
model = torch.nn.Linear(1,1)

state_dict = model.state_dict()
state_dict.keys() == set("weight", "bias") # << those are the FQNs
```

spmd/checkpoint.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Rename all keys of sharded tensors from sub-process groups by prefixing it 
+        with a PG specific string.
+        """
+        state_dict_copy = {}
+        for k, v in state_dict.items():
+            # Only rename the fqn if the current process group is a sub-process group.
+            if (
+                isinstance(v, ShardedTensor)
+                and state_dict[k]._process_group
```
kumpera, line 44:
```suggestion
                and v._process_group
```

spmd/checkpoint.py
```diff
@@ -0,0 +1,92 @@
...
+        with a PG specific string.
+        """
+        self.original_state_dict = state_dict
+
+        state_dict_copy = {}
+        for k, v in state_dict.items():
+            # Only rename the fqn if the current process group is a sub-process group.
+            if (
+                isinstance(v, ShardedTensor)
+                and state_dict[k]._process_group
```
kumpera, line 75:
```suggestion
                and v._process_group
```

spmd/checkpoint.py
```diff
@@ -0,0 +1,92 @@
...
+                != dist.distributed_c10d._get_default_group()
+            ):
+                pg_global_ranks = get_ranks(state_dict[k]._process_group)
+                fqn = "_".join([str(rank) for rank in pg_global_ranks]) + "_" + k
+                state_dict_copy[fqn] = state_dict[k]
+            else:
+                state_dict_copy[k] = ""
```
kumpera, line 82:
```suggestion
                state_dict_copy[k] = v
```

kumpera: Assigning an empty string will incorrectly try to load everything as a BytesIO which is not what we want.

spmd/test/test_checkpoint.py
```diff
@@ -0,0 +1,100 @@
...
+            state_dict = model.state_dict()
+
+            cp.load_state_dict(
+                state_dict=state_dict,
+                storage_reader=cp.FileSystemReader(path=CHECKPOINT_DIR),
+                planner=ProcessGroupAwareLoadPlanner(),
+            )
+            model.load_state_dict(state_dict)
+
+        tensor_dict = {
```
kumpera, line 83: Can you add some comments here explaining why those values are the ones we're expecting to see?

kumpera (changes requested):
Looks great, a couple of changes needed and it's ready.

spmd/checkpoint.py
```diff
@@ -0,0 +1,92 @@
...
+        self, state_dict: STATE_DICT_TYPE, metadata: Metadata, is_coordinator: bool
+    ) -> None:
+        """
+        Rename all keys of sharded tensors from sub-process groups by prefixing it 
+        with a PG specific string.
+        """
+        self.original_state_dict = state_dict
+
+        state_dict_copy = {}
+        for k, v in state_dict.items():
```
rohan-varma, line 71: looks like this renaming logic is the same for save aware and load aware planner, shall we separate it into a helper function?

kumpera (approved):
This looks great!

PR #371, opened 2022-08-15
Do not run CI for previous commit if PR gets updated

fduwjj:
If the same branch gets updated, we should cancel the CI for the previous push to save some resources.

Turns out that group: ${{ github.ref }} only works for one CI test script, we need to make it more detailed so that both PiPPy and spmd can use different groups here.

### DIFF
--- a/.github/workflows/pippy_tests.yaml
+++ b/.github/workflows/pippy_tests.yaml
@@ -15,7 +15,7 @@ on:
 
 concurrency:
   # Cancel CI on previous commit when a new commit is pushed to the same branch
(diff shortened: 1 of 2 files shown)


kwen2501 (approved):
LGTM.
Thanks a lot for fixing this!
Ref: https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre

wanchaol (approved):
lgtm, thanks!

PR #375, opened 2022-08-16
[spmd] Add reduction rule, use reduction rule in sum

wanchaol:
This PR enable reduction rule, add/rewrite sum to use the reduction
rule instead of manual op impl

### DIFF
--- a/spmd/tensor/ops/math_ops.py
+++ b/spmd/tensor/ops/math_ops.py
@@ -1,9 +1,8 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
-from typing import List, Dict
+from typing import List, Dict, cast
 from spmd.tensor.api import DTensor
 from spmd.tensor.dispatch import OpSchema, OutputSharding
-from spmd.tensor.placement_types import Replicate, _Partial, PlacementSpec
-from spmd.tensor.ops.utils import register_impl
+from spmd.tensor.placement_types import _Partial, PlacementSpec
 
 
 def _gen_placement_spec_with_pending_sum(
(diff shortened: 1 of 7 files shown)


fduwjj (approved):
lgtm.

spmd/tensor/ops/math_ops.py
```diff
@@ -117,24 +117,37 @@ def einop_rule(
...
-        return DTensor(
-            local_sum, device_mesh=device_mesh, placements=self.placements
+def reduction_rule(op_schema: OpSchema) -> OutputSharding:
+    """
+    Propagate the sharding for reduction operations. Examples:
+        ij->i - sum on dim
+    """
+    alphabet = "abcdefghijklmnopqrstuvwxyz"
+    # reduction op usually begin with a single tensor
+    input_spec = cast(PlacementSpec, op_schema.args_schema[0])
```
fduwjj, line 127: This is actually the spec of the input Distributed Tensor, right?

spmd/tensor/ops/math_ops.py
```diff
@@ -117,24 +117,37 @@ def einop_rule(
...
+    if len(op_schema.args_schema) > 1 and isinstance(
+        op_schema.args_schema[1], (int, list)
+    ):
+        # this is usually a dim-based reduction op schema pattern, it
+        # might not be true for every op for other special cases, we
+        # need to specialize them as needed.
+        # TODO: add support for things like `torch.unique` where it
+        # does not follow the reduction op convention.
+        dim_list = as_list(op_schema.args_schema[1])
+        out_dimchars = input_chars.translate(
```
fduwjj, line 139: Interesting.. Python is magic!

spmd/tensor/ops/math_ops.py
wanchaol: Yep

spmd/tensor/ops/math_ops.py
wanchaol: translate seems super fast compare with other string replacement after some search, so i use this haha, magic indeed.

PR #379, opened 2022-08-17
[spmd] Add tensor shape to spec and rename spec

wanchaol:
add tensor shape to placement spec, and rename to DTensorSpec instead.


aazzolini (approved):
Thanks!

PR #382, opened 2022-08-17
[Not for landing] Local change to enable TP in ViT model prototyping

fduwjj:
Local change for ViT TP work.

Not for landing but for reference for DT folks.


fduwjj:
Not needed and have a new PR for Multihead Attention in https://github.com/pytorch/tau/pull/477/files. Close this one.

PR #383, opened 2022-08-18
Add driver related fields in PiPPyTrainingArguments

kwen2501:


kwen2501:
This PR has been cherry-picked into #380 which has been landed.
Closing.

PR #385, opened 2022-08-18
[spmd] add support for inplace op registration

wanchaol:
inplace op should return self rather than being re-wrapped, also
add a simple aten.add_ op here to test inplace op semantic, for
add opertor itself we should test separately

### DIFF
--- a/spmd/tensor/dispatch.py
+++ b/spmd/tensor/dispatch.py
@@ -90,7 +90,9 @@ class OutputSharding:
 
 def is_inplace_op(op_call: torch._ops.OpOverload) -> bool:
     # util function to check if an op is inplace or not
-    # TODO: bind the core inplace op check to python
+    # TODO: This function is copied from ATen/BatchedFallback.cpp
+    # We should bind the inplace op check to schema python
+    # bindings directly.
     op_schema = op_call._schema
     if not op_schema.is_mutable or len(op_schema.returns) != 1:
         return False
--- a/spmd/tensor/dispatch.py
+++ b/spmd/tensor/dispatch.py
@@ -4,6 +4,7 @@ from typing import List, Callable, Dict, Tuple, Optional, cast
 
 import torch
 from torch.utils._pytree import tree_map
+from torchgen.model import FunctionSchema, SchemaKind
 
 import spmd.tensor.api as spmd_tensor
 from spmd.tensor.placement_types import DTensorSpec, OutputSpecType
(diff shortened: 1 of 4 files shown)


spmd/tensor/dispatch.py
```diff
@@ -85,8 +88,27 @@ class OutputSharding:
     failed_reason: Optional[str] = None
 
 
+def is_inplace_op(op_call: torch._ops.OpOverload) -> bool:
+    # util function to check if an op is inplace or not
+    # TODO: bind the core inplace op check to python
```
yhcharles, line 93: For my learning: what does "bind to python" mean?

yhcharles (approved):
LGTM

spmd/tensor/dispatch.py
```diff
@@ -85,8 +88,27 @@ class OutputSharding:
     failed_reason: Optional[str] = None
 
 
+def is_inplace_op(op_call: torch._ops.OpOverload) -> bool:
```
aazzolini, line 91: does this cover calls such as 
add(a, b, out=a)

or more generally the "out" argument?

spmd/tensor/dispatch.py
wanchaol: This is actually the logic I copied from aten C++ code, I think it should be binded to the `_schema` directly instead of we do this manual check here, hence the comment, I'll update with more details. https://github.com/pytorch/pytorch/blob/a8b93cb3ec7cb86a5c9200e8d58b37e9f793e82c/aten/src/ATen/BatchedFallback.cpp#L48

spmd/tensor/dispatch.py
wanchaol: any op with "out" argument is a separate overload compare to the inplace op. I think the reason is that it might be `add(a, b, out=b)` then it's not technically a inplace op, hence separating. we will need to add "out" argument overload to rules separately i.e. https://github.com/pytorch/pytorch/blob/2f95b61ceaab65580d8dece821693d8b31ecfca8/aten/src/ATen/native/native_functions.yaml#L478

wanchaol: actually let me see if it's easy to add `out` variant as well

spmd/tensor/ops/pointwise_ops.py
```diff
@@ -121,6 +121,7 @@
     "aten.relu.default",
     "aten.gelu.default",
     "aten.sigmoid.default",
+    "aten.add_.Scalar",
 ]
```
fduwjj, line 125: Maybe also add "aten.mul_.Scalar"?

PR #388, opened 2022-08-19
Add CUDA support for HF examples

pbelevich:


pippy/hf/utils.py
```diff
@@ -88,7 +88,13 @@ class PiPPyTrainingArguments(TrainingArguments):
 
     @property
     def device(self):
-        return super().device
+        if self.rank == -1:
```
kwen2501, line 91: nit: is `self.rank` expected to be -1? 
If not, we can add a warning here saying we are doing a fallback to default devices (so that user knows where to debug).

kwen2501 (approved):
Thanks for the fix. LGTM! Just added a minor comment.

PR #401, opened 2022-08-24
[spmd] use ops db for testing

wanchaol:
This PR use ops db to test all the ops that distributed tensor supports.
This is crucial for us as we are trying to increase the coverage of
different operators, and we can't write every tests for so many opereators.

So we pull in the operator tests from the core repo and use that to test
all the opereators to reduce the burden.


test/spmd/tensor/test_dtensor_ops.py
```diff
@@ -0,0 +1,875 @@
...
+# This environment variable controls whether or not we print expected failure
+# lists at the end of a test suite run.  The intended usage looks like this:
+#
+# 1. Run `PYTORCH_COLLECT_EXPECT=1 python test/test_dtensor_ops.py` on a CUDA build
+#    of PyTorch that has LAPACK/MAGMA installed.
+# 2. Given the printed skip/xfail list, add them to the corresponding lists;
+#    torch.* entries go in meta_function and aten.* entries go in meta_dispatch.
+#    If there are preexisting entries, you need to merge in the entries.
+#
+# This is somewhat manual but typically you shouldn't need to do this, unless
```
fduwjj, line 123: Do you mean this is just a manual test?

test/spmd/tensor/test_dtensor_ops.py
```diff
@@ -0,0 +1,875 @@
...
+    aten.var.correction: {f32},
+    aten.var_mean.correction: {f32},
+    aten.vdot.default: {f32},
+    aten.view_as_complex.default: {f32},
+    aten.view_as_real.default: {f32},
+    aten.where.self: {f32},
+    aten.xlogy.Tensor: {f32},
+    aten.zero_.default: {f32},
+    aten.zeros.default: {f32},
+    aten.zeros_like.default: {f32},
```
fduwjj, line 721: Out of curiosity, how did you generate this long list?

test/spmd/tensor/test_dtensor_ops.py
```diff
@@ -0,0 +1,875 @@
...
+    @property
+    def world_size(self) -> int:
+        return 4
+
+    # only allow float dytpe for now, we can relax this constraint
+    # when feel necessary later (i.e when adding quantization support).
+    @unittest.skipIf(TEST_WITH_ASAN, "Skipped under ASAN")
+    @skipIfCrossRef
+    @suppress_warnings
+    @ops(common_ops.op_db, allowed_dtypes=(torch.float,))
```
fduwjj, line 810: Do they regularly sync this list or it's automatically generated?

fduwjj (commented):
Nice!

test/spmd/tensor/test_dtensor_ops.py
wanchaol: No i mean this environment variable could be used to generate the failed list and we could copy them 1 by 1, normally we don't set this environment variable, but I leave it here in case in the future we want to generate failed list again

test/spmd/tensor/test_dtensor_ops.py
wanchaol: use the `COLLECT_EXPECT` to generate, actually i did some small trick to do fake run of all ops and write to a file directly

test/spmd/tensor/test_dtensor_ops.py
wanchaol: it's op db from pytorch itself, so operators stored in pytorch tests will be synced regularly if pytorch op tests get updated.

spmd/test/_utils.py
```diff
@@ -75,3 +90,162 @@ def wrapper(
...
+    def successful(self) -> bool:
+        return self.hit > 0 and self.miss == 0
+
+    def is_supported_tensor(self, t: torch.Tensor) -> bool:
+        # TODO: dist tensor need to support quantized and sparse
+        # tensors, quantized tensor might be relatively easy, but
+        # sparse tensor have special layouts that we need to possibly
+        # deal with, until we are clear about them, we don't officially
+        # support them.
+        return not any(
```
aazzolini, line 138: Interesting -- is there a reason why some of these tensors wouldn't work naturally? For example neg, conj, lazy or meta should just work -- did you try it out and it doesn't?

spmd/test/_utils.py
```diff
@@ -75,3 +90,162 @@ def wrapper(
...
+                # We need a way to test if a tensor is batched but there
+                # is no official APi to do it
+                # torch._C._is_batched(t),
+            ]
+        )
+
+    def gen_sharding_choices_for_arg(
+        self, arg: torch.Tensor
+    ) -> Sequence[Placement]:
+        # NOTE we assume cube mesh here to simplify things
```
aazzolini, line 158: Why is this / how much more complex would it be to add support to "rectangular meshes"? Some of the issues that I caught in my diff were only there once i had different mesh size for different dimensions (e.g. when using narrow to redistribute etc), wonder if we'll miss some stuff here.

spmd/test/_utils.py
```diff
@@ -75,3 +90,162 @@ def wrapper(
...
+            sharding_choices = sharding_choices * mesh_size
+        else:
+            # only generating choices with: replicate, or sharding
+            # evenly on a dimension that could be sharded
+            sharding_choices = sharding_choices + [
+                Shard(i)
+                for i, s, in enumerate(arg.shape)
+                if s > 1 and s % mesh_size == 0
+            ]
+        # TODO: add multi mesh choices
```
aazzolini, line 173: Is there a specific reason you're not enabling it? Is it becuase it woudl make the tests too slow?

spmd/test/_utils.py
```diff
@@ -75,3 +90,162 @@ def wrapper(
...
+                else:
+                    r = distribute_tensor(t, mesh, placements)
+                if type(t) is torch.nn.Parameter:
+                    r = torch.nn.Parameter(r, requires_grad=r.requires_grad)  # type: ignore
+                return r
+            else:
+                self.miss += 1
+                return t
+        elif torch.overrides.is_tensor_like(t):
+            # Blindly converting tensor subclasses to dist tensor can cause
```
aazzolini, line 243: Ideally we would just want to be able to have a DTensor of custom tensor, is this not allowed by design?

aazzolini (approved):
This is great! It will greatly streamline our operator implementation workload!

spmd/test/_utils.py
wanchaol: I didn't work actually, I want to explicitly forbidden them right now, until we find use cases of DTensor on such cases, I feel we should not promise too much on those different kind of tensors.

spmd/test/_utils.py
```diff
@@ -75,3 +90,162 @@ def wrapper(
...
+                t.is_mkldnn,
+                t.is_quantized,
+                t.is_nested,
+                torch._is_functional_tensor(t),
+                t.is_neg(),
+                t.is_conj(),
+                t.device.type in ("lazy", "meta"),
+                # We need a way to test if a tensor is batched but there
+                # is no official APi to do it
+                # torch._C._is_batched(t),
```
anj-s, line 151: What does it mean to "batch" a tensor?

spmd/test/_utils.py
wanchaol: Yep there're two reasons,  is that it would generate the combinations exponentially to the mesh.ndim, so it would make the tests become slow. Also the # GPUs in the CI likely be <=4, so the only multi-dim mesh we can support is a 2d mesh. I will adjust to 2d mesh once we confirm the GPUs we can get from CI.

spmd/test/_utils.py
wanchaol: it's a tensor coming from functorch that represents batching.

PR #402, opened 2022-08-24
[spmd] Migrate bmm and baddbmm ops to use Sharding Propagation rules instead

XilunWu:
Originally bmm and baddbmm ops were defined as custom operators to unblock MAT-ViT model training. This PR migrates them to adopt Sharding Propagation. 

TODO:
Fix the NaN problem in the test case for baddbmm when Beta=1. Currently this test case is simply skipped.


fduwjj (commented):
Thanks for working on this!

spmd/tensor/ops/tp_sharding_ops.py
```diff
@@ -4,40 +4,36 @@
...
-def dist_baddbmm(
-    self: DTensor,
-    batch1: DTensor,
-    batch2: DTensor,
-    beta: float = 1.0,
-    alpha: float = 1.0,
-) -> DTensor:
-    local_input, local_batch1, local_batch2 = pytree.tree_map(
-        unwrap_local_tensor, (self, batch1, batch2)
+@register_prop_rule("aten.bmm.default")
```
fduwjj, line 20: Can we move this op to maybe matrix_ops? So that this file only keeps the legacy code. So the ultimate goal is to remove this file completely.

spmd/tensor/ops/tp_sharding_ops.py
```diff
@@ -4,40 +4,36 @@
...
+def bmm_rules(op_schema: OpSchema) -> OutputSharding:
+    mat1_spec, mat2_spec = op_schema.args_spec
+    mat1_shape = list(mat1_spec.shape) # b * n * m
+    mat2_shape = list(mat2_spec.shape) # b * m * p
```
fduwjj, line 24: Sometimes, in backward, we will have a replicate as mat1 and a sharded as mat2. So can you also add logic to handle that. You can reference the code here: https://github.com/pytorch/PiPPy/pull/382/files#diff-44be2c8678cd9e83382ce3e5d09f972b06935f8980bc9410cce9cac0651552d0R143-R150. Thanks!

test/spmd/tensor/test_tp_sharding_ops.py
```diff
@@ -41,7 +41,8 @@ def test_sharded_transpose(self):
     # TODO: Need to investigate why test failed in CPU for baddbmm.
     @with_comms
     @skip_if_lt_x_gpu(TEST_GPU_NUM)
-    def test_sharded_baddbmm(self):
+    def test_sharded_baddbmm_beta_zero(self):
```
fduwjj, line 44: Can we merge the test of these two? Thanks!

fduwjj:
Also can we make linter happy? Thanks!

spmd/tensor/ops/tp_sharding_ops.py
```diff
@@ -4,40 +4,36 @@
...
+    return OutputSharding(
+        DTensorSpec(
+            mat1_spec.mesh,
+            mat1_spec.placements,
+            shape=list([mat1_shape[0], mat1_shape[1], mat2_shape[-1]]) # b * n * p
```
wanchaol, line 29: I actually don't think we need to specify the output shapes (the shape field is purely there used by inputs and sharding propagation rule, that's why it's optional), the output shapes should be auto calculated when generating the output DTensor, so the shape field can leave as empty.

spmd/tensor/ops/tp_sharding_ops.py
```diff
@@ -4,40 +4,36 @@
...
-    batch2: DTensor,
-    beta: float = 1.0,
-    alpha: float = 1.0,
-) -> DTensor:
-    local_input, local_batch1, local_batch2 = pytree.tree_map(
-        unwrap_local_tensor, (self, batch1, batch2)
+@register_prop_rule("aten.bmm.default")
+def bmm_rules(op_schema: OpSchema) -> OutputSharding:
+    mat1_spec, mat2_spec = op_schema.args_spec
+    mat1_shape = list(mat1_spec.shape) # b * n * m
```
wanchaol, line 23: are these rules only aim to support sharding on the batch dimension? bc i don't see any propagation base on the rule written here

test/spmd/tensor/test_tp_sharding_ops.py
XilunWu: Will it be good to move the tests into `test_matrix_ops.py` since these ops are now in `matrix_ops`?

spmd/tensor/ops/tp_sharding_ops.py
XilunWu: I guess this situation also applies to `baddbmm`, right?

spmd/tensor/ops/tp_sharding_ops.py
XilunWu: I see. So the original assignment `shape=mat1_spec.shape` is only to initialize `self.ndim` in `__post_init__` function? 

"the output shapes should be auto calculated when generating the output DTensor, so the shape field can leave as empty."
> That's sweet. Then assigning an `ndim` parameter seems more straightforward IMO because it implies that the output shape is auto calculated.

spmd/tensor/ops/tp_sharding_ops.py
XilunWu: According to this code:
https://github.com/pytorch/PiPPy/pull/382/files#diff-44be2c8678cd9e83382ce3e5d09f972b06935f8980bc9410cce9cac0651552d0R143-R150
I believe the current `bmm` only aims to support sharding on the batch dim. But I need to talk to @fduwjj to clarify on this.

> bc i don't see any propagation base on the rule written here
Right, my intention is to have default sharding prop on bmm (i.e. sharding on batch dim for both tensors, the result then has the same sharding). Or, I misunderstand the right approach to do it and my implementation won't do that?

wanchaol: I think it's fine to only support sharding on batch dim for `bmm`, at least initially. If so, could you check the input specs, and if the input specs not sharding on the batch dim, return `OutputSharding.output_spec` as None, and add failed reason to `OutputSharding.failed_reson`?

spmd/tensor/ops/tp_sharding_ops.py
wanchaol: Oh hmmm yeah it did get assigned in some ops. I guess we can also assign it in this case then. My point is that this shouldn't be the burden of op implementation (i.e. we shouldn't enforce it), but it's ok to add if it's easy to tell the output shape

XilunWu: I agree with you. By the way, could you share with a code pointer to where the shape information gets auto calculated? I did some search but did not find the place. Just to make sure that it happens somewhere.

spmd/tensor/ops/tp_sharding_ops.py
XilunWu: Thx! Will do it.  One question is, since Junjie wants to support the case where one matrix/tensor is replicated and the other is sharded, can I achieve it using DTensorSpec or do I have to manually redistribute the DTensor object like what's done here?
https://github.com/pytorch/PiPPy/pull/382/files#diff-44be2c8678cd9e83382ce3e5d09f972b06935f8980bc9410cce9cac0651552d0R146-R150

wanchaol: You don't need to manually do reshard, what you could actually do is, if input is replicate + sharded, then it's good to run it locally, but if input is not either `replicate + sharded` and `sharded + sharded`, you can set output.spec to None, and set schema suggestions to the suggestion that you are expecting. Then the operator dispatch could automatically do this for you https://github.com/pytorch/PiPPy/blob/main/spmd/tensor/dispatch.py#L92

spmd/tensor/ops/tp_sharding_ops.py
wanchaol: It's getting recovered during DTensor construction https://github.com/pytorch/PiPPy/blob/main/spmd/tensor/api.py#L147

aazzolini: Just to note that for tracing mode, we won't have that -- but we should be able to use other types of shape propagation. @wanchaol  do we have a plan for that? Would meta tensors work here?

aazzolini (commented):
Could we just implement these propagation rules directly as function fo the more general einsum rule? All of these ops are very easily representable as einsum.

I have a few examples that i can pull shortly.

aazzolini:
https://github.com/pytorch/PiPPy/pull/324/commits/bce9b65f9d61dfcef0aa2dd53e46e0f350cc6dcd#diff-91de4f53c681639f07b41ecf70e095b7a60e9d78e4d46b2f7888c904174c170c

^^ In the file above I started implementing a lot of linalg related ops as function of einsum + bias -- , but at that time we didnt' ahve prop_einsum ready , i think now we do since wanchao landed it , so maybe we could do it here?

wanchaol (changes requested):
Since this bmm/baddmm rule is requiring inputs that is not only sharded on the batched dimension, we need to make it more general here. 

Also suggested by @aazzolini, we should use einsum propagation here instead of manually writing rules to avoid all the special cases. Specifically, we should register the `bmm` by calling into this rule https://github.com/pytorch/PiPPy/blob/main/spmd/tensor/ops/math_ops.py#L21, 
 and `baddmm` rule should also call into this + a pointwise rule.

spmd/tensor/ops/tp_sharding_ops.py
wanchaol: @aazzolini I think for tracing mode we should use shape prop, and it should happen before the sharding prop here, because otherwise the view op propagation would fail, and I think we shouldn't worry about recovering the shape in sharding prop, because by the time we enter sharding prop, we already finish sharding prop for a graph with meta tensor

spmd/tensor/ops/tp_sharding_ops.py
fduwjj: Well, theoretically yes, but the backward of baddbmm is bmm. We can add this for baddbmm. Thanks!

test/spmd/tensor/test_tp_sharding_ops.py
fduwjj: Sounds good to me.

XilunWu:
Made mistakes in rebase. Fixing it....

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +129,52 @@ def test_t(self):
...
+        # If beta is 0, input tensor will be ignored
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        torch.manual_seed(self.rank)
+        tensor = torch.rand(3, 5, 6, device=self.device_type)
+        batch_1 = torch.rand(3, 5, 8, device=self.device_type)
+        batch_2 = torch.rand(3, 8, 6, device=self.device_type)
+        local_result = torch.baddbmm(
+            tensor, batch_1, batch_2, beta=0.0, alpha=0.5
+        )
+        sharding = [Shard(0)]
```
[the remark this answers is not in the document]
XilunWu: no problem. working on add something similar to https://github.com/pytorch/tau/blob/c4e9d3fc57bd2a624be68da13394fe4b6a9559be/test/spmd/tensor/test_matrix_ops.py#L83 in bmm/baddbmm tests.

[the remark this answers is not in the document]
XilunWu: > Since this bmm/baddmm rule is requiring inputs that is not only sharded on the batched dimension, we need to make it more general here.
> 
> Also suggested by @aazzolini, we should use einsum propagation here instead of manually writing rules to avoid all the special cases. Specifically, we should register the `bmm` by calling into this rule https://github.com/pytorch/PiPPy/blob/main/spmd/tensor/ops/math_ops.py#L21, and `baddmm` rule should also call into this + a pointwise rule.

For 2 tensors of shape "bmk" and "bkn", sharding on dim other than the batched dimension now gives "Two different input dims are sharded across the same mesh dim!" exception which I believe is justified because such a behavior (i.e. sharding on other dimensions) does not make sense. Do i understand correctly?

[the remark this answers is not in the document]
XilunWu: Fixing tests.

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +129,111 @@ def test_t(self):
...
+        )
+
+        test_comb(
+            shard2_spec, shard2_spec, shard2_spec,
+            beta=0.0, alpha=0.5, sharding_dim=0
+        )
+
+        # sharding on non-batching dim raises exception:
+        # Two different input dims cannot sharded across the same mesh dim
+        with self.assertRaises(Exception):
```
wanchaol, line 178: shall we group all those checks together? I guess sth like:
tensor_spec = [shard0/1/2]
batch1_spec = [shard0/1/2]
batch2_spec = [shard0/1/2]

combs = list(itertools.product(tensor_spec, batch1_spec, batch2_spec))
then remove the ones that get passed already (or have a pass list), and assert the rest of them having exceptions?

wanchaol (commented):
Looks great! Have a few more suggestions

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +129,111 @@ def test_t(self):
         self.assertEqual(tranposed_mat2.size(), torch.Size([12, 8]))
         self.assertEqual(tranposed_mat2.placements, shard_spec)
 
+    # TODO: Need to investigate why test failed in CPU for baddbmm.
```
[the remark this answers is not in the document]
XilunWu: As far as I recall, it's NaN in tensor that causes the test to fail. NaN only appear in DTensor not local tensor.

test/spmd/tensor/test_matrix_ops.py
wanchaol: btw you could just use DTensorConverter to automatically generate all combinations for every argument https://github.com/pytorch/tau/blob/c4158aa580f982517277e21e749f549d100c8952/spmd/test/common_utils.py#L96

XilunWu: > shall we group all those checks together? I guess sth like: tensor_spec = [shard0/1/2] batch1_spec = [shard0/1/2] batch2_spec = [shard0/1/2]
> 
> combs = list(itertools.product(tensor_spec, batch1_spec, batch2_spec)) then remove the ones that get passed already (or have a pass list), and assert the rest of them having exceptions?

This part for `bmm` is done. Next is to add for `baddbmm` and test on GPU.



> btw you could just use DTensorConverter to automatically generate all combinations for every argument
> 
> https://github.com/pytorch/tau/blob/c4158aa580f982517277e21e749f549d100c8952/spmd/test/common_utils.py#L96

Need to see how to use it. But my current implementation should be able to port to 3 operands case (i.e. `baddbmm`) easily.

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +131,150 @@ def test_t(self):
...
+        shard1_spec = Shard(1)
+        shard2_spec = Shard(2)
+        shard_specs = [shard0_spec, shard1_spec, shard2_spec]
+        shard_specs_comb = itertools.product(shard_specs, shard_specs)
+        passlist = [
+            [shard0_spec, shard0_spec],
+            [shard2_spec, shard1_spec],
+        ]
+
+        def placementEqual(
```
wanchaol, line 237: why we need this helper? it should be redundant I think, we can just use `p1 == p2` to check this

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +131,150 @@ def test_t(self):
...
+            if (p1.is_shard() and p2.is_shard()):
+                return p1.dim == p2.dim
+            elif (p1.is_replicate() and p2.is_replicate()):
+                return True
+            elif (p1.is_partial() and p2.is_partial()):
+                return True
+            else:
+                return False
+
+        def specEqual(
```
wanchaol, line 248: same for this, we can just do `s1 == s2`

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +131,150 @@ def test_t(self):
...
+            s1: Sequence[Placement], s2: Sequence[Placement]) -> bool:
+            if (len(s1) != len(s2)):
+                return False
+            else:
+                boolSeq = map(lambda t: placementEqual(t[0], t[1]), zip(s1, s2))
+                return functools.reduce(lambda x, y: x and y, boolSeq, True)
+
+        def specInSeq(
```
wanchaol, line 256: ditto, placement dataclass could automatically check if they are in the sequence

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +131,150 @@ def test_t(self):
...
+
+        assertionErrorFormat = \
+            "Test failed! AssertionError: Result tensors not equal! Spec: {0} {1}."
+        runtimeErrorFormat = \
+            "Test failed! RuntimeError: \"{2}\"! Spec: {0} {1}."
+        shard0_spec = Shard(0)
+        shard1_spec = Shard(1)
+        shard2_spec = Shard(2)
+        shard_specs = [shard0_spec, shard1_spec, shard2_spec]
+        shard_specs_comb = itertools.product(shard_specs, shard_specs)
```
wanchaol, line 231: you will need to convert `shard_specs_comb` to a list, if not, you can only loop it once.

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +131,150 @@ def test_t(self):
...
+            return functools.reduce(lambda x, y: x or y, boolSeq, False)
+
+        # tests that currently pass
+        for spec in passlist:
+            test_placement_comb([spec[0]], [spec[1]])
+
+        # TODO: support these tests
+        shard_specs_comb = [spec for spec in shard_specs_comb if not specInSeq(spec, passlist)]
+        for spec in shard_specs_comb:
+            try:
```
wanchaol, line 268: can we just do `self.assertRaise(Exception)` instead of print the exception to stdout? that way we can capture the failure in test when it silently pass later.

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +131,150 @@ def test_t(self):
...
+                test_placement_comb([spec[0]], [spec[1]])
+            except AssertionError:
+                print(assertionErrorFormat.format(spec[0], spec[1]))
+            except RuntimeError as e:
+                print(runtimeErrorFormat.format(spec[0], spec[1], str(e)))
+            else:
+                print(f"Test passed! Spec: {spec[0]} {spec[1]}")
+
+        # TODO: add more tests - what cases do we want to support?
```
wanchaol, line 277: I think this line can be deleted since now you product all combs?

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +131,150 @@ def test_t(self):
...
+            beta=0.0, alpha=0.5
+        )
+
+        with self.assertRaises(Exception):
+            test_comb(
+                shard1_spec, shard1_spec, shard1_spec,
+                beta=0.8, alpha=0.5
+            )
+
+        with self.assertRaises(Exception):
```
wanchaol, line 184: can you also make the `test_sharded_baddbmm` follow the same logic in `test_sharded_bmm`?

test/spmd/tensor/test_matrix_ops.py
XilunWu: ahh interesting. the `==` operator doesn't work on `Sequence` but works on `list`! Thx!

test/spmd/tensor/test_matrix_ops.py
XilunWu: Yes and no. My intention is to hint on bringing "replicate" into the combination in the future.

wanchaol (approved):
looking good, thanks for addressing the comments!

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +131,130 @@ def test_t(self):
...
+        torch.manual_seed(self.rank)
+        input = torch.rand(3, 5, 8, device=self.device_type)
+        mat_2 = torch.rand(3, 8, 6, device=self.device_type)
+        local_result = torch.bmm(input, mat_2)
+
+        def test_placement_comb(
+            placements1: Sequence[Placement],
+            placements2: Sequence[Placement],
+        ) -> None:
+            input_dt = DTensor.from_local(input, device_mesh, placements1)
```
wanchaol, line 223: I think your value inconsistency problem coming from the fact that you are using `from_local` and for the case [2, 1] you bmm it across local tensor boundaries, so i think you shouldn't use `from_local` here, can you just follow the `test_mm` and use `distribute_tensor` instead?
https://github.com/pytorch/tau/blob/main/test/spmd/tensor/test_matrix_ops.py#L86

test/spmd/tensor/test_matrix_ops.py
wanchaol: any reason we still have this? we should delete this function in `test_sharded_bmm` and `test_sharded_baddbmm` because we are actually using list not sequence.

test/spmd/tensor/test_matrix_ops.py
```diff
@@ -127,6 +130,119 @@ def test_t(self):
...
+            tensor_dt = distribute_tensor(
+                tensor, device_mesh, tensor_placements
+            )
+            batch_1_dt = distribute_tensor(
+                batch_1, device_mesh, batch_1_placements
+            )
+            batch_2_dt = distribute_tensor(
+                batch_2, device_mesh, batch_2_placements
+            )
+            new_dt = cast(
```
fduwjj, line 158: Not needed in this PR, but we also need a backward test for baddbmm and bmm since both are used in ViT.

PR #413, opened 2022-08-28
[DistributedTensor] Add support for torch.mul

anj-s:
Add support for `torch.mul` ops for DistributedTensor. 

Follow up questions(please see TODOs as well):
1. For scalar input, are users expected to convert inputs into DistributedTensors with placement specs?
2. Should we implicitly convert all args and kwargs to DistributedTensors with placements specs? Currently we assume that all input are DTs with placement specs and I did not see a test checking for other cases. 
3. Should all ops propose potential autoboxing rules incase sharding propagation are not available for a given input placement spec? For example, if mat1 is a _Partial and mat2 is Replicate for torch.mul

NOTE: Will fix failing test post the followup discussion since it affects the fix.


spmd/tensor/dispatch.py
```diff
@@ -132,6 +132,9 @@ def operator_dispatch(
     args_schema = tree_map(unwrap_schema, args)
     kwargs_schema = tree_map(unwrap_schema, kwargs)
 
+    # TODO(anj): Check if the input to ops are all DTs with placements specs.
```
wanchaol, line 135: this check has already been added here https://github.com/pytorch/tau/blob/main/spmd/tensor/api.py#L217

spmd/tensor/ops/matrix_ops.py
```diff
@@ -37,19 +38,28 @@ def mm_prop(
...
 
 
-@register_prop_rule("aten.mm.default")
 def mm_rules(op_schema: OpSchema) -> OutputSharding:
-    mat1_spec, mat2_spec = op_schema.args_spec
+    mat1_spec, mat2_spec = op_schema.args_schema
     return OutputSharding(mm_prop(mat1_spec, mat2_spec))
 
 
+default_mm_ops = ["aten.mm.default", "aten.mul.Tensor"]
```
wanchaol, line 53: why we are changing matrix ops for torch.mul? `torch.mul` is not a matrix op, we shouldn't change this file.

wanchaol (changes requested):
`torch.mul` is a pointwise op, not a matrix op, we should add it to the pointwise op list instead of adding to matrix ops.

wanchaol:
re 1: we don't need to convert scalar inputs, only tensor inputs
re 3. in general this should be done, but before we do it we probably need to categorize and define autoboxing rule for each operator (i.e. even for pointwise op, there might be some accept partial, some not)

spmd/tensor/ops/matrix_ops.py
anj-s: Clarified offline: Mistook the doc example that had a MM output. Updated to reflect pointwise.

anj-s:
Updated the PR but in light of your latest PR, should we abandon this?

spmd/tensor/ops/pointwise_ops.py
```diff
@@ -124,6 +124,7 @@
     "aten.add.out",
     "aten.add_.Scalar",
     "aten.mul_.Scalar",
+    "aten.mul.Tensor",
```
wanchaol, line 127: can we also add the `aten.mul.out` variant?

test/spmd/tensor/test_pointwise_ops.py
```diff
@@ -90,6 +90,35 @@ def test_dropout_errors(self):
...
 
+    @with_comms
+    def test_mul(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        torch.manual_seed(self.rank)
+        spec = [Replicate()]
+        input_tensor = torch.randn(
+            (2, 4), device=self.device_type, requires_grad=True
+        )
+        dtensor = DTensor(
```
wanchaol, line 101: nit: maybe try using the `distribute_tensor` API directly? `DTensor(tensor)` API maybe a bit hard to use as we always need to pass the `requires_grad` field.

test/spmd/tensor/test_pointwise_ops.py
```diff
@@ -90,6 +90,35 @@ def test_dropout_errors(self):
                 torch.nn.functional.dropout,
             )
 
+    @with_comms
+    def test_mul(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        torch.manual_seed(self.rank)
+        spec = [Replicate()]
```
wanchaol, line 97: can we test both replicate and shard?

test/spmd/tensor/test_pointwise_ops.py
```diff
@@ -90,6 +90,35 @@ def test_dropout_errors(self):
                 torch.nn.functional.dropout,
             )
 
+    @with_comms
+    def test_mul(self):
```
wanchaol, line 94: maybe this test can mainly testing the `aten.mm.out` (`torch.mul(...out=)`)? you can fix the CI by deleting the xfail in the `test_dtensor_ops.py`, which testing torch.mul, but we still need to test out variant.

spmd/tensor/ops/matrix_ops.py
```diff
@@ -37,19 +38,20 @@ def mm_prop(
...
 def mm_rules(op_schema: OpSchema) -> OutputSharding:
-    mat1_spec, mat2_spec = op_schema.args_spec
+    mat1_spec, mat2_spec = op_schema.args_schema
     return OutputSharding(mm_prop(mat1_spec, mat2_spec))
 
-
 @register_prop_rule("aten.addmm.default")
 def addmm_rules(op_schema: OpSchema) -> OutputSharding:
-    input_spec, mat1_spec, mat2_spec = op_schema.args_spec
+    input_spec, mat1_spec, mat2_spec = op_schema.args_schema
```
wanchaol, line 54: not sure if changing it to `args_schema` triggered the CI failures? i remembered addmm have some scalar inputs (alpha, beta), probably that's in the kwargs instead?

wanchaol (commented):
Thanks for addressing comments! Looking good, left a few suggestions.

> Updated the PR but in light of your latest PR, should we abandon this?

The PR I updated fixing the CI with side effects of adding `aten.mul`, but I didn't test them thoroughly (i.e. no `.out` variant testing, only with the op db testing), maybe we can land this PR first and I can rebase on this change?

test/spmd/tensor/test_pointwise_ops.py
```diff
@@ -90,6 +90,37 @@ def test_dropout_errors(self):
...
+            (8, 4), device=self.device_type, requires_grad=True
+        )
+        dtensor = DTensor.from_local(input_tensor, device_mesh, spec)
+        # Mutiplication with a scalar.
+        dt = torch.mul(dtensor, 8.0)
+        expected = torch.mul(input_tensor, 8.0)
+        self.assertEqual(input_tensor, dtensor.to_local())
+        self.assertEqual(expected, dt.to_local())
+
+        # Mutiplication with a matrix.
```
wanchaol, line 108: can we group the "Mutiplication with a scalar" and "Mutiplication with a matrix" together?
i.e. 
```
for rhs in [8.0, other_dtensor]:
    res = torch.mul(dtensor, rhs)
    assertEqual(...
```

test/spmd/tensor/test_pointwise_ops.py
```diff
@@ -90,6 +90,37 @@ def test_dropout_errors(self):
                 torch.nn.functional.dropout,
             )
 
+    @with_comms
+    def test_mul(self):
```
wanchaol, line 94: nit: rename this test case to `torch_mul_out`?

test/spmd/tensor/test_pointwise_ops.py
anj-s: I am testing Shard instead of Replicate since that is more relevant for the out variant.

test/spmd/tensor/test_pointwise_ops.py
anj-s: I removed the scalar test since we are already testing torch.mul and that should be covered.

wanchaol (approved):
looks great, thanks for addressing comments!

PR #415, opened 2022-08-29
[spmd] by default disable the fallback mode

wanchaol:
fallback mode is error prone and sometimes hard to track the errors, so in this PR we disable it by default and enable the necessary ops as a side effect.


anj-s (approved):

spmd/tensor/dispatch.py
```diff
@@ -22,7 +22,7 @@
 """
 If set to true, __DEBUG_STRICT will fail when an op doesn't have a sharding rule registered.
 """
-_DEBUG_STRICT = False
+_DEBUG_STRICT = True
```
anj-s, line 25: Should we remove this?

spmd/tensor/ops/pointwise_ops.py
```diff
@@ -118,12 +118,30 @@
...
+    "aten.add_.Scalar",
+    "aten.add.out",
+    "aten.abs.default",
+    "aten.bitwise_and.Tensor",
+    "aten.bitwise_and_.Tensor",
+    "aten.bitwise_or_.Tensor",
+    "aten.eq.Tensor",
     "aten.gelu.default",
+    "aten.relu.default",
+    "aten.le.Tensor",
```
fduwjj, line 135: Can we also add relu backward which is needed for ViT.

https://github.com/pytorch/tau/pull/382/files#diff-a6184a94d36719b27f3b3b5c3c4efd24261d7dc8830ed45e96eac330dbed5824R124

Thanks!

spmd/tensor/ops/tensor_ops.py
```diff
@@ -15,10 +15,13 @@ def default_prop_rule(op_schema: OpSchema) -> OutputSharding:
...
 default_prop_ops = [
+    "aten._local_scalar_dense.default",
+    "aten.clone.default",
+    "aten.copy_.default",
+    "aten.detach.default",
     "aten.is_same_size.default",
     "aten.ones_like.default",
-    "aten.detach.default",
-    "aten.clone.default",
+    "aten.new_empty_strided.default",
```
fduwjj, line 24: Can you also add "aten._to_copy.default",?

As in https://github.com/pytorch/tau/pull/382/files#diff-34b9e654af2d5e93ed673c8e416fe783e49d1e8a81b443f8d675725649429a77R22

fduwjj (approved):
Nice!! Thanks!!

spmd/tensor/dispatch.py
wanchaol: yeah I think we can remove it, but i guess we will have a similar flag later to enable the fallback of redistribute to replicate, let me change the name of this.

spmd/tensor/ops/pointwise_ops.py
wanchaol: Yeah sure we can add it in the following PRs, would like to keep this PR succinct to disable fallback mode.

spmd/tensor/dispatch.py
wanchaol: gonna merge this PR first, we can use this https://github.com/pytorch/tau/pull/413 to increase test coverage

PR #419, opened 2022-08-30
[spmd] fully implement distribute_module API

wanchaol:
This PR fully implements distribute_module API, the old API was only
used by prototype and does not have finer control on how to shard
each params/inputs.

This PR fully implements it by:
1. move distribute_tensor to `tensor` package
2. add partition_fn/input_fn/output_fn to distribute_module
3. restructure the api tests to test this API


spmd/__init__.py
```diff
@@ -1,106 +1,96 @@
...
 
-
-def distribute_tensor(
-    tensor: torch.Tensor,
+def distribute_module(
+    module: nn.Module,
     device_mesh: Optional[DeviceMesh] = None,
-    placements: Optional[Sequence[Placement]] = None,
-) -> DTensor:
+    partition_fn: Optional[Callable[[str, nn.Module], None]] = None,
```
wanchaol, line 14: API discussion: should we make `partition_fn`, `input_fn` and `output_fn` kw-only arguments?

test/spmd/tensor/test_api.py
```diff
@@ -0,0 +1,175 @@
...
+    def test_distribute_tensor(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        shard_spec = [Shard(0)]
+        tensor_to_shard = torch.randn(12, 3)
+        sharded_tensor = distribute_tensor(
+            tensor_to_shard, device_mesh, shard_spec
+        )
+        self.assertEqual(sharded_tensor.size(), torch.Size([12, 3]))
+        local_tensor = sharded_tensor.to_local()
+        self.assertEqual(local_tensor.size(), torch.Size([3, 3]))
```
anj-s, line 45: Should we condition on the size of the device mesh? Looks like this assumes 4 GPUs?

test/spmd/tensor/test_api.py
```diff
@@ -0,0 +1,175 @@
...
+            m.reset_parameters()
+
+
+class DTensorAPITest(DistTensorTestBase):
+    @with_comms
+    def test_distribute_tensor(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        shard_spec = [Shard(0)]
+        tensor_to_shard = torch.randn(12, 3)
+        sharded_tensor = distribute_tensor(
```
anj-s, line 40: nit: dist_tensor or dtensor instead of sharded_tensor?

test/spmd/tensor/test_api.py
```diff
@@ -0,0 +1,175 @@
...
+
+    @with_comms
+    def test_distribute_tensor_requires_grad(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        shard_spec = [Shard(0)]
+        tensor_to_shard = torch.randn(12, 3, requires_grad=True)
+        sharded_tensor = distribute_tensor(
+            tensor_to_shard, device_mesh, shard_spec
+        )
+        self.assertTrue(sharded_tensor.requires_grad)
```
anj-s, line 55: Can we bundle this test with the previous one?

test/spmd/tensor/test_api.py
```diff
@@ -0,0 +1,175 @@
...
+        replica_model = distribute_module(
+            model,
+            device_mesh,
+            input_fn=replicate_input_fn,
+        )
+        input = torch.randn(10, 10, requires_grad=True)
+        output = replica_model(input)
+        output.sum().backward()
+        param_grad = list(replica_model.parameters())[0].grad
+        self.assertTrue(isinstance(param_grad, DTensor))
```
anj-s, line 170: Should we add a test with a loss function and optimizer step as well?

spmd/__init__.py
```diff
@@ -1,106 +1,96 @@
...
+        device_mesh (:class:`DeviceMesh`): the device mesh to place the module.
+        partition_fn (Callable): the function to partition parameters (i.e. shard certain
+            parameters across the `device_mesh`). If `partition_fn` is not specified,
+            by default we replicate all module parameters of `module` across the mesh.
+        input_fn (Callable): specify the input distribution, i.e. could control how the
+            input of the module is sharded. `input_fn` will be installed as a module
+            forward_pre_hook.
+        output_fn (Callable): specify the output distribution, i.e. could control how the
+            output is sharded, or convert it back to torch.Tensor. output_fn will be
+            installed as a module forward_hook.
```
anj-s, line 34: Should this be post forward hook?

spmd/__init__.py
```diff
@@ -1,106 +1,96 @@
...
 
-    Returns:
-        A :class:`DTensor` object
+    Return:
```
anj-s, line 36: s/Return/Returns?

spmd/__init__.py
```diff
@@ -1,106 +1,96 @@
...
-    TODO: add a more flexible tagging, i.e. convert
-    certain param to a certain spec, like a PlacementPlan
-    """
-
-    # pyre-fixme[3]: Return type must be annotated.
-    # pyre-fixme[2]: Parameter must be annotated.
-    def to_dist_tensor(t):
-        if isinstance(t, nn.Parameter):
-            return distribute_tensor(t.data, device_mesh, spec)
+    # this function loop over the whole module parameters
```
anj-s, line 50: Can this be a function docstring?

spmd/__init__.py
```diff
@@ -1,106 +1,96 @@
...
+    # and buffers, replicate all non DTensor params/buffers
+    # to DTensor parameters/buffers
+    def replicate_module_params_buffers(t: torch.Tensor) -> torch.Tensor:
+        if isinstance(t, torch.Tensor) and not isinstance(t, DTensor):
+            # replicate the tensor if it has not been converted yet.
+            assert device_mesh is not None
+            return distribute_tensor(
+                t, device_mesh, [Replicate()] * device_mesh.ndim
+            )
         else:
```
anj-s, line 60: nit: else is redundant

spmd/__init__.py
anj-s: I guess its call forward_hook but was trying to see if we could clarify when it ran.

spmd/tensor/__init__.py
```diff
@@ -1,7 +1,82 @@
...
+    device_mesh = (
+        get_global_device_mesh() if device_mesh is None else device_mesh
+    )
+    # convert tensor to the correponding device type if it's not in that device type
+    tensor = tensor.to(device_mesh.device_type)
+    # set default placements to replicated if not specified
+    if placements is None:
+        placements = [Replicate() for _ in range(device_mesh.ndim)]
+
+    # distribute the tensor according to PlacementSpec
```
anj-s, line 47: If each local tensor contributes to form a DTensor, what is the use case for calling scatter and broadcast?

spmd/__init__.py
anj-s: Do you see potentially more args added to this API?

anj-s (approved):
Thank you for the PR! Small questions but otherwise it looks good.
We can do an API review in our next meeting to clarify the kwargs. Shall we also update the RFC?

test/spmd/tensor/test_megatron_example.py
```diff
@@ -58,23 +41,56 @@ def shard_module(m, device_type):
...
-        distribute_tensor(m.net2.bias, device_mesh, replicate)
-    )
-    m = _replicate_input_tensor(m, device_mesh, replicate)
-    m.net2 = _aggregate_local_tensor(m.net2)
-    m.net1.weight.register_hook(
-        functools.partial(_gradient_hook, m.net1.weight)
+
+    def shard_params(name, module):
+        if isinstance(module, nn.Linear):
+            if name == "net1":
```
fduwjj, line 47: Hmm, this is kind of hacky, right? Can we have something like a spec or json input as input like shard_module? I mean basically, can we make this shard_params a util function available for all users not just an internal api inside the test?

fduwjj (commented):
Thanks for making this happening.. Btw, Why Github does not support Macro.. I really miss the meme in Phabricator.. Anyway,

test/spmd/tensor/test_api.py
```diff
@@ -0,0 +1,175 @@
...
+        module_to_replicate = MyModel(20, 20, device=self.device_type)
+        replica_module = distribute_module(
+            module_to_replicate, device_mesh, replicate_fn
+        )
+        for param in replica_module.parameters():
+            self.assertIsInstance(param, DTensor)
+            self.assertEqual(param.placements, replica_spec)
+
+        # only shard part of module, and rest of module should be replicate
+        def shard_fn(name, module):
```
fduwjj, line 107: Ditto, I agree that it's good that we can let users to write whatever shard_fn to have some flexibility. But for cases we know like two-layer linear Megatron style, can we have a more simplified way to shard it with some default util APIs?

wanchaol: I think we should provide pre-set shard fns, like you said megatron linear should just be a pre-set function that user can select from, and inside the preset function it could search the model hierarchy and shard the two linears correspondingly.

test/spmd/tensor/test_megatron_example.py
wanchaol: I don't think `shard_module` api provide the flexibility of doing anything we want, for example the gradient hook you need to install for all module params, there might be other things advanced user would like to control, providing a partition_fn could make it more flexible, if we make it a spec or json, it's tight to each module architecture and user need to modify it whenever the module architecture changes. To provide the ease of use, we should provide a bunch of default sharding strategies (pre-set sharding funcs) to let user directly use them instead of writing their own. Does this make sense to you?

spmd/tensor/__init__.py
wanchaol: The only way we allow user to contribute a local tensor to DT is `from_local`, we don't allow `DTensor` constructor directly. If there're existing parameters, we need to shard them and do scatter/broadcast

spmd/__init__.py
wanchaol: Currently I don't see there will be newly added arguments, the function arguments should be flexible enough to do anything user want. But we should run this with a larger group and see if we need to add new things in the future.

spmd/__init__.py
wanchaol: yeah sure I'll update the comment to clarify

spmd/__init__.py
wanchaol: This is a function we used internally in `distribute_module`, I guess it does not need a docstring unless we want to make this a preset function that's visible to the user?

spmd/__init__.py
wanchaol: this is actually needed for `module._apply` where it requires you return the same tensor if not modified.

wanchaol: Let me simplify a bit about this logic, I guess we might not need `module._apply`

test/spmd/tensor/test_api.py
wanchaol: This is actually already been added in `test_megatron_example.py`

wanchaol:
GPU CI is broken, landing it first.

spmd/__init__.py
```diff
@@ -1,103 +1,86 @@
...
+        # buffers, replicate all non DTensor params/buffers to DTensor
+        # parameters/buffers, if they have not been partitioned in the
+        # partition_fn, we can't easily use `module._apply` here
+        # because we don't know what happened inside partition_fn as
+        # user could do anything, i.e. install hooks, and we want to
+        # preserve those.
+        full_replicate = [Replicate()] * mesh.ndim
+        for key, param in m._parameters.items():
+            if param is not None and not isinstance(param, DTensor):
+                submod.register_parameter(
```
XilunWu, line 54: it surprises me that `submod` is actually in scope. Can someone explain this to me because I thought this should be `m` not `submod` but the code passed tests and ran without a problem. @wanchaol 

same question on line 62.

spmd/__init__.py
```diff
@@ -1,103 +1,86 @@
...
-    tensor: torch.Tensor,
+def distribute_module(
+    module: nn.Module,
     device_mesh: Optional[DeviceMesh] = None,
-    placements: Optional[Sequence[Placement]] = None,
-) -> DTensor:
+    partition_fn: Optional[Callable[[str, nn.Module], None]] = None,
+    input_fn: Optional[Callable[..., None]] = None,
+    output_fn: Optional[Callable[..., None]] = None,
+) -> nn.Module:
```
XilunWu, line 17: Another question is, I believe we are doing in-place modification to `module` then why do we still return this module? It can be confusing if `module = distribute_module(module,...)` and `distribute_module(module, ...)` are equivalent in my opinion.

spmd/__init__.py
wanchaol: Yeah I think it should be `m` actually, the reason that this get passed in tests is because python eval values dynamically, when this function get called, `submod` already in the scope and the `submod` is actually equivalent to `m`

spmd/__init__.py
wanchaol: yes it's inplace modification, we don't need to return it if all people calling this API knows that it's a inplace  modification, but I think users won't care about too much details about how this function is implemented, what they care really is to call this function, and it returns a module that can be used for distributed training, I think this actually make a better UX, and it also aligns with `distribute_tensor` API offers.

spmd/__init__.py
XilunWu: Yes, `submod` somehow becomes in scope which I still don't get how. I tried to write the same code in Python but couldn't get it run. Never mind, not a big deal. Can I create a PR to rename `submod` to `m` on line 54 and 62?

spmd/__init__.py
XilunWu: No problem. I felt a bit confused because I saw people using both `module = distribute_module(module,...)` and `distribute_module(module, ...)` in different tests in PR. Thx for the explanation!

PR #420, opened 2022-08-30
Enable NCCL in GPU docker CI

kwen2501:
Add shared memory settings to docker run
Ref: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/troubleshooting.html#sharing-data


fduwjj (approved):
Thanks! LGTM.

PR #421, opened 2022-08-31
[spmd] fix redistribute replicate to partial with subgroup

wanchaol:
This PR fixes redistribute replicate to partial case, where it should
use rank cooridinate instead of the global rank, so that it could
correctly zero out other ranks tensor.


test/spmd/tensor/test_redistribute.py
```diff
@@ -165,6 +165,27 @@ def test_replicate_to_partial(self):
...
+        # 1) test replicate -> partial on subgroup
+        replica_tensor = DTensor(
+            local_tensor, device_mesh, replica_spec, requires_grad=True
+        )
+        partial_tensor = Redistribute.apply(
+            replica_tensor, device_mesh, partial_spec
+        )
+        self.assertEqual(partial_tensor.size(), local_tensor.size())
+
+        if self.rank == 1:
```
fduwjj, line 180: What about when self.rank == 3?

fduwjj (commented):
Thanks for fixing this and I leave some questions for the test.

test/spmd/tensor/test_redistribute.py
wanchaol: for replicate to partial, we zero out all other ranks content, so in this case `self.rank == 3` should be zero tensor. But I guess for rank=1 and rank=2 we shouldn't zero out, let me see what's going on

spmd/tensor/redistribute.py
```diff
@@ -150,8 +150,8 @@ def redistribute_spmd_tensor(
             if current.is_replicate():
                 # For replicate -> partial, we zero out all other ranks of the current mesh dim
                 # and leave only 1 rank have the data, to perform a "zero cost" reshard.
-                my_rank = device_mesh.get_rank()
-                if my_rank != 0:
+                my_rank_on_mesh_dim = device_mesh.get_coordinate_on_dim(i)
+                if my_rank_on_mesh_dim is not None and my_rank_on_mesh_dim != 0:
                     new_local_tensor = local_tensor.zero_()
                 else:
                     new_local_tensor = local_tensor
```
yhcharles, line 157: How about:
```
if my_rank_on_mesh_dim == 0:
    new_local_tensor = local_tensor
else:
    new_local_tensor = local_tensor.zero_()
```

yhcharles:
Can we also enable `distribute_tensor()` to partial?

yhcharles (approved):
lgtm

spmd/tensor/redistribute.py
wanchaol: Unfortunately we can't do this, because when `my_rank_on_mesh_dim` is None, it also goes to the `else` branch and accidentally zero out the tensor, which it should not

wanchaol:
> Can we also enable `distribute_tensor()` to partial?

what does it mean to enable `distributed_tensor` with partial? I think from user facing API prospective, we shouldn't enable user to directly create sth like `partial`, partial should always be intermediate results that user won't be aware of.

wanchaol:
GPU CI is broken, landing it

PR #425, opened 2022-09-01
[spmd] move test file and rename to testing

wanchaol:
[spmd] move test file and rename to testing
    
test_checkpoint.py is in the wrong folder, move it over and rename it to spmd/testing to avoid confusion


wz337 (approved):
LGTM! Thanks for fixing the tests!

PR #431, opened 2022-09-04
[spmd] pump pt requirement version and fix CI

wanchaol:
This should fix the GPU CI problem with correctly updated nightly links


fduwjj:
Nice! Thanks for fixing!!

PR #432, opened 2022-09-05
DDP2Pipe connector example

pbelevich:


pippy/IR.py
```diff
@@ -491,7 +499,7 @@ def _number_and_count_forward_stages(gm : pippy.fx.GraphModule):
                 found_idxs.setdefault(node.meta['stage_idx'])
                 num_stages += 1
 
-        assert all(i in found_idxs for i in range(num_stages))
+        # assert all(i in found_idxs for i in range(num_stages))
```
pbelevich, line 502: @kwen2501 WDYT about this assert?

examples/ddp2pipe/.gitignore
```diff
@@ -0,0 +1 @@
+data
```
kwen2501, line 1: Is this a folder you used locally, or a folder generated by this example?

pippy/PipelineDriver.py
```diff
@@ -1433,6 +1435,17 @@ def forward(self, *args, **kwargs):
         if DEBUG:
             self._check_stages_cleanup()
```
kwen2501, line 1437: nit: (ignore if you need to land first)
When we have a chance, shall we add a comment here explaining the purpose of the code block below? Thanks!

examples/ddp2pipe/ddp2pipe.py
```diff
@@ -0,0 +1,299 @@
...
+            logits_chunks = None
+            loss = torch.empty(1)
+
+        logits = torch.empty((x.size(0), DIMS[-1]))
+        torch.distributed.scatter(logits, logits_chunks, src=driver_rank, group=self.pp_group)
+        torch.distributed.broadcast(loss, src=driver_rank, group=self.pp_group)
+
+        if torch.is_grad_enabled():
+            if rank == driver_rank:
+                last_grads_chunks = [x[0] for x in self.pipeline.last_grads]  # TODO
```
kwen2501, line 117: nit: maybe good to write out this TODO?
(ignore if you need to land first)

kwen2501 (approved):
LGTM! Thanks!

pippy/PipelineDriver.py
```diff
@@ -1433,6 +1435,17 @@ def forward(self, *args, **kwargs):
...
+            local_results = []
+            last_grads = []
+            for local_result in local_results_and_last_grads:
+                local_results.append(local_result[0])
+                last_grads.append(local_result[1])
+
+            self.last_grads = last_grads
+        else:
+            local_results = local_results_and_last_grads
+
```
kwen2501, line 1448: nit: (ignore if you need to land first)
When we have a chance, shall we add a comment here explaining the purpose of the code block below? Thanks!

examples/ddp2pipe/ddp2pipe.py
```diff
@@ -0,0 +1,299 @@
+import argparse
```
aazzolini, line 1: This example is great! Could you add a comment in the preamble explaining the flow?

examples/ddp2pipe/ddp2pipe.py
```diff
@@ -0,0 +1,299 @@
...
+        if rank == driver_rank:
+            self.pipe_model = PipeModel()
+
+            for i in range(1, 4):
+                annotate_split_points(self.pipe_model, {f'mod.{i}': PipeSplitWrapper.SplitPoint.BEGINNING})
+
+            # print(self.pipe_model)
+
+            self.pipe = Pipe.from_tracing(self.pipe_model, output_loss_value_spec=(False, True))
+            args_chunk_spec = (TensorChunkSpec(0), TensorChunkSpec(0))
```
aazzolini, line 69: Oh this is very similar to @wanchaol 's Placement, maybe we can merge eventually?

examples/ddp2pipe/ddp2pipe.py
```diff
@@ -0,0 +1,299 @@
...
+            kwargs_chunk_spec = {}
+            output_chunk_spec = (TensorChunkSpec(0), CustomReducer(torch.tensor(0.0), lambda a, b: a + b))
```
aazzolini, line 71: Question, is CustomReducer acting on the output of the forward pass or the backward pass?

examples/ddp2pipe/ddp2pipe.py
```diff
@@ -0,0 +1,299 @@
...
+        driver_rank = self.pp_ranks[0]
+
+        if rank == driver_rank:
+            x_list = [torch.empty_like(x) for _ in range(len(self.pp_ranks))]
+            labels_list = [torch.empty_like(labels) for _ in range(len(self.pp_ranks))]
+        else:
+            x_list = None
+            labels_list = None
+
+        torch.distributed.gather(x, gather_list=x_list, dst=driver_rank, group=self.pp_group)
```
aazzolini, line 98: nice! question: does this support un-even splits?
In general it would be nice to see if we could adopt something like DTensor for this use case later on . c.c. @wanchaol

examples/ddp2pipe/ddp2pipe.py
aazzolini: Assuming the TODO here would be to create a torch.function as discussed?

aazzolini (approved):
This looks great! I just have a few questions which are mostly for my ramping up in Pippy in general.

examples/ddp2pipe/ddp2pipe.py
wanchaol: Yeah, agreed. looks like this is user facing API, it would be nice to merge these two concepts so that they have same API to the user https://github.com/pytorch/tau/blob/main/spmd/tensor/placement_types.py#L29

examples/ddp2pipe/.gitignore
pbelevich: generated by example for MNIST dataset

PR #435, opened 2022-09-06
[spmd] add more reshard suggestions to einop_rule

wanchaol:
This PR add reshard suggestions to einop rule, namely if we found
that the same tensor dimension is both replicated and sharded across
different inputs, we return the reshard suggestions to reshard the
replicated input to sharded. Taking an example:

If we call `bmm(replicate, shard(0))`, we should make the first one
be also shard on the batch dimension by performing a zero cost reshard

This rule is general to all einop related rule, as long as the named
dimension of inputs are the same, added tests to test bmm and baddbmm
works

TODO: merge the reshard suggestions on this with the partial linear reshard suggestions.

### DIFF
--- a/spmd/tensor/ops/matrix_ops.py
+++ b/spmd/tensor/ops/matrix_ops.py
@@ -58,16 +58,28 @@ def bmm_rules(op_schema: OpSchema) -> OutputSharding:
 @register_prop_rule("aten.baddbmm.default")
 def baddbmm_rules(op_schema: OpSchema) -> OutputSharding:
     input_spec, mat1_spec, mat2_spec = op_schema.args_spec
-    bmm_output_spec = bmm_rules(
-        OpSchema((mat1_spec, mat2_spec), {})
-    ).output_spec
-    if bmm_output_spec is None:
-        # TODO: add suggestion
-        return OutputSharding(None)
+    bmm_output_sharding = bmm_rules(OpSchema((mat1_spec, mat2_spec), {}))
+    if bmm_output_sharding.output_spec is None:
+        # TODO: add more suggestions
+        if bmm_output_sharding.schema_suggestions is not None:
+            input_suggestion = bmm_output_sharding.schema_suggestions[0]
+            # reconstruct baddbmm output sharding
+            bmm_output_sharding.schema_suggestions[0] = OpSchema(
+                args_schema=(
+                    input_spec,
+                    input_suggestion.args_schema[0],
+                    input_suggestion.args_schema[1],
+                ),
+                kwargs_schema=op_schema.kwargs_schema,
+            )
+            return bmm_output_sharding
+        else:
+            return OutputSharding(None)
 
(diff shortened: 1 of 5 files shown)


test/spmd/tensor/test_matrix_ops.py
```diff
@@ -223,19 +233,28 @@ def test_placement_comb(
...
         passlist = [
             (shard0_spec, shard0_spec),
+            (shard0_spec, replica_spec),
+            (shard1_spec, replica_spec),
+            (shard2_spec, replica_spec),
+            (replica_spec, shard0_spec),
+            (replica_spec, shard1_spec),
+            (replica_spec, shard2_spec),
             (shard2_spec, shard1_spec),
+            (replica_spec, replica_spec),
```
fduwjj, line 250: Can we also add backward in the test like what we have in https://github.com/pytorch/tau/pull/382/files#diff-156f4bb8984b2bbd435302d728ba6a9607d2a197873c03dd0c8a73dff5e8a06aR94?

spmd/tensor/ops/matrix_ops.py
```diff
@@ -58,16 +58,28 @@ def bmm_rules(op_schema: OpSchema) -> OutputSharding:
...
+            # reconstruct baddbmm output sharding
+            bmm_output_sharding.schema_suggestions[0] = OpSchema(
+                args_schema=(
+                    input_spec,
+                    input_suggestion.args_schema[0],
+                    input_suggestion.args_schema[1],
+                ),
+                kwargs_schema=op_schema.kwargs_schema,
+            )
+            return bmm_output_sharding
```
fduwjj, line 75: N00b question, looks like we are doing sharding propagation in https://github.com/pytorch/tau/blob/main/spmd/tensor/dispatch.py#L155. And this "output_sharding" is the input for the sharding propagation. Then why do we name it with "output"?

fduwjj (commented):
LGTM, thanks for making this happening, I just have some n00b questions.

aazzolini (approved):
Thanks @wanchaol ! At this point, are we ready to generalize to other linear operations as well?

test/spmd/tensor/test_matrix_ops.py
wanchaol: Sure, will add backward tests in a follow up PR

spmd/tensor/ops/matrix_ops.py
wanchaol: Output_Sharding refers to the output sharding after a op propagation. in dispatch it's also the output of a shard_prop_func? It's just that in case output sharding is not successful we will use the info to reshard inputs

wanchaol:
> Thanks @wanchaol ! At this point, are we ready to generalize to other linear operations as well?

Yep this should be generalized in addmm and mm natively, it's just that we need to add more tests to test this.

Issue #433, opened 2022-09-05
#372 broke pippy_mnist.py


pbelevich:
https://github.com/pytorch/tau/pull/372 broke pippy_mnist.py:

```
Traceback (most recent call last):
  File "/Users/pbelevich/miniconda3/envs/PiPPy/lib/python3.9/site-packages/torch/multiprocessing/spawn.py", line 69, in _wrap
    fn(i, *args)
  File "/Users/pbelevich/PycharmProjects/PiPPy/pippy/utils.py", line 103, in run_worker
    run_master(pp_ranks_per_dp_group[rank], args, *extra_args)
  File "/Users/pbelevich/PycharmProjects/PiPPy/examples/mnist/pippy_mnist.py", line 143, in run_master
    outp, _ = pipe_driver(x_batch, y_batch)
  File "/Users/pbelevich/miniconda3/envs/PiPPy/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/Users/pbelevich/PycharmProjects/PiPPy/pippy/PipelineDriver.py", line 1412, in forward
    return merge_chunks(local_results, self.output_chunk_spec, self._debug_mask_minibatches)
  File "/Users/pbelevich/PycharmProjects/PiPPy/pippy/microbatch.py", line 229, in merge_chunks
    args_flattened.append(torch.cat(values_to_cat, dim=arg.split_dim))
TypeError: expected Tensor as element 0 in argument 0, but got ValueReference
```

PR #438, opened 2022-09-06
Fix getitem's noop condition

kwen2501:
Was messed up in #372

Fixes issue #433 : #372 broke pippy_mnist.py (in the validation phase)


pbelevich (approved):
LGTM!

PR #439, opened 2022-09-06
[DistributedTensor] Add support for tanh

anj-s:
Removed xfail and added tanh to the list of pointwise ops


wanchaol (approved):
it looks like this also enables `nn.functional.tanhshrink`, could we remove that xfail as well?

PR #441, opened 2022-09-07
Add debugging msgs for figuring out sharding propragation issues.

aazzolini:


spmd/tensor/dispatch.py
```diff
@@ -26,6 +26,12 @@
 _ENABLE_FALLBACK = False
 
 
+"""
+Print information ops input and output sharding for debugging purposes.
```
wanchaol, line 30: it looks like we only print out the input sharding infos?

wanchaol (approved):
lgtm!

PR #443, opened 2022-09-07
[dtensor] bulk adding most pointwise ops

wanchaol:
This PR adding pointwise ops in bulk, since we have the dtensor op db testing in, we could directly enable all the pointwise ops easily. Note that it currently only testing forward, we need to enable the backward tests to test the backward and add more backward related ops.


aazzolini (approved):
This is great for improving coverage!

PR #444, opened 2022-09-07
[DTensor] Add support for BW for all ops

anj-s:
Add support for testing .backward() for all dtensor ops.
- Sets the requires_grad flag for the input tensor. 
- Catching all exceptions so that I can track all ops that need support. 

To follow up:
1. Failing transpose ops (_H, _T) and torch.isfinite
2. Failing test/spmd/tensor/test_dtensor_ops.py::TestDTensorOpsCPU::test_dtensor_op_db_nn_functional_tanhshrink_cpu_float32
and test/spmd/tensor/test_dtensor_ops.py::TestDTensorOpsCPU::test_dtensor_op_db_sgn_cpu_float32
3. Types of inputs such as list of tensors


test/spmd/tensor/test_dtensor_ops.py
```diff
@@ -669,6 +670,11 @@ def wrapped(fn):
 
 
 def run_dtensor_crossref(test_case, func, args, kwargs):
+    run_bw = False
+    if hasattr(args[0], "requires_grad"):
```
wanchaol, line 674: I think every tensor has `requires_grad` attribute? it's just that some is false and some is true. we should probably do it like this:

`if isinstance(args[0], torch.Tensor) and args[0].requires_grad`

anj-s: By default the value is False and instead of checking for if arg[0] is a Tensor, I check the presence of the attribute. I think we always want to run BW by default.

wanchaol: got it, yeah make sense.

test/spmd/tensor/test_dtensor_ops.py
```diff
@@ -668,7 +669,15 @@ def wrapped(fn):
 }
 
 
+skip_bw = [None, "torch.isfinite"]
```
wanchaol, line 672: what's None mean here?

test/spmd/tensor/test_dtensor_ops.py
```diff
@@ -668,7 +669,15 @@ def wrapped(fn):
...
+
+
 def run_dtensor_crossref(test_case, func, args, kwargs):
+    run_bw = False
+    if hasattr(args[0], "requires_grad") and resolve_name(func) not in skip_bw:
```
wanchaol, line 677: if we always run backward, why we need the `run_bw` flag? or we want to skip some op where it first input is not a tensor?

test/spmd/tensor/test_dtensor_ops.py
wanchaol, line 672: could we add `tanhshrink`, `H`, `T`, and `sgn` to this list if that's the only ones that are failing?

wanchaol (commented):
looking good, have a few comments.

test/spmd/tensor/test_dtensor_ops.py
anj-s: sorry, I should have added a comment. None is what func translates to for the transpose operations. There isn't a clean op name like torch.infinite.

test/spmd/tensor/test_dtensor_ops.py
anj-s: Thats right. I am following up on some of those cases as well to see if we need to support a list of input tensors from the ops test framework.

test/spmd/tensor/test_dtensor_ops.py
anj-s: It was only for CPU ops so I was still trying to figure it out I guess :)
Added it to the skip_bw list for now to unblock the PR.

anj-s: The issue was that we had to remove xfail for the tanhshrink ops which got added back due to a bad merge. Removing the xfail for the tanhshrink and sgn ops means the test passes locally. Waiting for the green signal on circle CI.

spmd/tensor/ops/pointwise_ops.py
```diff
@@ -142,6 +142,10 @@
     "aten.mul_.Tensor",
     "aten.ne.Scalar",
     "aten.tanh.default",
+    "aten.tanh_backward.default",
+    "aten.sgn.default",
+    "aten.neg.default",
+    "aten._softmax_backward_data.default",
```
wanchaol, line 148: I guess we can resolve this later, but is `_softmax_backward_data` a point-wise op? My impression is that it's the same type as `softmax`.

wanchaol (approved):

spmd/tensor/ops/pointwise_ops.py
XilunWu: same question. Tried to apply the same rule as `softmax` to `softmax_backward` but failed on test.

XilunWu: Regarding `_softmax_backward_data`, won't directly applying `pointwise_rule` miss the checking on shard dim vs. softmax dim?

PR #446, opened 2022-09-08
Test tracing DeviceMesh

mrshenli:
Add test cases that use CommTensor to trace DeviceMesh comms.

Land after https://github.com/pytorch/pytorch/pull/84655, which introduces `torch.distributed._spmd.CommTensor` and it's used by this PR. Tested locally, CommTensor can trace DeviceMesh comms.

### DIFF
--- a/test/spmd/test_tracing.py
+++ b/test/spmd/test_tracing.py
@@ -1,0 +1,146 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates
+import torch
+
+from torch.distributed._spmd.comm_tensor import CommTensor
+from torch.distributed.distributed_c10d import (
+    get_global_rank,
+    get_world_size,
+)
+from torch.fx.experimental.proxy_tensor import make_fx
+from torch.testing._internal.common_utils import run_tests
+from spmd.testing.common_utils import (  # type: ignore
+    DistTensorTestBase,
+    with_comms,
+)
+from spmd.tensor import DeviceMesh
+
+from typing import List
+
+
+class TraceDeviceMeshTest(DistTensorTestBase):
+    @property
+    def world_size(self) -> int:
+        return 4
+
+    @with_comms
+    def test_tracing_all_reduce_nd(self):
+        mesh_tensor = torch.arange(4).reshape(2, 2)
+        mesh = DeviceMesh(self.device_type, mesh_tensor)
+        local_tensor = torch.ones(3, 3, device=self.device_type) * self.rank
+
+        # check all dim groups
+        dim_to_subgroups = mesh.get_dim_groups()
+        for dim, dim_group in enumerate(dim_to_subgroups):
+            dim_group_size = get_world_size(dim_group)
+            global_ranks = [
+                get_global_rank(dim_group, i) for i in range(dim_group_size)
+            ]
+
+            def fn(tensor: torch.Tensor):
+                comm_tensor = CommTensor(tensor)
+                reduced_tensor = mesh.all_reduce(comm_tensor, mesh_dim=dim)
+                # multiply with 1 to trigger wait on read during tracing.
+                return reduced_tensor * 1
+
+            # use a local_tensor + 1 for tracing to make sure that we are not
+            # simply replaying recorded tensor value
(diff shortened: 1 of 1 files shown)


test/spmd/test_tracing.py
```diff
@@ -0,0 +1,142 @@
...
+        dim_to_subgroups = mesh.get_dim_groups()
+        for dim, dim_group in enumerate(dim_to_subgroups):
+            dim_group_size = get_world_size(dim_group)
+            global_ranks = [
+                get_global_rank(dim_group, i) for i in range(dim_group_size)
+            ]
+
+            def fn(tensor: torch.Tensor):
+                comm_tensor = CommTensor(tensor)
+                reduced_tensor = mesh.all_reduce(comm_tensor, mesh_dim=dim)
```
wanchaol, line 41: is there a reason we are calling the collective under the `for` loop over dim_groups?

test/spmd/test_tracing.py
```diff
@@ -0,0 +1,142 @@
...
+        # check all dim groups
+        dim_to_subgroups = mesh.get_dim_groups()
+        for dim, dim_group in enumerate(dim_to_subgroups):
+            dim_group_size = get_world_size(dim_group)
+            global_ranks = [
+                get_global_rank(dim_group, i) for i in range(dim_group_size)
+            ]
+
+            def fn(tensor: torch.Tensor):
+                comm_tensor = CommTensor(tensor)
```
wanchaol, line 40: does this imply that we need to add CommTensor to device mesh collectives in the future PRs?

test/spmd/test_tracing.py
```diff
@@ -0,0 +1,142 @@
...
+    with_comms,
+)
+from spmd.tensor import DeviceMesh
+
+from typing import List
+
+
+class TraceDeviceMeshTest(DistTensorTestBase):
+    @property
+    def world_size(self) -> int:
```
wanchaol, line 22: I think our CI could only get 4 GPUs (for budget reason), so world_size=8 might not work well with the GPU CI, it could only test CPU CI

wanchaol (approved):
Looks great! my understanding is this need to be landed after the corresponding pytorch PR merged and nightly build finishes, so that the tau CI could pick up the changes?

test/spmd/test_tracing.py
mrshenli: Yep, will try that in a followup PR

mrshenli (commented):
land after #454

test/spmd/test_tracing.py
mrshenli: I am reducing this PR to use world size 4. Will add another test for cpu-only 8 process coverage.

fduwjj: If 8 GPU is needed, we can have another conversation with Dev Infra team to ask for it. @mrshenli do we really want machines with GPUs more than 4? We can add a tag to enable tests in CI for scenarios when we need CI for cases when GPU>4. I remember PyTorch CI does have something similar.

mrshenli: IIUC, this is not urgent at this point, but will help us to improve test coverage. Do we know how long it might take to get a 8GPU machine? If it is prompt, we can wait till we have real 3D device mesh use cases? cc @wanchaol @aazzolini

fduwjj: As long as we can justify and get the budget. It should be very quick. Previously. Ke added new machines in https://github.com/pytorch/test-infra/pull/602. And to change our CI script(within Tau repo) should also straightforward. Adding to PyTorch CI can be another topics which I believe we already have a thread for that.

XilunWu: Just a note on the current divergence of GPU number used in tests:
mesh tests now are using 8 GPUs for testing and the others are still using `self.world_size` which is set to `TEST_GPU_NUM=4`. We need to unify this number later.

Code pointer:
https://github.com/pytorch/tau/blob/main/test/spmd/tensor/test_device_mesh.py#L20
https://github.com/pytorch/tau/blob/main/spmd/testing/common_utils.py#L31

PR #447, opened 2022-09-08
[dtensor] support implicit broadcasting for pointwise ops

wanchaol:
This PR support implicit broadcasting for pointwise ops, where it is able to propagate the sharding of "broadcasting to a common shape" case.

When the dimension of a input is 1, we mark it explictly during propagation, so that the sharding can be propagated without trying to reshard i.e.
input0 (8, 4, 8), input 1(8, 1, 8), if we shard on input0 dim 1, it could directly propagate the sharding without leeting input 1 be also shard on dim1

NOTE that since the op rules might be used by some chained suggestions (i.e. addmm or baddbmm chained suggestion), one spec might be the output of a sharding propagation, so it must have shape there to validate if there's singleton dimension. As a result, this PR forces sharding propagation rules to generate correct output shape

### DIFF
--- a/spmd/tensor/ops/pointwise_ops.py
+++ b/spmd/tensor/ops/pointwise_ops.py
@@ -167,13 +167,15 @@ def pointwise_rule(
         # we mark the dim char as a special "1" to distinguish with
         # the non-singleton dimension, so that sharding propagation
         # should just ignore the singleton dimension.
-        for i, dim_length in enumerate(input.shape):
-            if dim_length == 1:
-                # mark singleton dim char as a special "1" in einop rule
-                p = p[:i] + "1" + p[i + 1 :]
+        if len(input_specs) > 1:
+            for i, dim_length in enumerate(input.shape):
+                if dim_length == 1:
+                    # mark singleton dim char as a special "1" in einop rule
+                    p = p[:i] + "1" + p[i + 1 :]
         dimchars.append(p)
     out_dimchars = alphabet[:max_dim]
     fmt = f"{','.join(p for p in dimchars)}->{out_dimchars}"
+    print(f">>>> fmt: {fmt}")
     return einop_rule(fmt, op_schema, linearity=linearity)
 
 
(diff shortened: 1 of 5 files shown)


spmd/tensor/ops/pointwise_ops.py
```diff
@@ -154,18 +154,28 @@ def pointwise_rule(
...
+        # see https://pytorch.org/docs/stable/notes/broadcasting.html
+        # If any of the dimensions is singleton dimension (i.e. 1).
+        # we mark the dim char as a special "1" to distinguish with
+        # the non-singleton dimension, so that sharding propagation
+        # should just ignore the singleton dimension.
+        if len(input_specs) > 1:
+            for i, dim_length in enumerate(input.shape):
+                if dim_length == 1:
+                    # mark singleton dim char as a special "1" in einop rule
+                    p = p[:i] + "1" + p[i + 1 :]
```
aazzolini, line 174: Nice :)

spmd/tensor/ops/pointwise_ops.py
```diff
@@ -154,18 +154,28 @@ def pointwise_rule(
...
         dimchars.append(p)
     out_dimchars = alphabet[:max_dim]
     fmt = f"{','.join(p for p in dimchars)}->{out_dimchars}"
+    print(f">>>> fmt: {fmt}")
```
aazzolini, line 178: Remove?

aazzolini (approved):

spmd/tensor/placement_types.py
```diff
@@ -48,24 +48,19 @@ class _Partial(Placement):
...
     mesh: DeviceMesh
     placements: Sequence[Placement]
-    # ndim of the current dist tensor, if only pass in shape, this
-    # field will be updated to len(shape), if not pass in shape, must
-    # pass in ndim
-    ndim: int = -1
     # shape of the current dist tensor, this will be set upon
     # construction of the DTensor, prop rule could read it, and
-    # don't need to set in output spec when calculate the output
+    # would need to set in output spec when calculate the output
```
aazzolini, line 53: oh is this because we're chaining propagation rules?

wanchaol: Yep because we chain the propagation rule, so mm out need to have shape on its DTensorSpec so that we could detect singleton on its generated dtensor spec

PR #449, opened 2022-09-08
Add empty_like and such

aazzolini:


wanchaol (approved):
nice!

fduwjj (approved):

test/spmd/tensor/test_tensor_ops.py
```diff
@@ -84,6 +84,40 @@ def test_op_out_variant(self):
...
 
+    @with_comms
+    def test_empty_like(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        shard_spec = [Shard(0)]
+
+        input_tensor = torch.randn(4, 8, requires_grad=True)
+        dist_tensor = DTensor.from_local(input_tensor, device_mesh, shard_spec)
+        empty_like_dt = torch.empty_like(dist_tensor)
+        # empty is not deterministic, so we only check that the shard propagation worked
```
anj-s, line 95: Is empty not deterministic even if we set the seed on each rank to be the same?

test/spmd/tensor/test_tensor_ops.py
```diff
@@ -84,6 +84,40 @@ def test_op_out_variant(self):
...
+        self.assertEqual((4, 8), empty_like_dt.to_local().shape)
+
+    @with_comms
+    def test_fill_inplace(self):
```
anj-s, line 99: wondering if we can use test_dtensor_ops for the below tests?

test/spmd/tensor/test_tensor_ops.py
aazzolini: I believe it simply doesn't initialize the memory so it's probably going to be just trash.

test/spmd/tensor/test_tensor_ops.py
aazzolini: i tried remove xfail and it didn't work. i'll defer to @wanchaol

PR #452, opened 2022-09-10
Enable HF examples tests

pbelevich:


kwen2501 (approved):
Thanks for enabling these tests! (I was meant to test them one by one but this way is obviously faster.)
It looks like all tests have passed, maybe it is good to land?

PR #454, opened 2022-09-13
Remove deprecated eig

mrshenli:
`eig` is removed from PyTorch by
https://github.com/pytorch/pytorch/pull/70982. Update spmd accordingly.


wanchaol (approved):
Thanks!

PR #440, opened 2022-09-07
[spmd] complete softmax and _softmax_backward_data to support aggregate on sharding dim

XilunWu:
1. Adapt softmax and _softmax_backward_data (added in #382) ops to shard propagation rule and move them to tensor_ops.py
2. Move relevant tests to test_tensor_ops.py
3. Extend test coverage on (batch_dim, softmax_dim) combination except for case batch_dim == softmax_dim.

PR #455, opened 2022-09-13
[spmd] adapt to softmax ops to sharding prop rules

XilunWu:
This PR aims to enable softmax ops using sharding propagation rules and is a part of #440. Softmax was prototyped in #382 to enable ViT model and now it's time to productionize it with sharding propagation rules:

_softmax: a default propagation rule with a mandatory dimension checking `sharding dim != softmax dim`
_softmax_backward_data: it is fine to treat it as a pointwise rule when `sharding dim != softmax dim`. For the rest, simply applying pointwise rule produces correct result but this needs to be confirmed.

In this PR, softmax ops will raise exception when `sharding dim == softmax dim`. We plan to complete the ops by generating a schema suggestion to replicate the tensor instead in future.


test/spmd/tensor/test_tensor_ops.py
```diff
@@ -152,6 +154,82 @@ def test_zeros_like(self):
...
+                self.assertEqual(dist_y.to_local(), local_y)
+
+    @with_comms
+    def test_softmax_with_bwd(self):
+        device_mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
+        dims = range(3)  # used to convert -1 to the actual dim
+        softmax_dims = [-1, 0, 1, 2]
+        shard_dims = [-1, 0, 1, 2]
+        test_list = list(itertools.product(softmax_dims, shard_dims))
+        # DTensor's _softmax_backward_data produces wrong result on CPU on certain dimension.
```
wanchaol, line 191: can you add a TODO here to remove this inconsistency later?

test/spmd/tensor/test_tensor_ops.py
```diff
@@ -152,6 +154,82 @@ def test_zeros_like(self):
...
+        shard_dims = [-1, 0, 1, 2]
+        test_list = list(itertools.product(softmax_dims, shard_dims))
+
+        for softmax_dim, shard_dim in test_list:
+            local_y = torch.nn.functional.softmax(
+                x, dim=softmax_dim, dtype=torch.float32
+            )
+            dist_x = distribute_tensor(x, device_mesh, [Shard(shard_dim)])
+            if dims[shard_dim] == dims[softmax_dim]:
+                with self.assertRaises(Exception):
```
wanchaol, line 172: can you use assertRaiseRegex and assert the exception str to be the one you throw?

test/spmd/tensor/test_tensor_ops.py
```diff
@@ -152,6 +154,82 @@ def test_zeros_like(self):
...
+            else:
+                dist_y = torch.nn.functional.softmax(
+                    dist_x, dim=softmax_dim, dtype=torch.float32
+                )
+                self.assertTrue(dist_y.placements[0].is_shard(dim=shard_dim))
+                dist_y = dist_y.redistribute(device_mesh, [Replicate()])
+                self.assertEqual(dist_y.to_local(), local_y)
+
+    @with_comms
+    def test_softmax_with_bwd(self):
```
wanchaol, line 185: can we run this on GPU only for now with @skip_if_lt_x_gpu?

wanchaol (approved):
Looks good, have some suggestions inlined, and please fix the lint before landing.

test/spmd/tensor/test_tensor_ops.py
```diff
@@ -152,6 +154,82 @@ def test_zeros_like(self):
         zeros_expected = torch.zeros(4, 8)
         self.assertEqual(zeros_expected, zeros_like_dt.to_local())
 
+    @with_comms
+    def test_softmax_fwd(self):
```
fduwjj, line 158: Ideally, we can merge fwd and bwd into one test right?

fduwjj (commented):
Thanks!

spmd/tensor/ops/math_ops.py
```diff
@@ -211,3 +214,30 @@ def reduction_rule(op_schema: OpSchema) -> OutputSharding:
...
+    from spmd.tensor.ops.pointwise_ops import pointwise_rule
+
+    grad_out_spec, out_spec, softmax_dim, _ = op_schema.args_schema
+    input_specs = [
+        cast(DTensorSpec, spec) for spec in [grad_out_spec, out_spec]
+    ]
+    softmax_dim = cast(int, softmax_dim)
+    ops_dim_map = [spec.dim_map for spec in input_specs]
+    ops_dim_map = list(zip(*ops_dim_map))
+    if softmax_dim < len(ops_dim_map) and 0 in ops_dim_map[softmax_dim]:
```
wanchaol, line 241: this should be >0, 0 is only mean sharding on the 0 mesh dim

test/spmd/tensor/test_tensor_ops.py
XilunWu: Yes. We decide to temporarily keep fwd test because the bwd test fails on CPU on some cases and we will disable it on CPU with @skip_if_lt_x_gpu. Once the test passes on CPU, I'll remove fwd test.

spmd/tensor/ops/math_ops.py
```diff
@@ -211,3 +214,30 @@ def reduction_rule(op_schema: OpSchema) -> OutputSharding:
...
     DTensor._op_to_rules[reduction_op] = reduction_rule
+
+
+@register_prop_rule("aten._softmax.default")
+def softmax_rule(op_schema: OpSchema) -> OutputSharding:
+    input_spec, softmax_dim, _ = op_schema.args_schema
+    input_spec = cast(DTensorSpec, input_spec)
+    softmax_dim = cast(int, softmax_dim)
+    dim_map = input_spec.dim_map
+    if softmax_dim < len(dim_map) and dim_map[softmax_dim] >= 0:
```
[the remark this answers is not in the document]
XilunWu: len(dim_map) will always equal to tensor's `ndim`. In this way, calling softmax(dim=5) on a 3D tensor should give `IndexError: dimension out of range` (as PyTorch does)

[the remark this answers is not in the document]
fduwjj: Oh, interesting, I thought PyTorch might have more customized error msg.. If the error msg is the same, then we are good.

fduwjj (approved):
lgtm.

fduwjj:
You can use Squash and edit the commit msg when you merge. So that all commit info can be squashed. Thanks again for working on this.

PR #457, opened 2022-09-13
Infer Future.device from buffer.device

kwen2501:
Previously we infer the `device` for Future from the device of parameters. 
Try using the device of buffer when no parameter found.

### DIFF
--- a/pippy/PipelineDriver.py
+++ b/pippy/PipelineDriver.py
@@ -551,16 +551,21 @@ class PipeStageExecutor(EventRecorder):
         # We assume the output value is on the same device as the stage's
         # module, and that all parameters in the module are on the same device
         # HACK: we assume the module has at least one parameter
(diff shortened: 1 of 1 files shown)


pippy/PipelineDriver.py
```diff
@@ -551,16 +551,21 @@ def extract_value_ref_args(arg):
...
         param = next(self.mod.parameters(), None)
-        if param is None:
-            logging.warning(f"Module of stage {self.stage_id} has 0 parameters, "
-                            f"cannot figure out device. Setting it to cpu")
-        else:
+        buffer = next(self.mod.buffers(), None)
+        if param is not None:
             device = param.device
+        elif buffer is not None:
+            device = buffer.device
```
anj-s, line 560: what is an example when we don't find any parameters but only buffers?

PR #460, opened 2022-09-15
Trace local graph and then replace nodes with DT's dispatch graph

mrshenli:
* __->__ #460
* #476
* #521


mrshenli:
land after https://github.com/pytorch/pytorch/pull/84980 shows up in PT nightly

test/spmd/test_tracing.py
```diff
@@ -172,5 +178,134 @@ def test_all_gather_nd(self):
...
+        # walk through the traced local graph and expand node with DTensor's
+        # dispatch implementation
+        for node in traced_f.graph.nodes:
+            if node.op == "placeholder":
+                node_to_obj[node] = name_to_input(node.name)
+            elif isinstance(node.target, torch._ops.OpOverload):
+                args = tree_map(lambda n: node_to_obj[n], node.args)
+                kwargs = tree_map(lambda n: node_to_obj[n], node.kwargs)
+
+                out = operator_dispatch(
```
aazzolini, line 221: So this is doing "concrete" tracing, where we're tracing based on a concrete DTensor. Is this just for testing purposes or do we want to actually rely on concrete tracing for the final solution?

mrshenli: This is not yet tracing. It is getting the intermediate DTensor outputs, so that we know what DTensor specs to pass to the next operator. Tracing is on line 318, which only applies to local tensors.

aazzolini: Got it, yeah what I mean is that this corresponds to doing concrete tracing -- where you need the concrete tensors in order to propagate sharding.

aazzolini (approved):
Looks good, thanks!

PR #461, opened 2022-09-15
[dtensor] refactor mesh collectives and use tensor_split

wanchaol:
* #492
* #491
* #478
* __->__ #461

This PR refactors mesh collective to be more robust to use, and for sharding,
we drop torch.chunk and use torch.tensor_split instead. Specifically this PR:
1. adjust all collectives to use c10d collective directly instead of nn.functional
   collective.
2. make sure we handle all the output/input preparing logic inside the collective
   and make it more easy to use
3. switch sharding to use tensor_split instead of chunk, because chunk cann't handle
   things like chunking size 5 to 4 devices evenly (it have empty shard on some device).
4. delete all "base" related collectives (all_gather_base, reduce_scatter_base)
   as they will be deprecated in the future, and not really supporting
   all our use cases (the base collective only cat on prim dimension)


spmd/tensor/__init__.py
```diff
@@ -52,17 +52,15 @@ def distribute_tensor(
...
                 f"Only support chunk sharding evenly now, but tensor got "
                 f"dimension {shard_dim} of size {tensor.size(shard_dim)}, "
                 f"which does not divide number of shards {num_chunks}."
             )
-            chunk_size = tensor.size(shard_dim) // num_chunks
-            tensor_list = list(tensor.chunk(num_chunks, dim=shard_dim))
-            scatter_shape = list(tensor.size())
-            scatter_shape[shard_dim] = chunk_size
+
+            tensor_list = list(tensor.tensor_split(num_chunks, dim=shard_dim))
```
aazzolini, line 63: Nice, tensor_split this seems to be the semantic we're looking for.

spmd/tensor/device_mesh.py
```diff
@@ -261,18 +256,23 @@ def scatter(
...
             mesh_dim (int, optional): indicate which mesh dimension we want
                 to scatter on, we by default choose the first rank on the
-                mesh dimension.
+                mesh dimension as source of truth.
 
         Returns:
             A :class:`torch.Tensor` object
         """
+        my_coordinate = self.get_coordinate_on_dim(mesh_dim)
+        # TODO: what should happen if rank is not in the mesh?
```
aazzolini, line 265: Should we just silently return? I think we had discussed this before. In general we think that a good principle is that no-op for non participating members is a good behavior.

spmd/tensor/device_mesh.py
```diff
@@ -261,18 +256,23 @@ def scatter(
...
+        assert my_coordinate is not None
```
aazzolini, line 266: If we go with the assert, could you add a descriptive error message?

spmd/tensor/device_mesh.py
```diff
@@ -284,47 +284,72 @@ def scatter(
...
         if dim_group is not GroupMember.WORLD:
             src_for_dim = get_global_rank(dim_group, 0)
 
-        return broadcast(tensor.contiguous(), src=src_for_dim, group=dim_group)
-
-    # pyre-fixme[3]: Return type must be annotated.
-    def all_gather(self, tensor: torch.Tensor, mesh_dim: int = 0):
-        dim_group = self._dim_groups[mesh_dim]
-        return all_gather(tensor, group=dim_group)
+        tensor = tensor.contiguous().clone()
```
aazzolini, line 310: does clone trigger an additional copy, or is this something that happened anyways within the functional API beforehand?  (Also, if contigous actually triggered a copy already, then we don't need an additional copy?)

spmd/tensor/device_mesh.py
```diff
@@ -284,47 +284,72 @@ def scatter(
...
-            return output
-        else:
-            # if not nccl, fallback to use all_gather
-            # and reform the output tensor by concat
-            gathered_chunks = self.all_gather(tensor, mesh_dim=mesh_dim)
-            # TODO: find more performant way
-            output_tensor.copy_(torch.cat(gathered_chunks, dim=tensor_dim))
-            return output_tensor
+    ) -> torch.Tensor:
+        """
```
aazzolini, line 321: Thanks for adding comments!

spmd/tensor/redistribute.py
```diff
@@ -100,53 +100,57 @@ def redistribute_spmd_tensor(
...
+            elif current.is_replicate():
                 # slice/narrow the tensor to corresponding local shard then return shard tensor
-                # TODO: support multi dim redistribute require to not use get_rank, as it would
-                # get global_rank and in the case of multidim mesh it's not correct!
                 new_local_tensor = local_tensor.narrow(
-                    shard_dim, my_rank * chunk_size, chunk_size
+                    target_dim, my_rank * chunk_size, chunk_size
                 )
+            else:
+                assert current.is_shard()
```
aazzolini, line 146: Are we removing support for this? This will probably make view_ops tests fail. Do we need to remove it?

aazzolini: (As discussed offline , this case shouldn't hit since _decompose_sharding above should decompose Shard(0) -> Shard(1) into Shard(0) -> Replicate -> Shard(1). We can make this a logic error, and add a comment that it shouildn't happen.  The TODO to make it all_to_all still applies.

spmd/tensor/device_mesh.py
```diff
@@ -249,40 +244,61 @@ def get_coordinate_on_dim(self, dim: int) -> Optional[int]:
...
+        num_chunks = self.size(mesh_dim)
+        # TODO: handle uneven shard sizes
+        assert tensor_to_scatter.size(tensor_dim) % num_chunks == 0, (
+            f"Only support chunk sharding evenly now, but tensor got "
+            f"dimension {tensor_dim} of size {tensor_to_scatter.size(tensor_dim)}, "
+            f"which does not divide number of shards {num_chunks}."
+        )
+
+        scatter_list = list(
+            tensor_to_scatter.tensor_split(num_chunks, dim=tensor_dim)
```
fduwjj, line 285: Interesting, I didn't know that PyTorch also support this. We might want to document this somewhere. (I mean this is different from the default behavior of Chunk. What if users want the same behavior of Chunk? So we can set the expectation right here.)

spmd/tensor/device_mesh.py
wanchaol: Yeah, i think the tricky part is that we need to return something legit, i.e. for uneven sharding, the return tensor shape should be somewhat match, let me try remove this in a follow up PR and add some tests.

spmd/tensor/device_mesh.py
wanchaol: Yes updated, thanks for the suggestion!

spmd/tensor/device_mesh.py
wanchaol: yeah added to the doc of `distribute_tensor`

spmd/tensor/redistribute.py
wanchaol: updated with comments.

spmd/checkpoint.py
```diff
@@ -18,7 +17,7 @@
 from torch.distributed._shard.sharded_tensor.api import ShardedTensor
 
 
-def get_ranks(pg: ProcessGroup) -> List[int]:
+def get_ranks(pg: c10d.ProcessGroup) -> List[int]:
```
aazzolini, line 20: Curious on the reason for this change?

spmd/tensor/device_mesh.py
```diff
@@ -249,24 +244,46 @@ def get_coordinate_on_dim(self, dim: int) -> Optional[int]:
...
+                to split the `tensor_to_scatter` before scattering.
 
         Returns:
             A :class:`torch.Tensor` object
         """
+        my_coordinate = self.get_coordinate_on_dim(mesh_dim)
+        # TODO: what should happen if rank is not in the mesh?
+        assert (
+            my_coordinate is not None
+        ), "Rank if not part of mesh"  # TODO: figure out behavior here
```
aazzolini, line 274: Should we just ignore non-participating ranks? I think it would be a consistent behavior, but we would have to do it throughout the code. (We can do this later).

spmd/tensor/device_mesh.py
```diff
@@ -305,44 +340,60 @@ def broadcast(self, tensor: torch.Tensor, mesh_dim: int = 0):
...
+        # input tensor must be contiguous
+        tensor = CommTensor(tensor.contiguous())
+        all_gather(
+            gathered_list,
+            tensor,
+            group=dim_group,
+        )
+        return torch.cat(gathered_list, dim=tensor_dim)  # type: ignore
 
     # pyre-fixme[3]: Return type must be annotated.
```
aazzolini, line 398: is it possible to annotate this?

aazzolini (approved):
Looks great, tensor_split makes much more sense.

spmd/tensor/device_mesh.py
```diff
@@ -351,45 +402,72 @@ def all_reduce(
...
         # CommTensor does not change eager mode behavior. During tracing, it
         # makes sure communication result is properly waited before subsequent
         # read operations.
-        return all_reduce(
-            CommTensor(tensor.contiguous()), op=op, group=dim_group
-        )
+        if not tensor.is_contiguous():
+            tensor = CommTensor(tensor.contiguous())
+        else:
+            tensor = CommTensor(tensor.clone())
```
XilunWu, line 426: Curious why we need to call `.clone()` instead of `tensor = CommTensor(tensor)`?

spmd/checkpoint.py
wanchaol: oh this is sth coming from pytorch nightly changes, where type annotation seems going away for distributed.ProcessGroup, already fixed in https://github.com/pytorch/tau/pull/495, cc @kumpera

spmd/tensor/device_mesh.py
wanchaol: we actually need to make the receiving tensors be a separate receiving tensor rather than itself, otherwise collective might write into it and the result becomes wrong. Actually we can just call clone in this case. let me update

PR #466, opened 2022-09-18
Introduce a few linear pointwise ops

aazzolini:


mrshenli (approved):
LGTM!

spmd/tensor/ops/pointwise_ops.py
```diff
@@ -18,6 +18,13 @@
 #     "real",  # complex data type only
 # ]
 
+
+linear_pointwise_ops = [
+    "aten.div.Scalar",  # this op is linear on the first argument, and the second argument is scalar, so it fits as a linear op.
+    "aten.to.dtype",
```
wanchaol, line 24: why aten.to is a linear pointwise op?

PR #471, opened 2022-09-18
Implement torch.bucketize

aazzolini:


spmd/tensor/ops/tensor_ops.py
```diff
@@ -53,3 +54,35 @@ def no_shard_prop_rule(op_schema: OpSchema) -> OutputSharding:
 
 for op in no_shard_prop_ops:
     DTensor._op_to_rules[op] = no_shard_prop_rule
+
+
+@register_prop_rule("aten.bucketize.Tensor")
+def prop_bucketize(op_schema: OpSchema) -> OutputSharding:
+    """
+    Point-wise on the first input. Expect replicated for second input.
```
wanchaol, line 62: seems like we just forward the first input sharding instead of pointwise?

aazzolini: right what i mean is that since it's point-wise on the first input, we can simply forward the sharding.

aazzolini: let me improve the comment.

spmd/tensor/ops/tensor_ops.py
```diff
@@ -53,3 +54,36 @@ def no_shard_prop_rule(op_schema: OpSchema) -> OutputSharding:
...
+    Expect replicated for second input.
+    """
+    input_schema, boundaries = op_schema.args_schema
+    assert isinstance(input_schema, DTensorSpec)
+    assert isinstance(boundaries, DTensorSpec)
+
+    if all(isinstance(p, Replicate) for p in boundaries.placements):
+        return OutputSharding(output_spec=input_schema)
+    else:
+        return OutputSharding(
```
anj-s, line 72: Are we suggesting Replicate() for the second input if either of the inputs are Shard? Can you add a comment here as well?

anj-s (approved):

PR #472, opened 2022-09-18
Implement slice and slice_scatter

aazzolini:


spmd/tensor/ops/tensor_ops.py
```diff
@@ -53,3 +63,163 @@ def no_shard_prop_rule(op_schema: OpSchema) -> OutputSharding:
...
+        dim += input_spec.ndim
+    if start is None:
+        start = 0
+    if step is None:
+        step = 1
+    if end is None or end > input_spec.shape[dim]:
+        end = input_spec.shape[dim]
+    if end < 0:
+        end += input_spec.shape[dim]
+    if end > input_spec.shape[dim]:
```
anj-s, line 142: Is this already handled by line 138 or is this a catch for line 140's condition?

aazzolini: @anj-s good point, we can remove this.

spmd/tensor/ops/tensor_ops.py
```diff
@@ -111,3 +120,161 @@ def prop_bucketize(op_schema: OpSchema) -> OutputSharding:
...
+    assert isinstance(input, DTensorSpec)
+    assert isinstance(src, DTensorSpec)
+    assert isinstance(dim, int)
+
+    if dim < 0:
+        dim += input.ndim
+
+    # first, we keep the input sharding, except for the input dimension
+    # also, we cannot allow partial sum anymore.
+    input_suggestion = tuple(
```
wanchaol, line 237: can we merge this with `unshard_tensor_dim`? or is it because `aten.slice.Tensor` support forwarding partial?

wanchaol (approved):
lgtm, one comment

spmd/tensor/ops/tensor_ops.py
aazzolini: right, looking back i don't like unshard_tensor_dim since it's just a single line of code; i tought of adding allow_partial=True but then it would make it harder to read.

PR #474, opened 2022-09-18
Fix ones_like, zeros_like when input is _Partial

aazzolini:
When input is _Partial and we have operators that set the elements of the output to a given value, the output cannot be _Partial. Instead we should set it as Replicated.


anj-s (approved):

anj-s:
Thank you for the PR comments!

fduwjj (approved):
lgtm

PR #475, opened 2022-09-19
Add a bit more context on spec mismatch error.

aazzolini:
Helps debugging.

### DIFF
--- a/spmd/tensor/utils.py
+++ b/spmd/tensor/utils.py
@@ -31,12 +31,12 @@ def wrap(res: object, spec: OutputSpecType) -> object:
     if isinstance(res, torch.Tensor):
         assert spec is not None and isinstance(
             spec, DTensorSpec
-        ), "output spec does not match with output!"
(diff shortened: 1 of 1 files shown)


anj-s (approved):

spmd/tensor/utils.py
```diff
@@ -31,20 +31,20 @@ def wrap(res: object, spec: OutputSpecType) -> object:
...
+        ), f"output spec does not match with output! Expected list, got {spec}."
         return list(
             spmd_tensor.DTensor(e, s.mesh, s.placements)
             for e, s in zip(res, spec)
         )
     elif isinstance(res, tuple):
         assert spec is not None and isinstance(
             spec, tuple
-        ), "output spec does not match with output!"
+        ), f"output spec does not match with output! Expected tuple, got {spec}"
```
anj-s, line 47: nit: I would inline {spec} next to output spec and add "tuple" in brackets next to output.

PR #476, opened 2022-09-19
Refactor dispatch and redistribute to expose local tensor APIs

mrshenli:
* #460
* __->__ #476
* #521

Since `make_fx` cannot yet handle tensor subclasses correctly,
refactor dispatch code to expose APIs that takes local tensors and
then trace these APIs instead.


spmd/tensor/dispatch.py
```diff
@@ -111,20 +111,12 @@ def _reshape_alias(
 ] = {torch.ops.aten._reshape_alias.default: _reshape_alias}
 
 
-def operator_dispatch(
+def prepare_inputs(
```
aazzolini, line 114: Is the goal to use prepare_inputs directly, and bypass operator_dispatch, when using make_fx?

If this is the case, then we will probably miss some operators' implementations that directly use DTensor.

Could you provide a quick pseudo-code of the intended call sequence?

aazzolini (commented):
Could you provide a pseudo-code of the intended use of these functions exposed?

spmd/tensor/dispatch.py
mrshenli: This is how it is used in the PR on top

https://github.com/pytorch/tau/blob/a06ce4426bbddb84dc75eb9f0c10894c5c80bf41/test/spmd/test_tracing.py#L267-L301

we use `make_fx` to trace two things 1. redistributed inputs 2. local op. IIUC, we don't redistribute output at the moment in DT? If we do that in the future, we will also needs to add that to trace as well. 

Ideally, I wanted to trace `dispatch_with_local_tensors(local_args, arg_specs, output_specs)`, which will trigger redistribute on input, local comp op, redistribute on output.

spmd/tensor/redistribute.py
```diff
@@ -158,14 +157,34 @@ def redistribute_spmd_tensor(
...
 
         assert new_local_tensor is not None
         local_tensor = new_local_tensor
 
     assert new_local_tensor is not None, "redistribute failed!"
 
+    return new_local_tensor
+
+
+def redistribute_spmd_tensor(
```
wanchaol, line 171: can we merge this with the `Redistribute` autograd function?

spmd/tensor/dispatch.py
```diff
@@ -231,12 +146,107 @@ def operator_dispatch(
...
+    if str(op_call) in custom_dispatch_ops:
+        # dispatch to user defined custom distributed tensor ops
+        return custom_dispatch_ops[str(op_call)](*args, **kwargs)
+
+    target_schema, redistribute, output_sharding = prepare_inputs(
+        op_call, args, kwargs, op_to_rules
+    )
+
+    if output_sharding is None:
         # default to local tensor ops, this is wrong
```
wanchaol, line 205: this comment is invalid as of this change?

mrshenli: IIUC, the code below is still calling op on local tensors, no?

spmd/tensor/redistribute.py
mrshenli: In that case, I assume `pack_args_kwargs_with_local_tensor` will then also need to call into `Redistribute` autograd function, which occurs in the `__torch_dispatch__` function under `no_grad` mode?  

https://github.com/pytorch/tau/blob/c71e2866015ca23beee4e17c9d8dae415d5f86b4/spmd/tensor/utils.py#L57

spmd/tensor/dispatch.py
aazzolini: The issue i see is that if we do it this way we will be missing 1) the decompositions; 2) the custom op implementations (those that don't have a propagation rule, but that require direct DTensor-aware implementation). I think for some ops we will actually need those.

Would it be possible to work around make_fx limitations somehow and still be able to trace the implementation of DTensor at some level?

wanchaol, line 114: could we rename this to sth like `propagate_input_sharding`? as it also calling into sharding prop and does not only prepare the inputs

spmd/tensor/dispatch.py
```diff
@@ -111,20 +111,12 @@ def _reshape_alias(
...
-    custom_dispatch_ops: Dict[str, Callable[..., object]],
-) -> object:
-    # first we need to lift some private aten aliases to public calls
-    if op_call in _CURRENT_DECOMPOSITION_TABLE:
-        return _CURRENT_DECOMPOSITION_TABLE[op_call](*args, **kwargs)
-
-    func_schema = FunctionSchema.parse(str(op_call._schema))
-    schema_kind = func_schema.kind()
-
+) -> Tuple[OpSchema, bool, Optional[OutputSharding]]:
```
wanchaol, line 119: what is the second bool return argument mean?

spmd/tensor/dispatch.py
wanchaol: so if `output_sharding` is None, right now we fall back to executing local tensor ops? I think we disabled this by the `ENABLE_FALLBACK` flag, but now it seems we will still execute this code path?

spmd/tensor/redistribute.py
wanchaol: Oh I mean we should probably also change `pack_args_kwargs_with_local_tensor` by calling into `_redistribute_with_local_tensor` directly? so that we could safely delete this `redistribute_with_spmd_tensor` api

spmd/tensor/dispatch.py
mrshenli: Sure, let me update the name

spmd/tensor/dispatch.py
mrshenli: It should be the `redistributed` bool in the original implementation. IIUC, that dictates whether we distributed input tensors?

spmd/tensor/redistribute.py
mrshenli: sure, let me update that.

spmd/tensor/dispatch.py
mrshenli: The code snippet below is now in `prepare_inputs`. Basically line 200 will throw if `_ENABLE_FALLBACK=False`. Do you want me to move that check here?

```
        if not _ENABLE_FALLBACK:
            raise NotImplementedError(
                f"Operator {op_key} does not have a DistributedTensor rule registered."
            )
```

wanchaol: Ohhh i see, this make sense to me, we don't need to move the checks over, the current impl looks good to me.

wanchaol (approved):
lgtm!

spmd/tensor/redistribute.py
mrshenli: Hey @wanchaol , I tried update `pack_args_kwargs_with_local_tensor` and remove `redistributed_dtensor`, but code becomes a bit verbose, as `_redistribute_with_local_tensor` take more argument and requires one additional DTensor wrapping. Let me know if you prefer to get rid of `redistributed_dtensor`. I can do that in a follow up PR.

PR #477, opened 2022-09-19
Create a customized Multihead attention model for TP

fduwjj:
During TP work, we found that there are some special logic we need to enable for TP in the multi-head attention module. 

For example, sometimes, we need to implicitly change the sharding dim and DTensor view without any comms.
Also, to ensure mathematically equivalence, we also need to make the local tensor's stride same as the sharded tensor.

This is the first step for upstreaming TP code and work. We also added a unit test for multi-head attention module to compare the unsharded one and sharded one and shows that math-wise, they are equal. 

We use the same signature of torch.nn.MultiheadAttention as defined here in: https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html. Since the full logic is too much for one PR, we just focus on self-attention for now.


test/spmd/tensor/test_megatron_example.py
```diff
@@ -96,18 +115,57 @@ def aggregate_output(outputs):
...
-    def test_simple_megatron_e2e(self):
+    def test_mlp_megatron_e2e(self):
         LR = 0.5
         inp_size = [5, 10]
         torch.manual_seed(0)
         inp = torch.rand(*inp_size, device=self.device_type)
         torch.manual_seed(5)
-        model = SimpleModel(self.device_type)
+        model = MLPModule(self.device_type)
         torch.manual_seed(5)
```
anj-s, line 166: Why do we reset the manual_seed here after setting it in line 164?

test/spmd/tensor/test_megatron_example.py
```diff
@@ -96,18 +115,57 @@ def aggregate_output(outputs):
...
+        distribute_tensor(m.qkv.weight, device_mesh, col_wise_sharding)
+    )
+    m.proj.weight = torch.nn.Parameter(
+        distribute_tensor(m.proj.weight, device_mesh, row_wise_sharding)
+    )
+    if m.qkv.bias is not None:
+        m.qkv.bias = torch.nn.Parameter(
+            distribute_tensor(m.qkv.bias, device_mesh, col_wise_sharding)
+        )
+    if m.proj.bias is not None:
```
anj-s, line 139: nit: I would group projection params together for readability

test/spmd/tensor/test_megatron_example.py
```diff
@@ -96,18 +115,57 @@ def aggregate_output(outputs):
...
+        m.proj.bias = torch.nn.Parameter(
+            distribute_tensor(m.proj.bias, device_mesh, replicate)
+        )
+    m = _replicate_input_tensor(m, device_mesh, replicate)
+    m.proj = _aggregate_local_tensor(m.proj)
+    m.qkv.weight.register_hook(functools.partial(_gradient_hook, m.qkv.weight))
```
anj-s, line 145: Why do we need to do this for the TP case? I don't think I set a grad hook for any of my examples.

test/spmd/tensor/test_megatron_example.py
```diff
@@ -96,18 +115,57 @@ def aggregate_output(outputs):
...
         torch.manual_seed(0)
         inp = torch.rand(*inp_size, device=self.device_type)
         torch.manual_seed(5)
-        model = SimpleModel(self.device_type)
+        model = MLPModule(self.device_type)
         torch.manual_seed(5)
-        model_tp = SimpleModel(self.device_type)
-        model_tp = shard_module(model_tp, self.device_type)
+        model_tp = MLPModule(self.device_type)
+        shard_mlp(model_tp, self.device_type)
```
anj-s, line 168: Initialize the optimizer here before running the forward pass?

test/spmd/tensor/test_megatron_example.py
```diff
@@ -129,6 +187,44 @@ def test_simple_megatron_e2e(self):
         output_tp = model_tp(inp)
```
anj-s, line 187: I would just do a 2 step loop to check that we can do a FW + BW post optimizer.step(). That way we can check if the grads match as well as the params for both TP and non-TP

test/spmd/tensor/test_megatron_example.py
```diff
@@ -129,6 +187,44 @@ def test_simple_megatron_e2e(self):
...
         self.assertEqual(output, output_tp)
 
+    @with_comms
+    def test_self_attn_megatron_e2e(self):
+        LR = 0.05
+        inp_size = [8, 12, 16]
+        torch.manual_seed(0)
+        inp = torch.rand(*inp_size, device=self.device_type)
+        torch.manual_seed(5)
+        model = TensorParallelMultiheadAttention(
```
anj-s, line 197: Should this be initialized to the non TP MHA version?

test/spmd/tensor/test_megatron_example.py
```diff
@@ -129,6 +187,44 @@ def test_simple_megatron_e2e(self):
...
+
+        output.sum().backward()
+        output_tp.sum().backward()
+
+        optim = torch.optim.SGD(model.parameters(), lr=LR)
+        optim.step()
+        optim = torch.optim.SGD(model_tp.parameters(), lr=LR)
+        optim.step()
+
+        torch.manual_seed(3)
```
anj-s, line 222: Same comments as the previous example.

anj-s:
Thank you for the PR! nice work :) Left a few questions and suggestions.

test/spmd/tensor/test_megatron_example.py
fduwjj: Because we want to ensure models are initialized the same way.

spmd/tensor/parallel/multihead_attention_tp.py
```diff
@@ -0,0 +1,254 @@
...
+    Multi-head Attention block from Transformer models.
+    Since we need some customizations for the attention layer,
+    we are writing a customized but mathematically equivalent
+    attention module as defined in torch.nn.
+
+    Note that:
+    We now only support the case when it's self attention with
+    limited input args and we also assume that the input tensor
+    has a dimension of three. Although we do implement the logic
+    for multihead attention, it was not fully tested.
```
XilunWu, line 50: Do we need to extend this test to cover the case in future?

test/spmd/tensor/test_megatron_example.py
```diff
@@ -29,11 +20,39 @@ def forward(self, x):
...
+    module.register_forward_pre_hook(hook_func)
+    return module
+
+
 def _gradient_hook(param, grad):
     param._local_tensor.grad = grad._local_tensor
 
 
-def shard_module(m, device_type):
+def shard_mlp(m, device_type):
```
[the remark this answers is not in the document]
XilunWu: Agree with @anj-s. If `dim_groups` is absent and `mesh` is 1D, then `DeviceMesh` uses the default process group.

test/spmd/tensor/test_megatron_example.py
XilunWu: My understanding is resetting `manual_seed` ensures that `model` and `model_tp` have the same random input.

test/spmd/tensor/test_megatron_example.py
fduwjj: This is indeed needed because it's for 2D parallel. The way FSDP was designed that they will use gradients from local tensors and we need to pass gradients from DT to local tensors. cc: @wanchaol

test/spmd/tensor/test_megatron_example.py
fduwjj: Out of curiosity, is there any reason for that?

test/spmd/tensor/test_megatron_example.py
fduwjj: I think we are already doing that?

test/spmd/tensor/test_megatron_example.py
fduwjj: This is creating the input for post-optimization.

spmd/tensor/parallel/multihead_attention_tp.py
fduwjj: Yep, we will add more tests down the road as we implemented more features for Multihead Attention.

test/spmd/tensor/test_megatron_example.py
anj-s: I am suggesting the following for the sake of readability:
1. Initialize model
2. Initialize optimizer
3. for each step in the loop
-> check initial params match
-> check grads match
-> check params match post step()

Right now by running a FW post step() call, you are trying to verify the same criteria as the above FW pass.

test/spmd/tensor/test_megatron_example.py
anj-s: Yes, I am suggesting we use a training loop closer to how it is usually written.

test/spmd/tensor/test_megatron_example.py
anj-s: I only mention it as a way to improve readability. Wrote down what I meant in more detail below.

spmd/tensor/_view_n_reshard.py
```diff
@@ -0,0 +1,81 @@
...
+        ctx,
+        self: "spmd_tensor.DTensor",
+        sharding_dim,
+        shape,
+    ):
+        ctx.previous_placement = self.placements
+        ctx.previous_device_mesh = self.device_mesh
+        ctx.previous_local_shape = self.to_local().size()
+        if (
+            self.placements[0].is_shard(dim=sharding_dim)
```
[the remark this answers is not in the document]
fduwjj: QQ, do we have any use cases for multi-dim mesh?

test/spmd/tensor/test_megatron_example.py
```diff
@@ -4,23 +4,14 @@
...
-    distribute_tensor,
-    DeviceMesh,
-    DTensor,
-    Shard,
-    Replicate,
-)
-
-
-class SimpleModel(torch.nn.Module):
+from spmd.testing.common_utils import DistTensorTestBase, with_comms  # type: ignore
```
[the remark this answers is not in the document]
fduwjj: I feel this test indeed belong to tp. Maybe we can move the whole test under the new folder?

test/spmd/tensor/test_megatron_example.py
fduwjj: Because we initialize the module twice. We can definitely move the resetting seed into the module initialization.

test/spmd/tensor/test_megatron_example.py
fduwjj: I see. Got it.

spmd/tensor/_view_n_reshard.py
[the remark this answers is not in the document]
wanchaol: Yeah I think we don't in the case of 2-D Parallel if user use it that way, so maybe let's just add a check throwing if it's being passed a device mesh with more than 1d?

.github/workflows/spmd_gpu_tests.sh
```diff
@@ -18,6 +18,7 @@ pip3 --version
 # Turn off progress bar to save logs
 pip3 install --upgrade pip
 pip3 install --pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cu113
+pip3 install hypothesis
```
fduwjj, line 21: Looks like we need to install this for CI to enable test randomness.

.github/workflows/spmd_tests.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install pytest-shard numpy
+          pip install pytest-shard numpy hypothesis
```
fduwjj, line 36: We need to have this to enable randomness for CPU test.

.github/workflows/code-quality.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         pip install --pre torchvision -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
         pip install git+https://github.com/pbelevich/transformers.git@compatible_with_pt_master
         pip install "black<23" pylint==v3.0.0a5 mypy==v0.902 flake8 pyre-check==0.9.15
+    - name: Pyre check for SPMD
```
fduwjj, line 30: Move Pyre to the first place so that we can get signals faster. Mypy runs about 10 mins.

test/spmd/tensor/parallel/test_tp_examples.py
```diff
@@ -0,0 +1,376 @@
...
+                    )
+                    module.register_parameter("bias", replicated_bias)
+                    module.bias.register_hook(
+                        functools.partial(_gradient_hook, module.bias)
+                    )
+
+    def replicate_input(inputs):
+        DTensors = []
+        for tensor in inputs:
+            DTensors.append(DTensor(tensor, device_mesh, replicate))
```
wanchaol, line 165: why we can using `DTensor` constructor here instead of `from_local`? Let's not use DTensor constructor directly in non-op impls, as we probably need to make the __new__ constructor be more restrictive and only be used internally (we need to add things like shape, stride to the constructor as it will be hard to use by caller)

wanchaol (commented):
Looks good, have a few more comments

spmd/tensor/parallel/_view_with_dim_change.py
```diff
@@ -0,0 +1,101 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates
+import math
+
+import torch
+import spmd.tensor.api as spmd_tensor
```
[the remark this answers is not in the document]
fduwjj: Oh, let me correct this. I copied the code from old PR for this API.

spmd/tensor/parallel/multihead_attention_tp.py
```diff
@@ -0,0 +1,257 @@
...
+def _stride_same_as_shard(
+    tensor: torch.Tensor, tp_size: int, chunk_dim: int, cat_dim: int
+) -> torch.Tensor:
+    """
+    Adjust local tensor's stride same as the sharded situation.
+    So that view result will keeps the same.
+    """
+    if isinstance(tensor, DT):
+        return tensor
+    view_size = list(tensor.size())
```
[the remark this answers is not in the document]
fduwjj: This is not equivalent to `nn.MultiheadAttention` indeed and having this is to make non-DT mode equivalent to DT mode.

test/spmd/tensor/parallel/test_tp_examples.py
fduwjj: Sure. Let's use the public API.

spmd/tensor/parallel/multihead_attention_tp.py
```diff
@@ -0,0 +1,257 @@
...
+        value: Union[torch.Tensor, DT],
+        key_padding_mask: Optional[torch.Tensor] = None,
+        need_weights: bool = True,
+        attn_mask: Optional[torch.Tensor] = None,
+        average_attn_weights: bool = True,
+    ) -> Union[torch.Tensor, DT]:
+        b, sq, h = query.shape
+        sk = key.size(1)
+        nh = self.num_heads
+        hn = self.hidden_size_per_attention_head
```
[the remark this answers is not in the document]
fduwjj: Will add check in the self attention part.

spmd/tensor/parallel/multihead_attention_tp.py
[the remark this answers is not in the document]
fduwjj: Add a TODO for adding a test to test equivalence.

spmd/tensor/parallel/_view_with_dim_change.py
```diff
@@ -0,0 +1,103 @@
...
+from spmd.tensor.placement_types import Shard
+from typing import Tuple, Union
+
+
+def _view_with_sharding_dim_change(
+    tensor: Union[torch.Tensor, DT], sharding_dim: int, shape: Tuple[int, ...]
+) -> Union[torch.Tensor, DT]:
+    """
+    We change the implicit sharding dim for a distributed tensor without comms.
+    Because if we don't change sharding dim, we will end up having an invalid DTensor.
```
wanchaol, line 15: "Because if we don't change sharding dim, we will ended up having more comms that are not necessary. Note that this op will produce invalid DTensor, you will need to call this op in pair to recover it back to a valid DTensor."

wanchaol (approved):
lgtm, one suggestion about the comment.

PR #478, opened 2022-09-19
[dtensor] enable scatter to handle uneven case

wanchaol:
* #492
* #491
* __->__ #478

This PR added support to enable mesh.scatter handle the case where
the tensor to scatter cann't be devided evenly, note that currently
ProcessGroupNCCL already naturally support uneven collective, but
ProcessGroupGLoo does not, where we need to speical handle the gloo
backend.

This is the second step to support uneven sharding, more collective
support will be added in follow up PRs


spmd/tensor/device_mesh.py
```diff
@@ -271,34 +271,65 @@ def scatter(
         assert my_coordinate is not None
 
         num_chunks = self.size(mesh_dim)
-        # TODO: handle uneven shard sizes
-        assert tensor_to_scatter.size(tensor_dim) % num_chunks == 0, (
```
aazzolini, line 275: Nice!

spmd/tensor/device_mesh.py
```diff
@@ -271,34 +271,65 @@ def scatter(
...
+            else:
+                scatter(
+                    tensor, scatter_list=None, src=src_for_dim, group=dim_group
+                )
+            return tensor
         else:
-            scatter(tensor, scatter_list=None, src=src_for_dim, group=dim_group)
-        return tensor
+            raise RuntimeError(
+                f"backend {self._backend} not supporting scatter!"
```
aazzolini, line 331: suggestion:

  f"backend {self._backend} does not support scatter!"

spmd/tensor/device_mesh.py
```diff
@@ -426,16 +457,23 @@ def reduce_scatter(
...
             input_list = list(input.tensor_split(num_chunks, dim=tensor_dim))
             output = torch.empty_like(input_list[my_coordinate])
             dim_group = self._dim_groups[mesh_dim]
             reduce_scatter(output, input_list, op=op, group=dim_group)
             return output
-        else:
+        elif self._backend == "gloo":
             # it's gloo, which does not have reduce_scatter
             # we have to do all_reduce + scatter
+            # warnings.warn(
```
aazzolini, line 470: Is there a reason this warning is commented out?

aazzolini (approved):

spmd/tensor/device_mesh.py
```diff
@@ -274,41 +275,69 @@ def scatter(
...
+            # resize to uneven size if needed
+            if my_rank_shape != tensor.shape:
+                tensor.resize_(my_rank_shape)
+            return tensor
+        elif self._backend == "nccl":
+            to_scatter = [tensor.contiguous() for tensor in scatter_list]
+
+            tensor = torch.empty_like(to_scatter[my_coordinate])
+            if src_for_dim == get_rank():
+                scatter(
```
fduwjj, line 326: Does this mean NCCL actually supports uneven scatter? But I remembered we did resize_ for ShardedTensor. For learning purpose, why is it working now?

wanchaol: pg nccl recently supported this I think, although gloo not, and likely uneven scatter is not performing better than even one, so I updated to do pad/unpad before/after so that it could reuse the even size collective.

spmd/tensor/device_mesh.py
wanchaol: just added this to the PR above which handles reduce_scatter, agree that it's valuable to warn it explicitly :)

PR #480, opened 2022-09-20
mark transpose is a linear op

aazzolini:
a _Partial() tensor , when transposed, remains a valid _Partial() tensor.

### DIFF
--- a/test/spmd/tensor/test_matrix_ops.py
+++ b/test/spmd/tensor/test_matrix_ops.py
@@ -3,6 +3,7 @@ import torch
 from torch.testing._internal.common_utils import run_tests
 from torch.testing._internal.common_distributed import skip_if_lt_x_gpu
 from spmd.tensor.api import DTensor
+from spmd.tensor.redistribute import redistribute_spmd_tensor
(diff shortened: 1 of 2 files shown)


test/spmd/tensor/test_matrix_ops.py
```diff
@@ -3,6 +3,7 @@
 from torch.testing._internal.common_utils import run_tests
 from torch.testing._internal.common_distributed import skip_if_lt_x_gpu
 from spmd.tensor.api import DTensor
+from spmd.tensor.redistribute import redistribute_spmd_tensor
```
wanchaol, line 6: any reason we use this function instead of directly `dc.redistribute`?

aazzolini: not really, i can change to that .

wanchaol (approved):

test/spmd/tensor/test_matrix_ops.py
wanchaol: yep, `dc.redistribute` is the public API and it deals with the bwd pass as well

PR #488, opened 2022-09-21
[spmd] pump pt version in setup.py

wanchaol:


crutcher:
The following needs to be added to make the versions resolve:
```
dependency_links=[
  "https://download.pytorch.org/whl/nightly/torch",
]
```

Or this, if we use CUDA:
`"https://download.pytorch.org/whl/nightly/cu117"`

Not sure how to make that paramaterizable yet

PR #490, opened 2022-09-23
[spmd] Add all_to_all to n-d device mesh with tests

XilunWu:
* Implemented all_to_all on n-d device mesh and added 1-d and n-d test;

* dist.all_to_all now does not have support for gloo backend. Added CPU fallback as a series of separate `scatter` op calls;


spmd/tensor/device_mesh.py
```diff
@@ -391,3 +392,22 @@ def reduce_scatter_base(
...
+        self, input_tensor_list: List[torch.Tensor], mesh_dim: int = 0
+    ) -> List[torch.Tensor]:
+        # TODO: shall we make input tensors contiguous???
+        to_broadcast = [
+            CommTensor(tensor.contiguous()) for tensor in input_tensor_list
+        ]
+        dim_group = self._dim_groups[mesh_dim]
+        # all_to_all is not supported on 'gloo'
+        if self.backend() == "gloo":
+            raise RuntimeError(
```
wanchaol, line 406: I know that all_to_all don't have a gloo backend, but we do need all_to_all like feature in CPU tests, could you add a fallback for all_to_all in gloo? i.e. In this case you can `all_gather` the tensor first and split the all_gathered tensor on the other tensor dimension, see a similar handling about reduce_scatter in Gloo [here](https://github.com/pytorch/tau/pull/461/files#diff-7011fb0a506b9cac502d4295ccb898f9a987fa42837fe0eda650143fcf2113dbR464) (where only nccl have reduce_scatter)

spmd/tensor/device_mesh.py
```diff
@@ -391,3 +392,22 @@ def reduce_scatter_base(
 
             chunks = reduced_tensor.chunk(num_chunks, dim=shard_dim)
             return self.scatter(chunks, mesh_dim=mesh_dim)
+
+    def all_to_all(
+        self, input_tensor_list: List[torch.Tensor], mesh_dim: int = 0
+    ) -> List[torch.Tensor]:
+        # TODO: shall we make input tensors contiguous???
```
wanchaol, line 399: we should always make collective inputs be contiguous tensors, otherwise the collective won't work correctly as we expected.

test/spmd/tensor/test_device_mesh.py
```diff
@@ -422,6 +423,33 @@ def test_scatter_nd(self):
             received_tensor = mesh.scatter(scattered_tensors, mesh_dim=dim)
             self.assertEqual(received_tensor, torch.ones(3, 3) * self.rank)
 
+    @with_comms
+    @skip_if_lt_x_gpu(8)  # meaning self.world_size but self is not in scope
+    def test_all_to_all_1d(self):
```
wanchaol, line 428: could you also add a nd all_to_all test similar to others?

spmd/tensor/device_mesh.py
```diff
@@ -391,3 +392,22 @@ def reduce_scatter_base(
...
+        to_broadcast = [
+            CommTensor(tensor.contiguous()) for tensor in input_tensor_list
+        ]
+        dim_group = self._dim_groups[mesh_dim]
+        # all_to_all is not supported on 'gloo'
+        if self.backend() == "gloo":
+            raise RuntimeError(
+                f"torch.distributed.all_to_all does not support {self.backend()} backend."
+            )
+        output_tensor_list = [torch.empty_like(to_broadcast[0])] * len(
```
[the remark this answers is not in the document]
XilunWu: I want to make a list of tensors that share the same size as each tensor in `to_broadcast` to hold the output of `all_to_all`. Is there any better way to do that in Python?

spmd/tensor/device_mesh.py
XilunWu: I see. So I use `dist.scatter` and `dist.gather_all` instead of calling `dist.all_to_all` directly on CPU to perform the operation, right?

spmd/tensor/device_mesh.py
XilunWu: Thx for explanation!

test/spmd/tensor/test_device_mesh.py
XilunWu: Yes. I was about to do that in a follow up PR because I thought that supporting n-d broadcast needed to change my `all_to_all` implementation but actually it seems no. So I can just add n-d test without changing the ops code.

spmd/tensor/device_mesh.py
wanchaol: Yes, that virtually looks like a transpose or what the all_to_all does

wanchaol (commented):
Looks pretty good overall! Only have a few comments

spmd/tensor/device_mesh.py
```diff
@@ -393,3 +394,43 @@ def reduce_scatter_base(
...
+
+    # TODO: rewrite interface to adopt PR #491, #492
+    def all_to_all(
+        self, input_tensor_list: List[torch.Tensor], mesh_dim: int = 0
+    ) -> List[torch.Tensor]:
+        # input tensors are expected to be congtiguous by the collective backend
+        to_scatter = [
+            CommTensor(tensor.contiguous()) for tensor in input_tensor_list
+        ]
+        output_tensor_list = [torch.empty_like(to_scatter[0])] * len(to_scatter)
```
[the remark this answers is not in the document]
wanchaol: btw looking at the [all_to_all API](https://pytorch.org/docs/stable/distributed.html?highlight=all_to_all#torch.distributed.all_to_all), it seems we can directly support uneven case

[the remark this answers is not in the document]
XilunWu: Yes we're assuming that tensors are of equal size in the list. TODO added and I'll create a new PR to add uneven test for add_to_all and see if it's natively supported on NCCL backend. Also I'll create a task to add uneven support for GLOO backend.

wanchaol (approved):
lgtm

PR #491, opened 2022-09-23
[dtensor] enable reduce_scatter to handle uneven case

wanchaol:
* #492
* __->__ #491
* #478

This PR added support to enable mesh.reduce_scatter handle the case where
the tensor to reduce and scatter cann't be devided evenly, note that currently
ProcessGroupNCCL already naturally support uneven collective, but
ProcessGroupGLoo does not, even if nccl do, it might be slower compare to
even collective, so we pad the tensor and unpad it before/after collective

This is the second step to support uneven sharding, more collective
support will be added in follow up PRs


spmd/tensor/device_mesh.py
```diff
@@ -481,10 +481,30 @@ def reduce_scatter(
...
+            # suupport uneven list, and collective perfomance on par
+            idx_start_to_pad = input.size(tensor_dim) % num_chunks
+            to_scatter = []
+            for i, scatter_tensor in enumerate(input_list):
+                if idx_start_to_pad != 0 and i >= idx_start_to_pad:
+                    scatter_tensor = self._pad_tensor_dim_by_1(
+                        scatter_tensor, tensor_dim
+                    )
+                scatter_tensor = scatter_tensor.contiguous()
+                to_scatter.append(CommTensor(scatter_tensor))
```
aazzolini, line 497: Do we need to use CommTensor explicitly here?

spmd/tensor/device_mesh.py
```diff
@@ -481,10 +481,30 @@ def reduce_scatter(
         num_chunks = self.size(mesh_dim)
         if self._backend == "nccl":
             input_list = list(input.tensor_split(num_chunks, dim=tensor_dim))
-            to_scatter = [tensor.contiguous() for tensor in input_list]
-            output = torch.empty_like(input_list[my_coordinate])
+            # gloo does not support uneven scattering, nccl supports it, but it
+            # might be slow compare to even size collective, we need to pad tensor
```
aazzolini, line 485: s/compare/compared

spmd/tensor/device_mesh.py
```diff
@@ -481,10 +481,30 @@ def reduce_scatter(
...
+            # before really calling scatter, and unpad/narrow it after the collective
+            # TODO: consider if we should remove this logic once ProcessGroupGloo
+            # suupport uneven list, and collective perfomance on par
+            idx_start_to_pad = input.size(tensor_dim) % num_chunks
+            to_scatter = []
+            for i, scatter_tensor in enumerate(input_list):
+                if idx_start_to_pad != 0 and i >= idx_start_to_pad:
+                    scatter_tensor = self._pad_tensor_dim_by_1(
```
aazzolini, line 493: does this require static shapes? e.g. when tracing, will this make shape a constant?

aazzolini (commented):
A few questions.

aazzolini (approved):

spmd/tensor/device_mesh.py
wanchaol: Yeah we need to use CommTensor explicitly there so that CommTensor could inline the fx Node to the graph.

spmd/tensor/device_mesh.py
wanchaol: I am not sure how this would entail when tracing, i think it might make the shape as a constant, let me check. I discussed this with @mrshenli and it looks like this is not a big concern since compiler teams could trace symbolic shapes?

wanchaol: update: did a simple try on uneven sharding + tracing, here is the graph it generates:
```
def forward(self, tensor_to_scatter_1):
    slice_1 = torch.ops.aten.slice.Tensor(tensor_to_scatter_1, 0, 0, 7)
    slice_2 = torch.ops.aten.slice.Tensor(tensor_to_scatter_1, 0, 7, 13);  tensor_to_scatter_1 = None
    constant_pad_nd = torch.ops.aten.constant_pad_nd.default(slice_2, [0, 0, 0, 1], 0.0);  slice_2 = None
    empty_like = torch.ops.aten.empty_like.default(constant_pad_nd, dtype = torch.float32, layout = torch.strided, device = device(type='cuda', index=1), pin_memory = False);  constant_pad_nd = None
    _tensor_constant0 = self._tensor_constant0
    scatter__default = torch.ops.c10d.scatter_.default([empty_like], [], _tensor_constant0, 0, -1);  empty_like = _tensor_constant0 = None
    comm_result = torch.distributed._spmd.comm_tensor._wrap_comm_result(scatter__default);  scatter__default = None
    getitem = comm_result[0]
    getitem_1 = getitem[0];  getitem = None
    getitem_2 = comm_result[1];  comm_result = None
    wait_comm = torch.distributed._spmd.comm_tensor._wait_comm(getitem_1);  getitem_1 = None
    slice_3 = torch.ops.aten.slice.Tensor(wait_comm, 0, 0, 6);  wait_comm = None
    mul = torch.ops.aten.mul.Tensor(slice_3, 1);  slice_3 = None
    return mul
```
it looks to me that it adds an additional node when padding:
```
torch.ops.aten.constant_pad_nd.default(slice_2, [0, 0, 0, 1], 0.0)
```
And it actually encoded some constant (i.e. pad_tuple). This should only happens in uneven sharding case, for evenly sized sharding, the graph should not change.

PR #492, opened 2022-09-23
[dtensor] enable all_gather to handle uneven case

wanchaol:
* __->__ #492

This PR added support to enable mesh.all_gather handle the
case where the tensor to all_gather cann't be devided evenly, note
that currently ProcessGroupNCCL already naturally support uneven collective,
but ProcessGroupGLoo does not, where we need to speical handle the gloo
backend.

This is the second step to support uneven sharding


kwen2501:
Thanks for the adoption 👍

spmd/tensor/device_mesh.py
```diff
@@ -396,14 +396,35 @@ def all_gather(
         Returns:
             A :class:`torch.Tensor` object
         """
+        my_coordinate = self.get_coordinate_on_dim(mesh_dim)
+        # TODO: what should happen if rank is not in the mesh?
```
aazzolini, line 400: Let's open an issue / task to consolidate the behavior here?

spmd/tensor/device_mesh.py
```diff
@@ -396,14 +396,35 @@ def all_gather(
...
-        gathered_list = [
-            CommTensor(torch.empty_like(input_tensor))
-            for _ in range(num_chunks)
-        ]
+        quot, rem = divmod(output_shape[tensor_dim], num_chunks)
+        gathered_list = []
+        for _ in range(num_chunks):
+            recv_shape = list(output_shape)
+            # create recv tensor with padded shape
+            recv_shape[tensor_dim] = quot + (1 if rem > 0 else 0)
```
aazzolini, line 411: Should we reuse local_shape() or somehow consolidate here our formula for shapes?

aazzolini (commented):
do we need to modify local_shape() as well to allow for un-even shards now?

spmd/tensor/device_mesh.py
wanchaol: yeah created a issue https://github.com/pytorch/tau/issues/511, will update the comment to link this issue as well

spmd/tensor/device_mesh.py
wanchaol: good point! Yeah we should consolidate local shape calculation, the tricky part on this is that it involves padded shape, i think we can abstract this local shape + padded logic out and make a common calculation, i'll do this in my next PR about the uneven sharding

aazzolini (approved):
Thanks for the answers and follow up !

PR #493, opened 2022-09-24
Elevating PipelineDriver logs to debug level

kwen2501:
Porting changes made by Huaqing.

Currently PipelineDriver logs are at the info level, which makes the training log too verbose. Elevating the PipelineDriver logging to "debug" level.

### DIFF
--- a/pippy/PipelineDriver.py
+++ b/pippy/PipelineDriver.py
@@ -217,13 +217,13 @@ class RankWorker(EventRecorder):
 
     def enqueue_ready_runlist(self, unique_key, work_item):
         with self.ready_runlist_cv:
-            logging.info(f'[{self.rank}] Current ready runlist keys: {self.ready_runlist.keys()}')
+            logging.debug(f'[{self.rank}] Current ready runlist keys: {self.ready_runlist.keys()}')
             self.ready_runlist[unique_key] = work_item
             self.ready_runlist_cv.notify()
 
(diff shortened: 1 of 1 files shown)


XilunWu:
Hi Ke, I have the same `pyre check` error in my PR as this one in yours:
https://github.com/pytorch/tau/actions/runs/3120890274/jobs/5061824857#step:5:54

Do we need to wait for someone to fix that to pass this check?

kwen2501:
> 

If it wasn't caused by your PR then no need to wait.

pippy/PipelineDriver.py
```diff
@@ -1301,7 +1301,7 @@ def run_until(self, predicate: Callable[[pippy.fx.Node], bool]) -> Optional[pipp
 
     def run_one(self, node):
         # TODO: hoist run() implementation
-        logging.info(f'[{self.cur_microbatch}] Issue command to run {node.format_node()}')
+        logging.debug(f'[{self.cur_microbatch}] Issue command to run {node.format_node()}')
```
XilunWu, line 1304: Issue -> issuing to keep a consistency maybe?

XilunWu (approved):
LGTM besides we can double check if there's any polishing can be done to the print info. Thx for making the logging output more readable @kwen2501 !

pippy/PipelineDriver.py
kwen2501: Will fix in new PRs.

PR #495, opened 2022-09-26
[spmd] fix linting

wanchaol:


spmd/checkpoint.py
```diff
@@ -5,7 +5,6 @@
 import torch
 import torch.distributed.distributed_c10d as c10d
 from torch import distributed as dist
-from torch.distributed import ProcessGroup
```
wanchaol, line 8: @kumpera did we remove some bindings from `torch.distributed.ProcessGroup`, and only leave them available in `c10d`?

kwen2501 (approved):
LGTM.

PR #496, opened 2022-09-26
[dtensor] add reshard suggestion to dfferent sharding on same mesh dim

wanchaol:
* __->__ #496

This add reshard suggestion for the case where different inputs shrading
on the same mesh dim. In order to perform local computation, we need to
all_gather one of the inputs on that mesh dimension. We uses a simple
heuristic to determine which input to all_gather (i.e. min comm volume)


spmd/tensor/ops/common_rules.py
```diff
@@ -147,6 +148,29 @@ def merge_sharding(dim: str, a: int, b: int) -> int:
             if value != len(input_specs):
                 needs_reshard = True
 
+    for mesh_dim, dims in seen_shardings.items():
```
aazzolini, line 151: Do we align broadcast dimensions before doing this?

aazzolini: E.g. when we are summing 

A(20,6)[shard(1)]  + B(6)[shard[0])

the shards are aligned . Just makeing sure that this is taken into account.

wanchaol: It should be handled in pointwise ops before calling the einop rule, logic [here](https://github.com/pytorch/tau/blob/main/spmd/tensor/ops/common_rules.py#L180), also added this as a test case in https://github.com/pytorch/tau/pull/496/files#diff-d12eecf13eb5e75f53a3542af9b905a09717da64a3586df228c0502a61c17492R279

test/spmd/tensor/test_dtensor_ops.py
```diff
@@ -140,10 +140,7 @@ def wrapped(fn):
     xfail("_masked.softmin"),
     xfail("_masked.softmax"),
     xfail("_masked.sum"),
-    xfail("add"),
```
aazzolini, line 143: So nice to see these xfail go away in bulk!

test/spmd/tensor/test_common_rules.py
```diff
@@ -251,6 +266,47 @@ def test_pointwise_rules_suggestion(self):
...
+        output_sharding = pointwise_rule(OpSchema((mat1_spec, mat2_spec), {}))
+        output_spec = output_sharding.output_spec
+        self.assertIsNone(output_spec)
+        self.assertIsNotNone(output_sharding.schema_suggestions)
+
+        # ensure that the suggestion is to reshard the first
+        # arg by all_gather first tensor dim sharding
+        schema_suggestion = output_sharding.schema_suggestions[0]
+        self.assertEqual(
+            schema_suggestion.args_schema[0].dim_map, [-1, -1, -1, 1]
```
aazzolini, line 306: Is this cheaper than [-1, 0, -1, 1] ?

aazzolini (commented):
Just have one question about th

test/spmd/tensor/test_common_rules.py
wanchaol: note that we can't shard to `[-1, 0, -1, 1]` because the size of this input is `torch.Size([12, 1, 1, 8])`

aazzolini: got it , so it actually would do [-1, 0, -1, 1] if the dimension was not a singleton?

aazzolini (approved):
Looks good to me, good to see all the xfail going away :)

test/spmd/tensor/test_common_rules.py
wanchaol: yeah it should be as long as this reshard is cheaper in terms of comm volume, let me see if i can add a test case for this as well.

wanchaol:
ci should be fixed after https://github.com/pytorch/tau/pull/504

test/spmd/tensor/test_common_rules.py
wanchaol: I'll land this first to unblock your work and add more tests later.

PR #497, opened 2022-09-26
pin python version for linter ci until mypy fix

wanchaol:
mypy failed on python 3.10.7, so pin python version until that mypy fix get released. https://github.com/python/mypy/issues/13627


kwen2501 (approved):
This is SUPER! Thanks!

PR #498, opened 2022-09-26
[spmd] fix typos

XilunWu:
* fixes typos which does not cause any problem in execution but hurts readability


wanchaol (approved):
lgtm, thanks for fixing this!

PR #500, opened 2022-09-27
Support Index ops

aazzolini:
Adds support for index ops such as x[:,tensor,...] and for index_select.
Currently only testing for 1D mesh. Will add tests for N-D mesh once DTensorConverter supports it.


spmd/tensor/ops/tensor_ops.py
```diff
@@ -278,3 +279,176 @@ def prop_slice_scatter(op_schema: OpSchema) -> OutputSharding:
...
+    multi_indices_spec = cast(
+        Union[DTensorSpec, List[Optional[DTensorSpec]]], multi_indices_spec
+    )
+
+    if isinstance(multi_indices_spec, list):
+        valid_indices_spec: List[Tuple[int, DTensorSpec]] = [
+            (i, a) for i, a in enumerate(multi_indices_spec) if a is not None
+        ]
+    else:
+        valid_indices_spec = [(0, multi_indices_spec)]
```
wanchaol, line 330: what's this "0" for, `multi_indices_spec` is non-zero, so we need to prepend 0 for indexing in this case?

wanchaol: Oh I see, it's the default dim to index into?

spmd/tensor/ops/tensor_ops.py
```diff
@@ -278,3 +279,176 @@ def prop_slice_scatter(op_schema: OpSchema) -> OutputSharding:
...
+        ).output_spec
+        assert isinstance(indices_output_spec, DTensorSpec)
+        indices_spec = indices_output_spec
+
+    lookup_dims = set(v[0] for v in valid_indices_spec)
+
+    need_reshard_on_values = tuple(
+        (
+            isinstance(vp, Shard)
+            and (vp.dim in lookup_dims or isinstance(ip, Shard))
```
wanchaol, line 376: could you document a bit more on the semantics of this rule, so it seems like we only reshard values_spec to replicate in the case the value's sharding dim is the lookup dim in indices?

wanchaol (approved):
lgtm, only have a few questions

spmd/tensor/ops/tensor_ops.py
```diff
@@ -278,3 +279,176 @@ def prop_slice_scatter(op_schema: OpSchema) -> OutputSharding:
...
+        OpSchema(
+            args_schema=tuple(v[1] for v in valid_indices_spec),
+            kwargs_schema={},
+        )
+    )
+    need_reshard_on_indices = indices_out.output_spec is None
+
+    # N.B.: the weird typing below is needed to make pyre happy.
+    indices_suggestion: Union[
+        List[Optional[DTensorSpec]], List[DTensorSpec]
```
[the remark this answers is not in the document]
aazzolini: I know ...

spmd/tensor/ops/tensor_ops.py
aazzolini: Yes, let me add a more thorough comment here.

spmd/tensor/ops/tensor_ops.py
aazzolini: Let me take this back. I don't think this is actually a valid input. I removed this and it simplifies things.

PR #502, opened 2022-09-27
Fix improper tuple constructor bug in instantiate_device_type_tests()

crutcher:
`only_for` should be treated as an `Optional[<Container>]`; is an optional container.

This code treats `(<V>)` as a tuple constructor, but it is not.

The resulting call was broken:
```
DEVICE_TYPE : str = "cuda"

E := (DEVICE_TYPE)
E ~= DEVICE_TYPE
E : str
```

When `instantiate_device_type_tests()>filter_desired_device_types()` calls:
```
if only_for:
  ... = filter(lambda x: x.device_type in only_for, ...)
```

the device type is checked via substring inclusion, not container inclusion.
("cuda" is a substring of "cuda")

It is entirely an accident that this works; the pytorch testing code should be enforcing a proper container here.

### DIFF
--- a/test/spmd/tensor/test_dtensor_ops.py
+++ b/test/spmd/tensor/test_dtensor_ops.py
@@ -691,7 +691,7 @@ class TestDTensorOps(DistTensorTestBase):
 
 
 # only instantiate tests for DEVICE_TYPE alone (i.e. either CPU or GPU)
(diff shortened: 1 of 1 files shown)


aazzolini (approved):
Looks good to me.

aazzolini:
Seems like there's a unit test failing, not sure how it relates to the changes.

PR #507, opened 2022-09-28
Support for new_empty, new_zeros, new_full

aazzolini:


spmd/tensor/ops/tensor_ops.py
```diff
@@ -59,6 +59,23 @@ def no_shard_prop_rule(op_schema: OpSchema) -> OutputSharding:
...
+    # this op would benefit from backward sharding propagation!
+    # Since we cannot do that yet, just return replicated
+    input = op_schema.args_schema[0]
+    size = torch.Size(cast(Sequence[int], op_schema.args_schema[1]))
+    assert isinstance(input, DTensorSpec)
+
+    return OutputSharding(
+        output_spec=DTensorSpec(
+            mesh=input.mesh,
+            placements=(Replicate(),) * input.mesh.ndim,
```
wanchaol, line 72: nit: [Replicate()]?

spmd/tensor/ops/tensor_ops.py
```diff
@@ -59,6 +59,23 @@ def no_shard_prop_rule(op_schema: OpSchema) -> OutputSharding:
     return OutputSharding(tensor_spec)
 
 
+def new_no_arg_prop_rule(op_schema: OpSchema) -> OutputSharding:
```
wanchaol, line 62: maybe call it `new_factory_rule`?

wanchaol (approved):
lgtm, thanks!

spmd/tensor/ops/tensor_ops.py
aazzolini: Do we have a preference for list vs. tuple on this one? If so, should we make typing more strict?

wanchaol: I think it's actually accepting both tuple and list, I don't have a preference on this actually, usually we use List in examples or tests, but tuple should also work

PR #509, opened 2022-09-28
Add contiguous as pointwise rule

aazzolini:
Question, what's the best way to add a test for this op?

### DIFF
--- a/spmd/tensor/ops/pointwise_ops.py
+++ b/spmd/tensor/ops/pointwise_ops.py
@@ -20,6 +20,7 @@ from spmd.tensor.ops.common_rules import linear_pointwise_rule, pointwise_rule
 
 
 linear_pointwise_ops = [
+    "aten.contiguous.default",
     "aten.div.Scalar",  # this op is linear on the first argument, and the second argument is scalar, so it fits as a linear op.
     "aten.to.dtype",
 ]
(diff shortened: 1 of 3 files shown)


spmd/tensor/ops/pointwise_ops.py
```diff
@@ -20,6 +20,7 @@
 
 
 linear_pointwise_ops = [
+    "aten.contiguous.default",
```
wanchaol, line 23: Could we make this a tensor_ops instead? I feel we don't need to do resharding for contiguous(), it should just preserve the existing sharding?

wanchaol (commented):
There're some existing tests in https://github.com/pytorch/tau/blob/815aa87e71961af87dcc85b1a85728275554e099/test/spmd/tensor/test_tensor_ops.py#L36 that test the contiguous(), maybe manually add some tests there?

spmd/tensor/ops/pointwise_ops.py
aazzolini: oh of course, here, right? https://github.com/pytorch/tau/blob/main/spmd/tensor/ops/tensor_ops.py#L80

wanchaol: Yeah that should be the right place

wanchaol (approved):
Nice!

PR #513, opened 2022-09-30
Support resharding suggestion for view_ops

aazzolini:


spmd/tensor/ops/view_ops.py
```diff
@@ -405,6 +422,25 @@ def dim_unsqueeze(ndim: int, dim: int) -> DimMap:
     return dims[:dim] + (Singleton(),) + dims[dim:]
 
 
+def dim_reduction(
```
wanchaol, line 425: I think we should probably move this to the math_ops.py and register `torch.var` there? it's a bit weird to see torch.var in view_ops.py

spmd/tensor/ops/view_ops.py
```diff
@@ -405,6 +422,25 @@ def dim_unsqueeze(ndim: int, dim: int) -> DimMap:
...
+) -> DimMap:
+    """
+    General fallback for reduction ops where _Partial() does not apply.
+    This will cause incoming tensor to be replicated on the reducing dimensions.
+    """
+    if dim_or_dims is None:
+        dim_or_dims = tuple(range(ndim))
+    if isinstance(dim_or_dims, int):
+        dim_or_dims = (dim_or_dims,)
+    dim_or_dims = tuple(d if d >= 0 else d + ndim for d in dim_or_dims)
```
wanchaol, line 436: is this normalize_dims doing?

spmd/tensor/ops/view_ops.py
```diff
@@ -405,6 +422,25 @@ def dim_unsqueeze(ndim: int, dim: int) -> DimMap:
...
+    return tuple(
+        InputDim(i) if i not in dim_or_dims else Singleton()
```
wanchaol, line 438: Is `InputDim` and `DimMap` needed for things like reduction in general? If so, maybe we should move the DimSpec and related utils to a common place and let them be used in different op impls?

spmd/testing/common_utils.py
```diff
@@ -31,6 +36,35 @@
 TEST_GPU_NUM = 4
 
 
+@dataclass
+class RedistributeProfile:
+    num_calls: int
+
+
+@contextmanager
+def redistribute_profiler() -> Generator[RedistributeProfile, None, None]:
```
wanchaol, line 45: This is nice!!

spmd/tensor/ops/view_ops.py
```diff
@@ -566,30 +656,66 @@ def reshape_prop(op_schema: OpSchema) -> OutputSharding:
...
+                    args[: spec.shape_argnum]
+                    + (tuple(local_out_shape),)
+                    + args[cast(int, spec.shape_argnum) + 1 :]
+                )
+
+            return OutputSharding(output_spec=output_dtensor_spec)
 
-        return OutputSharding(output_spec=output_dtensor_spec)
+        else:
+            # TODO: optimize this. we shouldn't simply blindly replicate
```
wanchaol, line 692: is this suggesting we might switch sharding dims instead of replicating the unshardable dims?

wanchaol (commented):
Looks great! Have some questions about generalizing the dim spec thing so that other reduction ops can use.

spmd/tensor/ops/view_ops.py
aazzolini: deals with negative dimensions like -1, -2 etc.

spmd/tensor/ops/view_ops.py
aazzolini: it's definitely a possibility -- maybe we can land this one wihtout the var first and do that next?

spmd/tensor/ops/view_ops.py
aazzolini: i think would be in general more cost-effective (at least memory-wise), but not sure? didn't think it through.

spmd/tensor/ops/view_ops.py
aazzolini: makes sense -- what about i remove this for now and then add it bakc on the next PR ?

spmd/tensor/ops/view_ops.py
wanchaol: yeah, i think a general reduce_scatter is better and looks similar to zero 3 from memory prospective.

spmd/tensor/ops/view_ops.py
wanchaol: I recalled there're multiple callsites of normalize dims to deal with negative dims, maybe we should make this a util function instead?

spmd/tensor/ops/view_ops.py
wanchaol: yeah sounds good to me!

wanchaol (approved):
stamp to unblock.

PR #516, opened 2022-09-30
Land 2D parallel support

kumpera:
This is a huge collection of changes to DCP. In no particular order:

Two kitchen-sync Planners that effectively replace the default ones from PTD and provide the following set of features:

1)nested state dict handling.
This is a quite common scenario and is essential to enable any optimizer state checkpointing.

2)nested ShardedTensor flattening
This is the core feature that enables 2D checkpointing as FSDP will produce them in 2d mode

- dedup replicated tensors
This happens when using DT with Replicated placement.

FSDP-based optimizer loading.
load_sharded_optimizer_state_dict handles 1D and 2D loading transparently.
It won't work when FSDP is not used so that needs to be addressed.

Finally, it's a tech demo of how to compose multiple checkpoint transformations.

### DIFF
--- a/spmd/checkpoint/__init__.py
+++ b/spmd/checkpoint/__init__.py
@@ -1,4 +1,17 @@
-from .utils import (
-    traverse_state_dict,
-    print_tensor
+# Copyright (c) Meta Platforms, Inc. and affiliates
+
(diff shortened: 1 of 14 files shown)


wz337 (approved):
LGTM!

PR #517, opened 2022-09-30
Fix print_readable

yunxing:
Fix print_readable by recursively calls into itself and print submodule codes.

### DIFF
--- a/pippy/fx/graph_module.py
+++ b/pippy/fx/graph_module.py
@@ -713,8 +713,7 @@ class {module_name}(torch.nn.Module):
     def __copy__(self):
         return GraphModule(self, self.graph)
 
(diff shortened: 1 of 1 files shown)


kwen2501:
Hi @yunxing, thanks for your contribution!

I'd like to share some context about pippy/fx:
It is a fork of torch.fx, with very little difference that can unblock our development but is temporarily hard to land in torch.fx (for example, if the difference breaks some existing applications for which the reason is yet to be investigated).
Thus, we try to minimize our difference from torch.fx as we may move back to it one day.

Back to your PR, if it is important to you, can you please consider pulling it against torch.fx? Then we will pull it down to pippy.fx.

PR #519, opened 2022-10-02
[dtensor] Enforce sharding for inplace/out_variant ops

wanchaol:
* (to be filled)

For inplace ops and ops with out variant, the correct behavior is to
preserve the sharding of the DTensor we write into, as it matches to
the case of shape property too. In this PR, we enforce the sharding
for inplace/out_variant ops by introducing a enforce sharding check for
einop/pointwise rule, where it takes a enforce sharding dict and use
that to do sharding propagations

### DIFF
--- a/spmd/tensor/dispatch.py
+++ b/spmd/tensor/dispatch.py
@@ -46,6 +46,7 @@ class OpSchema(object):
     Here are some basic rules about what can be used and what can be changed.
 
     Args:
+        func_schema: the function schema of the operator
         args_schema: contains args except that the DTensor args have been replaced
             with its DTensorSpec
         kwargs_schema: contains kwargs except that the DTensor kwargs have been replaced
@@ -55,6 +56,7 @@ class OpSchema(object):
         - every attribute within this class could be read to conduct
           sharding propagation.
     What can be changed:
+        - only the args_schema and kwargs_schema could be changed.
         - every non-tensor args could be changed to accomodate for local tensor
           operations (i.e. for ops like view/reshape/...)
         - every "DTensorSpec" attribute inside `args_schema`, `kwargs_schema` and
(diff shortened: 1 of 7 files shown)


aazzolini:
Looks good to me . One concern is that we're changing the signature of the propagation rules. Mostly two concrete concerns:
-  I'd rely on @mrshenli to confirm that the new signature is compatible with the traced mode. (ie. we can extract signature from the traced graph).
- When we're decomposing a larger rule into smaller rules (e.g. decompose addmm into add(mm)) we'll have a harder time to construct the inputs to the underlying rule. Could we devise a helper to make this easier, if possible? Or maybe we should implement auto-sharding propagation decomposition?

wanchaol:
> * I'd rely on @mrshenli to confirm that the new signature is compatible with the traced mode. (ie. we can extract signature from the traced graph).

Yeah agree that we should do some test with tracing and ensure it works. 

> * When we're decomposing a larger rule into smaller rules (e.g. decompose addmm into add(mm)) we'll have a harder time to construct the inputs to the underlying rule. Could we devise a helper to make this easier, if possible? Or maybe we should implement auto-sharding propagation decomposition?

Yeah agreed, today there're only addmm and baddmm decompositions, which I manually pass over the `addmm` function schema to add/mm rules within the addmm rule. My feeling here on decomposition is that we need a more canonical approach than the current one. For the decomposition_table approach that you implemented, the function_schema shouldn't be an issue iiuc? because we basically goes through _torch_dispatch_ again and it would automatically parse the func schema of the decomposed ops.

aazzolini (approved):
Looks good to me. Moving forward we should devise a mechanism that would allow to refactor OpSchema without having to refactor all the propagation rules -- this will become important if we plan to open the propagation rule API to external users.

PR #523, opened 2022-10-03
[spmd] Self attention module sharding api

XilunWu:
Background:
#477 enabled tensor parallelism on MLP (Multilayer Perceptron) and Multi-head Attention (though right now only Self Attention is supported). 

Goal:
* Make this ad-hoc enablement a set of public APIs so that users of `spmd` can easily shard their models that contain MLP or Multi-head Attention. 

* For MLP module sharding, we look for the pattern of MLP within each module node's immediate children and shard.

* For Self Attention, any `torch.nn.MultiheadAttention` will be replaced with `TensorParallelMultiheadAttention` which gets further sharded.

* Verify if our custom `TensorParallelMultiheadAttention` can work as an equivalent replacement to `torch.nn.MultiheadAttention`. If not, why?


spmd/tensor/parallel/__init__.py
```diff
@@ -2,3 +2,7 @@
 from spmd.tensor.parallel.multihead_attention_tp import (
     TensorParallelMultiheadAttention,
 )
+
+from spmd.tensor.parallel.api import (
+    my_shard_self_attn,
```
wanchaol, line 7: nit: rename to `tp_shard_self_attn`?

spmd/tensor/parallel/api.py
```diff
@@ -0,0 +1,124 @@
...
+                        functools.partial(_gradient_hook, module.bias)
+                    )
+
+    def replicate_input(inputs):
+        DTensors = []
+        for tensor in inputs:
+            DTensors.append(DTensor.from_local(tensor, device_mesh, replicate))
+        return tuple(DTensors)
+
+    dist_mod = distribute_module(
```
wanchaol, line 92: why we call dist_mod again in the shard_fn? I feel like we don't need to.

spmd/tensor/parallel/api.py
```diff
@@ -0,0 +1,124 @@
...
+                if module.bias is not None:
+                    replicated_bias = nn.Parameter(
+                        distribute_tensor(module.bias, device_mesh, replicate)
+                    )
+                    module.register_parameter("bias", replicated_bias)
+                    module.bias.register_hook(
+                        functools.partial(_gradient_hook, module.bias)
+                    )
+
+    def replicate_input(inputs):
```
wanchaol, line 86: this should be passed as input of `distribute_module` in the tests.

spmd/tensor/parallel/api.py
```diff
@@ -0,0 +1,124 @@
...
+    return module
+
+
+def _gradient_hook(param, grad):
+    param._local_tensor.grad = grad._local_tensor
+
+def _shard_custom_multi_head_attn(m, device_type, tp_size):
+    # note: do we really need start_idx since it's 0???
+    start_idx = 0
+    device_mesh = DeviceMesh(
```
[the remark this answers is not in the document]
XilunWu: Yeah. Feel the same. Right now I'm using `mesh.size(dim=0)` but not sure if this holds generically for all kinds of meshes.

spmd/tensor/parallel/api.py
XilunWu: rewritten.

spmd/tensor/parallel/api.py
XilunWu: Agreed. Now `replicate_input` is also a public API.

XilunWu:
Progress Update:

1.  Made `shard_self_attn` and `replicate_input` a public API. 

2. converted `test_self_attn_megatron_e2e` test from #477 into `test_self_attn_megatron_e2e_0` with the new API. This test compares our customized Multihead Attention `TensorParallelMultiheadAttention` with its sharded version.
> Status: test passed but small numeric error observed.

3. added `test_self_attn_megatron_e2e_1`. This test compares `TensorParallelMultiheadAttention` with a `torch.nn.MultiheadAttention` that is later replaced with a sharded `TensorParallelMultiheadAttention`.
> Status: test passed on 1st iteration but failed on 2nd. small numeric error observed and accumulated. 
> Analysis: Different part compared to the above test is the initial value of parameters. `TensorParallelMultiheadAttention` and `torch.nn.MultiheadAttention` initialize in different ways and in this test we use that of `torch.nn.MultiheadAttention` and copy the parameters into `TensorParallelMultiheadAttention`. However, this doesn't mean that the numeric error won't accumulate in the above test when # of iterations goes up.

4. added `test_self_attn_megatron_e2e_2`. This test compares `torch.nn.MultiheadAttention` with a `torch.nn.MultiheadAttention` that is later replaced with a sharded `TensorParallelMultiheadAttention`.
> Status: this one is blocked by the above test. Plan to proceed when the above tests get passed.

TODO: 

1.

spmd/tensor/parallel/multihead_attention_tp.py
```diff
@@ -251,3 +251,35 @@ def forward(
...
+        if that.in_proj_weight is not None:
+            self.qkv.register_parameter("weight", that.in_proj_weight)
+        if that.in_proj_bias is not None:
+            self.qkv.register_parameter("bias", that.in_proj_bias)
+        if that.out_proj.weight is not None:
+            self.proj.register_parameter("weight", that.out_proj.weight)
+        if that.out_proj.bias is not None:
+            self.proj.register_parameter("bias", that.out_proj.bias)
+        """
+        x = that.in_proj_weight.data.clone().detach()
```
awgu, line 270: Just passing by and leaving a nit:
I think all of these `.data` calls are unnecessary if followed by `.clone().detach()`. (One way to view this: At the aten op level, `.data` becomes `aten.detach.default`, which would be redundant with the later `detach()` call.)

XilunWu: Thx @awgu for the explanation!! I'll put it in my note. ;-)

XilunWu:
Currently, the correctness check cannot pass due to potential `TensorParallelMultiheadAttention` internal issue #540 #541 after 1st iteration in training. We plan to bypass it by loosing the check to only 1st iteration which is enough to show that the module replacement and sharding logic is correct.

TODO:
1. make _aggregate_local_tensor a public API.
2. remove unnecessary manual hook registration.
3. polish test cases.

spmd/tensor/parallel/api.py
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates
+import torch
+import torch.nn as nn
+import functools
+from typing import Sequence, Tuple, Callable
+from spmd import (
```
wanchaol, line 6: nit: let's tensor.parallel should ideally only depends on `tensor` subpackage, let's directly import from `spmd.tensor` instead.

spmd/tensor/parallel/api.py
```diff
@@ -0,0 +1,100 @@
...
+    return functools.partial(_replicate_input, device_mesh=device_mesh)
+
+
+def aggregate_output(output: DTensor) -> torch.Tensor:
+    if isinstance(output, DTensor):
+        replica_placement = [Replicate()]
+        return (
+            output.redistribute(output.device_mesh, replica_placement)
+            .to_local()
+            .contiguous()
```
wanchaol, line 99: do we really need the output be contiguous?

spmd/tensor/parallel/api.py
```diff
@@ -0,0 +1,100 @@
...
+    return functools.partial(_shard_self_attn, device_mesh=device_mesh)
+
+
+def replicate_input(
+    device_mesh: DeviceMesh,
+) -> functools.partial[Tuple[DTensor, ...]]:
+    return functools.partial(_replicate_input, device_mesh=device_mesh)
+
+
+def aggregate_output(output: DTensor) -> torch.Tensor:
```
wanchaol, line 93: maybe call this `replicate_output` since that matches the behavior here?

spmd/tensor/parallel/api.py
```diff
@@ -0,0 +1,100 @@
...
+    if isinstance(output, DTensor):
+        replica_placement = [Replicate()]
```
wanchaol, line 95: nit: `[Replicate()] * output.device_mesh.ndim`

wanchaol (approved):
Awesome work! This lgtm, only have some nit suggestions.

awgu (commented):
Hi @XilunWu. Thanks for the nice PR! I am trying to learn a bit about the DT effort, so I am occasionally reading some PRs. I hope you do not mind :)

test/spmd/tensor/parallel/test_tp_examples.py
```diff
@@ -371,5 +364,164 @@ def test_self_attn_megatron_e2e(self):
         self.assertEqual(output, output_tp)
 
 
+    # TensorParallelMultiheadAttention == dist_module(torch.nn.MultiheadAttention)
```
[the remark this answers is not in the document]
XilunWu: Ah that's a good question. Your understanding is correct, `MultiheadAttnWrap` is simply a wrapper over `torch.nn.MultiheadAttention`. In short, we wrap `MultiheadAttention` because `_shard_self_attn` cannot replace the top level module in the module tree with our custom `TensorParallelMultiheadAttention` module.

In detail, if you look into the `_shard_self_attn` (https://github.com/pytorch/tau/blob/self-attn/spmd/tensor/parallel/api.py#L26) which we use to parallelize each `TensorParallelMultiheadAttention`/`torch.nn.MultiheadAttention` module in tree, you'll see that we search the immediate child nodes (`module.named_children()`) to find out those target modules, replace with `TensorParallelMultiheadAttention` if the module is a `torch.nn.MultiheadAttention`, and parallelize it in case of a `TensorParallelMultiheadAttention` module. The problem happens when we need to replace a module. What we currently adopt is, we register the new module under the same name within its parent module, and this determines that we cannot replace the top-level module whose parent module doesn't exist. However, unlikely will people use this API to parallelize a sole `torch.nn.MultiheadAttention` (simply not its use case) so we can just wrap it in our test. In real world cases where `torch.nn.MultiheadAttention` is used as a component, the wrap is not needed.

spmd/tensor/parallel/api.py
```diff
@@ -0,0 +1,98 @@
...
+    DeviceMesh,
+)
+from spmd.tensor import Placement
+from spmd.tensor.parallel import TensorParallelMultiheadAttention
+
+
+def _replicate_input(
+    inputs: Sequence[torch.Tensor], device_mesh: DeviceMesh
+) -> Tuple[DTensor, ...]:
+    DTensors = []
```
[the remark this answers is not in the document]
XilunWu: good point!

spmd/tensor/parallel/api.py
XilunWu: Good question. I created a new task to later investigate this but I'll keep it as `contiguous` in this PR.

test/spmd/tensor/parallel/test_tp_examples.py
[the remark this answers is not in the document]
fduwjj: If so, let's call it `DummyModule` or let's create a dummy transformer here. one MultiheadAttn + one MLP. WDYT?

fduwjj (commented):
Thanks for working on this and I left some comments here.

PR #524, opened 2022-10-04
Additional pointwise ops

crutcher:
Unify `TestDTensorOpsTest` methods.
Re-gen lagging ops db.


spmd/tensor/ops/pointwise_ops.py
```diff
@@ -28,10 +28,20 @@
 pointwise_ops = [
     # please keep the entries below alphabetically sorted
     "aten.abs.default",
+    "aten.acos.Scalar",
```
wanchaol, line 31: I think these op overloads are not actually going through `__torch_dispatch__`, so even if we added there they might not really being used, they are getting preprocessed and passed to a op that goes through dispatch op (i.e. aten.cos.default). For ops that are actually going through, they are mainly the ops in native_functions.yaml and have a `dispatch` bounded method https://github.com/pytorch/pytorch/blob/e217b30b0fc98df226654cef0617eb41a177531f/aten/src/ATen/native/native_functions.yaml#L474
`

spmd/tensor/ops/pointwise_ops.py
```diff
@@ -259,6 +333,8 @@
     "aten.nextafter.default",
     "aten.nextafter.out",
     "aten.nextafter_.default",
+    "aten.ones_like.default",
```
wanchaol, line 336: we already have ones_like in https://github.com/pytorch/tau/blob/eba6bdf1ec50cfa2d86d032b66da1a356b0eeefc/spmd/tensor/ops/tensor_ops.py#L94 which is not really a pointwise rule

spmd/tensor/ops/pointwise_ops.py
```diff
@@ -346,6 +436,8 @@
     "aten.xlogy_.OutScalar_Other",
     "aten.xlogy_.Scalar_other",
     "aten.xlogy_.Tensor",
+    "aten.zeros_like.default",
```
wanchaol, line 439: similarly, zero_like already in other rules

spmd/testing/dtensor_lagging_op_db.py
```diff
@@ -312,6 +293,25 @@
     ("lu_unpack", ""),
     ("mH", ""),
     ("mT", ""),
+    ("masked.amax", ""),
```
wanchaol, line 296: what's the pointwise op that is really enabled in this PR? it looks like we just re-generate the op db with additional entries of operator tests, but nothing get removed from this `xfail` list.

crutcher: A lot of things got removed from the xfail list.

crutcher: All of these were removed from xfail. I was given the mandate of "as many as I could"; and this was me pulling on that and related strings.

And I asked you and @aazzolini a fair number of questions.

If you don't want me contributing, ok. 

```
    xfail("clamp"),
    xfail("clamp_max"),
    xfail("clamp_min"),
    xfail("fmax"),
    xfail("fmin"),
```

PR #527, opened 2022-10-04
Standardized format script

crutcher:
Normalize format kick-off and check.

Extracted from https://github.com/pytorch/tau/pull/505 , minus the static analysis passes which aren't unified yet.


kwen2501 (commented):
The changes to the .py files look okay to me.

I only have a general question of how we plan to maintain format.sh and document usage thereof moving forward.

crutcher:
having it has got to be more documented than the current "you should know what directories to run black on" answer; right?

kwen2501 (approved):

PR #528, opened 2022-10-04
Add env PIPPY_PIN_DEVICE

### LINKED ISSUE #526: Buck run device error
buck run reported the following error:
```
[trainer1]:RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0! (when checking argument for argument weight in method wrapper__native_layer_norm)
```

kwen2501:
When set to 1, pin RPC process to a single CUDA device.
Default value = 1 (same as before).

Adding this env to accept different preferences.

resolves #526


anj-s (approved):
Can you add a comment as to why we need the env var and which user should set it.

kwen2501:
> Can you add a comment as to why we need the env var and which user should set it to 0

Good idea! Added! Thanks!

PR #530, opened 2022-10-04
Disable FX tests in CI

kwen2501:
CI broke because FX is updated but tests are not.
Disable FX tests in CI.


anj-s (approved):

anj-s:
@kwen2501  any ideas on if we can depend on something that is more stable than error messages for this test?
Do we now need to update the message?

PR #536, opened 2022-10-06
Introducing SPMD API

mrshenli:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #536

PR #537, opened 2022-10-06
Add immutable_list support to enable tracing aten.sum.dim_IntList

mrshenli:
* #536
* __->__ #537
* #538


spmd/tensor/ops/utils.py
```diff
@@ -41,7 +42,9 @@ def wrapper(impl):
 
 
 def as_list(x: Union[List[object], object]) -> List[object]:
-    if type(x) is list:
+    if type(x) is list or isinstance(
+        x, torch.fx.immutable_collections.immutable_list
```
wanchaol, line 46: Curious what is immutable_list in fx and why we need to add it here for tracing, maybe add a comment?

mrshenli: yep, I should have also marked this as WIP. 

During tracing, I found that `as_list` treats `immutable_list` as an object. `aten.sum.dim_IntList` used that for its args. As a result, `as_list` turns `[0]` into `[[0]]` and hit error later when trying to use `[0]` as the index. 

I will update this one, when the PR on top is finished.

wanchaol (approved):
lgtm, thanks for adding comment!

PR #539, opened 2022-10-06
[dtensor] move dropout to use prop rule

wanchaol:
This PR moves dropout to use prop rule

### DIFF
--- a/spmd/tensor/ops/math_ops.py
+++ b/spmd/tensor/ops/math_ops.py
@@ -2,7 +2,7 @@
 from typing import cast
 
 from spmd.tensor.api import DTensor
-from spmd.tensor.placement_types import DTensorSpec, Replicate, _Partial
+from spmd.tensor.placement_types import DTensorSpec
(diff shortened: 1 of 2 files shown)


spmd/tensor/ops/math_ops.py
```diff
@@ -45,3 +45,23 @@ def softmax_bwd_rule(op_schema: OpSchema) -> OutputSharding:
             "Cannot run _softmax_backward_data on sharding dimension!"
         )
     return pointwise_rule(op_schema)
+
+
+@register_prop_rule("aten.native_dropout.default")
```
aazzolini, line 50: Isn't dropout a pointwise rule? Are we specialcasing because of the non-deterministic behavior?

aazzolini (approved):

spmd/tensor/ops/math_ops.py
wanchaol: yeah we are specializing because of non-deterministic behavior, but it is indeed a pointwise rule, let me move it to pointwise file instead and add a comment.

PR #542, opened 2022-10-07
Add split into nstages API

kwen2501:
### Added API
```
Split a model into given number of stages, based on equal stage size
Input:
  mod: `torch.nn.Module` to split
  nstages: number of stages to split the module into
Output:
  a `fx.GraphModule` transformed from the input module with `pipe_split` inserted
```

```
def split_into_nstages_equal_size(
    mod: torch.nn.Module,
    nstages: int = 1,
) -> pippy.fx.GraphModule
```

### Implementation
Using the previous `split_on_size_threshold` API as a helper to implement it --
We calculate a total model size, divide it by the number of stages wanted, and use the divided size as the bucket threshold.

### Test
Modified `local_test_autosplit.py` to test this API as well. In the meantime, moved a couple code blocks as a common function.

### Future plan
In longer term, the implementation can be improved to be more like "dynamic programming", but it can be done behind the scene without introducing user-facing changes.

### DIFF
--- a/pippy/ModelSplit.py
+++ b/pippy/ModelSplit.py
@@ -54,6 +54,7 @@ Split a model based on a maximum number of parameter and buffer elements a pipel
 Input:
   mod: `torch.nn.Module` to split
   threshold: maximum number of parameter and buffer elements a stage can have
+  max_stages: maximum number of stages; default = -1, no limit
 Output:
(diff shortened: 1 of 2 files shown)


kwen2501:
The CI failure does not seem related.

PR #543, opened 2022-10-10
[spmd] expose public APIs

wanchaol:
use __all__ to expose public APIs in the package level


spmd/__init__.py
```diff
@@ -1,6 +1,5 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
-from typing import Optional, Callable, Tuple, Union
-import torch
+from typing import Optional, Callable
```
XilunWu, line 2: just being curious, why `Tuple` and `Union` were imported even it wasn't used before?

spmd/tensor/__init__.py
```diff
@@ -77,3 +77,13 @@ def distribute_tensor(
         placements,
         requires_grad=tensor.requires_grad,
     )
+
+
+# All public APIs from dtensor package
+__all__ = [
```
XilunWu, line 83: Will adding them to multiple packages introduce confusion to users? For example, `spmd.DTensor` and `spmd.tensor.DTensor`.

test/spmd/tensor/test_dtensor.py
```diff
@@ -8,7 +8,8 @@
     DistTensorTestBase,
     with_comms,
 )
-from spmd.tensor import DeviceMesh, DTensor, Replicate, Shard, _Partial
+from spmd.tensor import DeviceMesh, DTensor
+from spmd.tensor.placement_types import _Partial, Replicate, Shard
```
XilunWu, line 12: Move `Replicate` and `Shard` to `spmd.tensor` import maybe?

XilunWu (approved):
LGTM. Thx for exposing public APIs to top level package and hiding non-public ones.

spmd/tensor/__init__.py
```diff
@@ -4,7 +4,7 @@
 import torch
 from spmd.tensor.api import DTensor
 from spmd.tensor.device_mesh import DeviceMesh, get_global_device_mesh
-from spmd.tensor.placement_types import Placement, Shard, Replicate, _Partial
+from spmd.tensor.placement_types import Placement, Shard, Replicate
```
fduwjj, line 7: Is it possible we only allow user to import these three from just one place? Or this is just internal and in public users should always do from spmd.tensor import Placement, Shard, Replicate?

Another question, will this be part of TP beta release? I guess this belongs to DTensor right?

spmd/__init__.py
wanchaol: idk tbh.. maybe the import lint check ignoring `__int__.py`

spmd/tensor/__init__.py
wanchaol: good point, I'll remove the spmd.DTensor import.

test/spmd/tensor/test_dtensor.py
wanchaol: it's test so I would like to keep placement type import be consistent because it also need to import `_Partial`.

spmd/tensor/__init__.py
wanchaol: It's internal and in public user always suppose to import from `spmd.tensor` instead of submodules.

For the second question, no, I think TP beta release should not expose the placement types.

spmd/__init__.py
```diff
@@ -84,3 +84,14 @@ def replicate_module_params_buffers(m: nn.Module, mesh: DeviceMesh) -> None:
...
+
+
+# All public APIs from spmd package
+__all__ = [
+    "DeviceMesh",
+    "Schema",
+    "SPMD",
+    "distribute_module",
+    "Shard",
+    "Replicate",
```
aazzolini, line 96: What is the reason for having Shard and Replicate exposed both under spmd and spmd.dtensor?

aazzolini (approved):
Looks good, just one question about same API being exposed from two different places.

spmd/__init__.py
wanchaol: I think that's because user facing API needs to construct a `Schema` in SPMD, and user need to use `Shard` and `Replicate`? But it's a good point, maybe user can directly import from `spmd.tensor` for those placement types. Let me remove these placement APIs from spmd for now.

PR #544, opened 2022-10-10
[dtensor] supports uneven sharding

wanchaol:
This PR supports uneven sharding, we remove the constraint that it
must be sharding evenly as most of the tensors/parameters on one
tensor dimension might be not divisible to world_size

Concretely this PR:
1. Add the capability for uneven size sharding in distribute_tensor
2. Add the capability to redistribute a uneven sized tensor
3. standardize most sharding calculation to `Shard` spec directly
4. refactor DeviceMesh to support more flexible sharding in the future

TODO:
1. make `from_local` to accept uneven sharding
2. make sure empty shards be ignored during computation
3. more tests for corner cases, i.e. stride testing
4. dynamic shapes for tracing


spmd/tensor/device_mesh.py
```diff
@@ -531,102 +460,87 @@ def reduce_scatter(
...
-                    scatter_tensor = self._pad_tensor_dim_by_1(
-                        scatter_tensor, tensor_dim
-                    )
-                scatter_tensor = scatter_tensor.contiguous()
-                to_scatter.append(CommTensor(scatter_tensor))
-
-            output = torch.empty_like(to_scatter[my_coordinate])
             dim_group = self._dim_groups[mesh_dim]
-            reduce_scatter(output, to_scatter, op=op, group=dim_group)
+            fut = reduce_scatter(output, input_list, op=op, group=dim_group)
```
awgu, line 470: Curious: Does `reduce_scatter()` need to use `CommTensor` as well?

spmd/tensor/device_mesh.py
```diff
@@ -531,102 +460,87 @@ def reduce_scatter(
...
 
-            # resize to uneven size if needed
-            if idx_start_to_pad != 0 and my_coordinate >= idx_start_to_pad:
-                # tensor = tensor.narrow(tensor_dim, start=0, length=tensor.size(tensor_dim) - 1)
-                output = self._unpad_tensor_dim_by_1(output, tensor_dim)
-
-            return output
         elif self._backend == "gloo":
             # it's gloo, which does not have reduce_scatter
             # we have to do all_reduce + scatter
```
awgu, line 474: This is probably premature optimization but IIUC:

For `mesh_dim` equal to `k` and assuming equal-sized size-`n`-byte tensors in `tensor_list` , the current gloo mesh reduce-scatter allocates `(k-1) * n` bytes (via `clone()`); copies `k * n` bytes (`n` via `copy_()` and `(k-1) * n` via `clone()`); and calls `k` all-reduces.

I think by increasing the memory usage slightly, you may decrease the number of all-reduces to just `1`: You allocate `k * n` bytes at once in a single tensor (i.e. `torch.empty()`), copy `k * n` bytes into tensor (i.e. the `input_list` contents), call `1` all-reduce, and copy `n` bytes into `output`. This allocates the same number of bytes and requires copying `n` additional bytes but makes the number of all-reduces a constant `1`.

I am not sure how the single `n * k`-byte tensor allocation compares to `k` separate `n`-byte allocations in terms of memory fragmentation,. but I hypothesize that for small and medium sizes (which DT may target), the `n * k` byte allocation will be better.

kumpera:
With this change DCP will require DT to expose the bounding box of the local shard as it's becoming increasingly trickier to find it. cc @wz337

spmd/tensor/device_mesh.py
wanchaol: Thanks for the suggestion! Yeah this looks better, will update.

wanchaol:
> With this change DCP will require DT to expose the bounding box of the local shard as it's becoming increasingly trickier to find it. cc @wz337

Done this in a separate PR https://github.com/pytorch/tau/pull/550, after that landed, this PR will add more uneven testing cases for shard offsets.

spmd/tensor/utils.py
```diff
@@ -24,43 +22,30 @@ def wrap(res: object, spec: OutputSpecType) -> object:
...
         ), f"output spec does not match with output! Expected DTensorSpec, got {spec}."
-        return dtensor.DTensor(res, spec.mesh, spec.placements)
+        res = dtensor.DTensor(
+            res,
+            spec.mesh,
+            spec.placements,
+            size=spec.shape,
+            requires_grad=res.requires_grad,
+        )
+        return res
```
aazzolini, line 32: nit: inline return?

spmd/tensor/dispatch.py
```diff
@@ -119,6 +120,28 @@ class OutputSharding:
     failed_reason: Optional[str] = None
 
 
+def pack_args_kwargs_with_local_tensor(
```
aazzolini, line 123: what is the reason for moving this?

spmd/tensor/ops/common_rules.py
```diff
@@ -326,11 +326,12 @@ def reduction_rule(op_schema: OpSchema) -> OutputSharding:
         keep_dim = len(op_schema.args_schema) > 2 and bool(
             op_schema.args_schema[2]
         )
-        dim_list = as_list(op_schema.args_schema[1])
+        dims_list = cast(List[int], as_list(op_schema.args_schema[1]))
+        dim_list = normalize_dims(dims_list, input_spec.ndim)
```
aazzolini, line 330: nit: is there a reason to keep dims_list around before normalization? is that used below? if not , better remove to avoid confusion

spmd/tensor/ops/tensor_ops.py
```diff
@@ -224,6 +224,8 @@ def prop_slice(op_schema: OpSchema) -> OutputSharding:
         step = 1
     if end is None or end > input_spec.shape[dim]:
         end = input_spec.shape[dim]
+    if start < 0:
```
aazzolini, line 227: nice catch! thanks for the fix.

spmd/tensor/ops/tensor_ops.py
```diff
@@ -442,7 +444,7 @@ def place(vp: Placement, ip: Placement) -> Placement:
         value_shape = torch.Size(
             tuple(value_shape[:insert_dim])
             + tuple(indices_spec.shape)
-            + tuple(value_shape[insert_dim + 1 :])
+            + tuple(value_shape[insert_dim + len(valid_indices_spec) :])
```
aazzolini, line 447: thanks for the fix!

spmd/tensor/ops/tp_sharding_ops.py
```diff
@@ -19,11 +19,11 @@
 def dist_cat(tensor_list: List[DTensor], dim: int = 0) -> DTensor:
     local_inputs = pytree.tree_map(unwrap_local_tensor, tensor_list)
     local_tensor = torch.ops.aten.concat(local_inputs, dim=dim)
-    return DTensor(
+    return DTensor.from_local(
```
aazzolini, line 22: Was this a bug, previously?

spmd/tensor/ops/view_ops.py
```diff
@@ -411,7 +404,7 @@ def dim_squeeze(shape: Shape, dim: Optional[int] = None) -> DimMap:
     return tuple(
         InputDim(i)
         for i, s in enumerate(shape)
-        if s > 1 or (dim is not None and i != dim)
+        if s > 1 or (dim is not None and i != normalize_dim(dim, len(shape)))
```
aazzolini, line 407: nice catch!

spmd/tensor/ops/view_ops.py
```diff
@@ -221,22 +226,10 @@ def dim_flatten(ndim: int) -> DimMap:
...
-def normalize_dim(dim: int, ndim: int) -> int:
-    return dim if dim >= 0 else dim + ndim
-
-
 def dim_movedim(
     ndim: int,
-    input: Union[int, Tuple[int, ...]],
-    destination: Union[int, Tuple[int, ...]],
+    input: Union[int, Tuple[int, ...], List[int]],
+    destination: Union[int, Tuple[int, ...], List[int]],
```
aazzolini, line 232: Would Sequence work instead of Tuple | List?

spmd/tensor/parallel/_view_with_dim_change.py
```diff
@@ -37,6 +37,7 @@ def forward(  # type: ignore
         ctx.previous_placement = self.placements
         ctx.previous_device_mesh = self.device_mesh
         ctx.previous_local_shape = self.to_local().size()
+        ctx.previous_logical_shape = self.size()
```
aazzolini, line 40: Is logical shape the same as "global" shape?

spmd/tensor/__init__.py
```diff
@@ -69,34 +69,32 @@ def distribute_tensor(
...
             # scatter call could not return a tensor with correct requires_grad
             # field, as ProcessGroupNCCL refuse to take a tensor with requires_grad
             # to do inplace update! So we manually set it here
-            local_tensor.requires_grad_(tensor.requires_grad)
-            tensor = local_tensor
+            output.requires_grad_(tensor.requires_grad)
+            local_tensor = output
         elif placement.is_replicate():
-            tensor = device_mesh.broadcast(tensor, mesh_dim=idx)
+            local_tensor = local_tensor.contiguous()
```
aazzolini, line 85: Hadn't we added "contiguous" to DeviceMesh impl directly?

spmd/tensor/api.py
```diff
@@ -77,12 +82,24 @@ def forward(  # type: ignore
                 if placement.is_replicate():
                     # broadcast rank 0 tensor to all ranks
                     # only broadcast if run_check is True
+                    input = input.contiguous()
```
aazzolini, line 85: Is there a case where we wouldn't want to make it contiguous before broadcast?

spmd/tensor/api.py
```diff
@@ -77,12 +82,24 @@ def forward(  # type: ignore
...
                     device_mesh.broadcast(input, mesh_dim=idx)
 
+        # if it's not by default run_check, we assume user is certain that each
+        # rank has the same tensor shape, and we just use that to calculate the
+        # global shape
+        tensor_shape = list(input.size())
```
aazzolini, line 91: Sounds risky -- is run_check=True by default?

spmd/tensor/api.py
```diff
@@ -143,29 +161,29 @@ def __new__(
...
-                            idx
-                        )
+                    if (
+                        i != shard_dim
+                        and tensor_stride[i] >= tensor_stride[shard_dim]
+                    ):
+                        # rescale the stride by the shard size
+                        tensor_stride[i] = (
+                            tensor_stride[i] // local_size[shard_dim]
+                        ) * size[shard_dim]
```
aazzolini, line 181: How does this actually work? Are we creating a tensor with custom strides? In case the tensor is not contiguous, what happens in practice?

aazzolini: For uneven sharding, would we need to keep uneven strides?

spmd/tensor/api.py
```diff
@@ -304,8 +322,10 @@ def redistribute(
         # This API perform necessary transformations and get
         # a new DTensor with the new spec. i.e. for
         # sharding it's a reshard behavior.
-        # TODO: handle last shard uneven with padding
-        # right now we assume all local shard equal size
+        # Note that redistribute currently only supports out
+        # of place redistribution, i.e. it always create a new
+        # DTensor object and leave the original one unchanged.
+        # TODO: consider if we should offer in place redistribution
```
aazzolini, line 328: My gut feeling is that we should leave that to the compiler.

spmd/tensor/ops/utils.py
```diff
@@ -54,3 +54,19 @@ def as_list(
         return x
     else:
         return [x]
+
+
+def normalize_dim(dim: int, ndim: int) -> int:
+    return dim if dim >= 0 else dim + ndim
+
+
+def normalize_dims(
```
aazzolini, line 63: could we add quick docstring?

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+        # it might be slow compared to even size collective, we need to pad tensor
+        # before really calling the collective, and unpad/narrow it afterwards
+        # TODO: consider if we should remove this logic once ProcessGroupGloo
+        # support uneven list, and collective perfomance on par
+        assert (
+            self.dim <= tensor.ndim
+        ), f"Sharding dim {self.dim} greater than tensor ndim {tensor.ndim}"
+        assert (
+            tensor.size(self.dim) >= num_chunks
+        ), f"Tensors to be sharded on dim {self.dim} must be at least as large as the number of devices in that dimension {num_chunks}"
```
aazzolini, line 55: Is this assertion necessary now that we support uneven shards? E.g. if we have 4 devices but a tensor of size 3, we could support it by having a device holding an empty tensor (or a 1 element tensor in case with_padding=True).

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+            shard_list = []
+            for i, shard in enumerate(tensor_list):
+                if (
+                    with_padding
+                    and idx_start_to_pad != 0
+                    and i >= idx_start_to_pad
+                ):
+                    shard = self._pad_tensor(shard)
+                # input tensors are expected to be congtiguous by the collective backend
+                shard = shard.contiguous() if contiguous else shard
```
aazzolini, line 69: are there cases where we don't expect contiguous to be true but still want padding? Also is there a reason not leave contiguous for the comm backend to do in case needed?

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+                shard_list.append(shard)
+            return shard_list, idx_start_to_pad
+        else:
+            return tensor_list, idx_start_to_pad
+
+    def _pad_tensor(self, tensor: torch.Tensor) -> torch.Tensor:
+        # pad tensor by 1 on the shard dim
+        pad = [0, 0] * (tensor.ndim - self.dim)
+        pad[-1] = 1
+        return torch.nn.functional.pad(tensor, pad)
```
aazzolini, line 79: what does this lower to in terms of aten ops? Does it bake in the static tensor shape when doing tracing? This would be an issue if we would like to support dynamic shapes on tracing mode.

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+
+    def _unpad_tensor(self, tensor: torch.Tensor) -> torch.Tensor:
+        # unpad tensor by 1 on the shard dim
+        return tensor.narrow(
+            self.dim, start=0, length=tensor.size(self.dim) - 1
```
aazzolini, line 84: Does this bake in a static shape when tracing? This would be a problem if we want to support dynamic shapes.

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+        )
+
+    def local_shard_size_on_dim(
+        self,
+        size_on_dim: int,
+        num_chunks: int,
+        rank: int,
+        return_offset: bool = False,
+    ) -> Tuple[int, int]:
+        # return the local shard size and offset on a given tensor dim
```
aazzolini, line 94: turn this into a proper docstring?

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+        assert (
+            size_on_dim >= num_chunks
+        ), f"Size to be sharded on dim {self.dim} must be at least as large as the number of devices in that dimension {num_chunks}"
```
aazzolini, line 97: Same question about why not supporting size_on_dim < num_chunks since empty tensors should work ok ?

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+        return tensor.narrow(
+            self.dim, start=0, length=tensor.size(self.dim) - 1
+        )
+
+    def local_shard_size_on_dim(
+        self,
+        size_on_dim: int,
+        num_chunks: int,
+        rank: int,
+        return_offset: bool = False,
```
aazzolini, line 92: in case of tracing dynamic shapes would we be getting SymInt as input here during tracing instead of ints?

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+            local_offset_on_dim = (
+                rank * split_size + pad_idx if rank >= pad_idx else rank
+            )
+        return (local_shard_size, local_offset_on_dim)
+
+    def shard_tensor(
+        self, tensor: torch.Tensor, mesh: DeviceMesh, mesh_dim: int
+    ) -> torch.Tensor:
+        # shard and scatter a tensor on a mesh dimension (use coordinate
+        # 0 on the mesh dimension as source of truth)
```
aazzolini, line 113: Make this into a proper docstring?

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+        mesh.scatter(output, scatter_list, mesh_dim=mesh_dim)
+
+        if pad_idx != 0 and my_coordinate >= pad_idx:
+            output = self._unpad_tensor(output)
+        return output
+
+    def reduce_shard_tensor(
+        self, tensor: torch.Tensor, mesh: DeviceMesh, mesh_dim: int
+    ) -> torch.Tensor:
+        # reduce and scatter a tensor on a mesh dimension
```
aazzolini, line 134: turn into docstring

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+        )
+        output = torch.empty_like(scatter_list[my_coordinate])
+        mesh.scatter(output, scatter_list, mesh_dim=mesh_dim)
+
+        if pad_idx != 0 and my_coordinate >= pad_idx:
+            output = self._unpad_tensor(output)
+        return output
+
+    def reduce_shard_tensor(
+        self, tensor: torch.Tensor, mesh: DeviceMesh, mesh_dim: int
```
aazzolini, line 132: It feels a bit weird that this is a member of Shard class and that it takes a _non_ DTensor. 
1) does this need to be a member of Shard?
2) what really are we expecting/allowing as "tensor" argument? Is it supposed to be one _shard_ of a DTensor? i can see that but this would at most be ok as a private API. Seems too confusing/risky for end-user to use.

spmd/tensor/placement_types.py
```diff
@@ -30,6 +32,170 @@ class Shard(Placement):
...
+        )
+        # wrap with comm tensor
+        scattered_list = [CommTensor(t) for t in scattered_list]
+        output = torch.empty_like(scattered_list[my_coordinate])
+        mesh.reduce_scatter(CommTensor(output), scattered_list, mesh_dim=mesh_dim)  # type: ignore
+        if pad_idx != 0 and my_coordinate >= pad_idx:
+            output = self._unpad_tensor(output)
+        return output
+
+    def to_replicate(
```
aazzolini, line 153: Same comment here -- this feels a bit weird as a member of Shard, specially as a public API. I'd be more comfortable with an _

aazzolini (commented):
This is a great piece of work, thanks for doing this!
My main concern is whether we should expose these APIs as public under Shard. I'd feel more comfortable making them available only to the implementation of higher level methods.

spmd/tensor/placement_types.py
wanchaol: It adds an additional node like below:

`torch.ops.aten.constant_pad_nd.default(slice_2, [0, 0, 0, 1], 0.0)`

IF we want to do dynamic shapes, my reading with dynamic shapes is that tracing would generate symbolic shapes instead of static constant shapes, so we should be fine, but we would have to test how it looks like when symbolic shape tracing works.

spmd/tensor/placement_types.py
wanchaol: Yeah i think so, but I think we don't trace sharding propagation directly? We trace user model, and expand user model with dtensor expansion, which we can still run sharding propagation on python?

spmd/tensor/placement_types.py
wanchaol: The reason I keep all the sharding related logic into `Shard` directly is that the way to do sharding `tensor_split`, and padding with `tensor_split` are specific to `tensor_split`, if we want to do another type of sharding, it's easy to just extend the `Placement` directly and implement some sharding logic that is required by `distribute_tensor` and `redistribute`, (i.e. for interleaved sharding we can easily implement a interleaved placement that would work). 

Agree that this is actually all internal logic and we should not expose this to end-user, I'll make the methods be all private.

spmd/tensor/parallel/_view_with_dim_change.py
aazzolini: global makes more sense to me but i'm okay either way, as long as we're consistent.

spmd/tensor/__init__.py
aazzolini: I see, in-place support really complicates things.

spmd/tensor/api.py
aazzolini: Makes sense, given support for in-place operations.

spmd/tensor/placement_types.py
aazzolini: Got it so you did try it out and it didn't work? (I know that c10d segfaults when passed an empty tensor ... but here we're padding, so that shouldn't be a problem).

aazzolini (approved):
Thanks for the changes and for the clarifications. We can follow up on separate diffs re. 1) dynamic shapes; 2) tensors smaller than number of ranks.

PR #546, opened 2022-10-11
[spmd] update setup.py and requirements.txt

wanchaol:
This updates setup.py and requirements.txt to include newer version of
nightly build and necessary depedency

### DIFF
--- a/spmd/requirements.txt
+++ b/spmd/requirements.txt
@@ -1,10 +1,14 @@
 # Python dependencies required for release
---pre --extra-index-url https://download.pytorch.org/whl/nightly/cu113 sys_platform == "linux"
-torch >= 1.14.0.dev0; sys_platform == "linux"
-
---pre --extra-index-url https://download.pytorch.org/whl/nightly/cpu sys_platform == "darwin"
-torch >= 1.14.0.dev0; sys_platform == "darwin"
-
 expecttest
 numpy
 pyyaml
+
+--pre
+--extra-index-url https://download.pytorch.org/whl/nightly/cpu
+torch >= 1.14.0.dev0; sys_platform == "darwin"
+
+--pre
+--extra-index-url https://download.pytorch.org/whl/nightly/cu113
+--extra-index-url https://download.pytorch.org/whl/nightly/cu116
+torch >= 1.14.0.dev0; sys_platform == "linux"
+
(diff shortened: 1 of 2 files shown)


XilunWu (approved):
Thx a lot for making installation of our repo easy!

spmd/requirements.txt
```diff
@@ -1,2 +1,10 @@
 # Python dependencies required for release
-torch >= 1.13.0.dev0
+--pre --extra-index-url https://download.pytorch.org/whl/nightly/cu113 sys_platform == "linux"
```
XilunWu, line 2: Unfortunately https://pytorch.org/get-started/locally/ now only provides `cu102`, `cu116`, and `cu117`.

XilunWu (changes requested):
Thx a lot for making installation of our repo easy! One issue is, I tried to install the CUDA version PyTorch on AWS but `pip install -r requirements.txt` failed to install the correct version. Instead it installed the CPU version.

spmd/requirements.txt
wanchaol: it actually have more than these three, that's how our CI works before this change. I think for AWS node it's very tricky, I just tested even without this PR, with a clean env and do instruction exactly as https://pytorch.org/get-started/locally/  but it still can't install correct nightly (installed a stable version instead..)

XilunWu: Interesting, so it's actually a problem AWS node. 

My date point:
- clean env
- Python=3.10
- nightly/cu113 failed
- nightly/cu116 succeeded

If we only need to make sure this PR doesn't break CI, then we can keep `cu113` in the file.

fduwjj: You can ssh into the CI machine if you want to really verify that.

fduwjj (approved):
Nice, this is indeed needed. I have run into some packages missing when installing tau in a new Conda environment. This is definitely going to help this, thanks!

spmd/requirements.txt
fduwjj: @XilunWu Have you tried command here: https://pytorch.org/get-started/locally/?

XilunWu: @fduwjj Yes they have `cu116` which works on AWS. I think this is a minor issue so I'll withdraw the change request.

wanchaol: Thanks for discussing this, I did some experiment and it looks like AWS cluster can't install cu113, but CI can only install cu113, so I updated the PR to include both cu113 and cu116, so that both environment could install their corresponding version

PR #548, opened 2022-10-11
Fix ST create to support calling .cpu() on the 2D ST.

kumpera:


wanchaol (commented):
it looks like the 2d parallel integration test not getting triggered in the CI https://github.com/pytorch/tau/actions/runs/3229864318/jobs/5287642613 could you check what might be happening there?

kumpera:
> 

it's being skipped because CI has an old version of PT that doesn't support the hook

kumpera:
The bug this fixes doesn't happen unless you're using PT master

wanchaol:
> > 
> 
> it's being skipped because CI has an old version of PT that doesn't support the hook

I see. hmm i thought in the CI we should be using the lastest nightly https://github.com/pytorch/tau/blob/main/.github/workflows/spmd_gpu_tests.sh#L20

kumpera:
> 

It's odd because we skip if we don't find the type  on PT. I'll add some logging to help here.

PR #549, opened 2022-10-12
[dtensor] more error checking in distribute_tensor API

wanchaol:
* #550
* __->__ #549

more checks:
1. check if placements is same length as device_mesh.ndim
2. check if it's already passed in a dtensor, and with different
  mesh and placements


fduwjj (approved):
lgtm

PR #550, opened 2022-10-12
[dtensor] add offset calculation in DTensorSpec

wanchaol:
* (to be filled)

This is primarily used by checkpointing to accurately calculate
the offsets of each shard

### DIFF
--- a/test/spmd/tensor/test_dtensor.py
+++ b/test/spmd/tensor/test_dtensor.py
@@ -8,7 +8,7 @@ from spmd.testing.common_utils import (  # type: ignore
     DistTensorTestBase,
     with_comms,
 )
-from spmd.tensor import DeviceMesh, DTensor, Replicate, Shard, _Partial
+from spmd.tensor import DeviceMesh, DTensor, Replicate, Shard, _Partial, distribute_tensor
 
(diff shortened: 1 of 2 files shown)


test/spmd/tensor/test_dtensor.py
```diff
@@ -192,6 +199,40 @@ def test_dtensor_spec_read_only_after_set(self):
         self.assertTrue(sharded_tensor.placements is not shard_spec)
         self.assertNotEqual(sharded_tensor.placements, shard_spec)
 
+    @with_comms
+    def test_dtensor_spec_local_shard_offset(self):
+        device_mesh = DeviceMesh(
```
kumpera, line 204: It should be possible to write a test that uses local_offset and local_shape to construct a ShardedTensor out of a DT that is only sharded, right?

kumpera (approved):

test/spmd/tensor/test_dtensor.py
wanchaol: should be, local_offset + local_shape should be sufficient to construct ShardedTensor's ShardMetadata except the placement device.

PR #554, opened 2022-10-14
[dtensor] support keep_dim for reduction rules

wanchaol:
This PR add supports for keep_dim option with reduction rule, where if this arg is True, we keep the output arg and output shape with the singleton dimension preserved.


test/spmd/tensor/test_dtensor_ops.py
```diff
@@ -488,7 +488,6 @@ def wrapped(fn):
     xfail("std"),
     xfail("std_mean"),
     xfail("stft"),
-    xfail("sum_to_size"),
```
aazzolini, line 491: always nice to see these go!

aazzolini (approved):

PR #556, opened 2022-10-14
LargeModels Support: Slow Mode

pbelevich:
Add intermediate device argument for submodule materialization

### DIFF
--- a/pippy/PipelineDriver.py
+++ b/pippy/PipelineDriver.py
@@ -1193,6 +1193,7 @@ class PipelineDriverBase(torch.nn.Module):
         interleave_stages=False,
         _record_mem_dumps=False,
         checkpoint=False,
+        device: Optional[torch.device] = None,
     ):
         super().__init__()
         self.pipe = pipe
@@ -1213,6 +1214,7 @@ class PipelineDriverBase(torch.nn.Module):
         self._record_mem_dumps = _record_mem_dumps
         self.optimizer_inited = False
         self.checkpoint = checkpoint
+        self.device = device
 
     def _init_remote_executors(self):
         self.rank_worker_rrefs: Dict[int, torch.distributed.rpc.RRef] = {}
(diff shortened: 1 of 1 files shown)


kwen2501:
A comment about the scheme:
If (just an if) each `RankWorker` or `PipeStageExecutor` knows its device, can we potentially save the ask for an intermediate device argument at the `PipelineDriver` API? I am thinking if the `PipelineDriver` could just send the submodule as it is passed in by the user, and let the `PipeStageExecutors` materialize their own submodule locally. There would be some parallelism here since the executors would perform the materialization in parallel.

pippy/PipelineDriver.py
```diff
@@ -1305,12 +1307,31 @@ class ExecutorDescriptor:
         for stage_id, descr in enumerate(executor_descriptors):
             # Assign stages to rank workers in a round-robin fashion
             rank = self.all_ranks[stage_id % self.world_size]
+            if self.device is not None:
+                logging.debug(
+                    f"[root] Moving stage_id = {stage_id} mod to {self.device}"
+                )
```
kwen2501, line 1313: nit: maybe use info for the above message?

pippy/PipelineDriver.py
```diff
@@ -1305,12 +1307,31 @@ class ExecutorDescriptor:
...
                 .create_stage_executor(stage_id=stage_id, mod=descr.mod),
             )
+            if self.device is not None:
+                logging.debug(
+                    f"[root] Waiting stage_id = {stage_id} mod to be confirmed by worker"
+                )
+                while not self.remote_stage_executor_rrefs[descr.name][
+                    1
+                ].confirmed_by_owner():
+                    pass
```
kwen2501, line 1329: Maybe the above wait should apply generally? (regardless of whether device is None or not)

pippy/PipelineDriver.py
```diff
@@ -1305,12 +1307,31 @@ class ExecutorDescriptor:
...
+                logging.debug(
+                    f"[root] Deleting stage_id = {stage_id} mod on master"
+                )
+                descr.mod.to("cpu")
```
kwen2501, line 1333: Why do we need to move back to cpu before deleting?

PR #559, opened 2022-10-16
Add large-cpu mode to examples/hf/gpt2

pbelevich:
Depends on https://github.com/pytorch/tau/pull/556

### DIFF
--- a/examples/hf/gpt2/pippy_gpt2.py
+++ b/examples/hf/gpt2/pippy_gpt2.py
@@ -4,6 +4,7 @@ import inspect
 import logging
 import os
 from functools import reduce
+from typing import Optional
 
 import torch
 from transformers import GPT2LMHeadModel, GPT2Config
@@ -20,7 +21,6 @@ from pippy.visualizer import events_to_json
 
 PROFILING_ENABLED = True
 CHECK_NUMERIC_EQUIVALENCE = True
-
 
 schedules = {
     'FillDrain': PipelineDriverFillDrain,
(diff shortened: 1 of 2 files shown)


examples/hf/gpt2/pippy_gpt2.py
```diff
@@ -169,6 +153,15 @@ def run_master(_, args):
     parser.add_argument('--cuda', type=int, default=int(torch.cuda.is_available()))
     parser.add_argument('--record_mem_dumps', type=int, default=0, choices=[0, 1])
     parser.add_argument('--checkpoint', type=int, default=0, choices=[0, 1])
+    parser.add_argument('--rpc_timeout', type=int, default=1800)
+    parser.add_argument('--num_worker_threads', type=int, default=512)
+
+    parser.add_argument('--n_embd', type=int, default=None)
+    parser.add_argument('--n_layer', type=int, default=None)
+    parser.add_argument('--n_head', type=int, default=None)
```
kwen2501, line 161: nit: would it be possible to move the above three model config arguments to a json config file?
Just thinking of minimizing the diff compared to the original HF example.
For example, we support using a json file with the Translation example:
```
python -u run_translation.py --model_name_or_path t5-3b  --config_name=t5-3b_config.json ...
```
https://github.com/pytorch/tau/blob/main/examples/hf/translation/t5-3b_config.json

examples/hf/gpt2/pippy_gpt2.py
```diff
@@ -169,6 +153,15 @@ def run_master(_, args):
...
+
+    parser.add_argument('--mode', type=str, default='small', choices=['small', 'large-cpu'])
+
```
kwen2501, line 164: Please see my comment in #556 
If we can achieve that, then here we don't need to use `mode` to choose device. We would just need to pass a flag like `remote_materialize=[0,1]` to `PipelineDriver` by which `PipelineDriver` would ask each `PipeStageExecutor` to do so.

kwen2501 (approved):
LGTM.

PR #558, opened 2022-10-16
[PiPPy] Move tensor P2P to c10d

kwen2501:
```
# Turn on NCCL logging
export NCCL_DEBUG=INFO
export NCCL_DEBUG_SUBSYS=INIT,COLL
# Run test
python pippy_dynamo.py
```
Then we can see NCCL P2P logs as proof of successful migration to c10d:
```
train-st-p3dn24xlarge-1:86947:87404 [1] NCCL INFO Recv: opCount 0 sendbuff (nil) recvbuff 0x7f0126e00000 count 257536 datatype 7 op 0 root 0 comm 0x7f00ec014410 [nranks=2] stream 0x564fa15b6750
train-st-p3dn24xlarge-1:86947:87404 [1] NCCL INFO Recv: opCount 1 sendbuff (nil) recvbuff 0x7f0126efb800 count 262144 datatype 7 op 0 root 0 comm 0x7f00ec014410 [nranks=2] stream 0x564fa15b6750
train-st-p3dn24xlarge-1:86948:87111 [2] NCCL INFO Recv: opCount 0 sendbuff (nil) recvbuff 0x7febfba00000 count 257536 datatype 7 op 0 root 0 comm 0x7febb0014710 [nranks=2] stream 0x563286e79820
train-st-p3dn24xlarge-1:86948:87111 [2] NCCL INFO Recv: opCount 0 sendbuff (nil) recvbuff 0x7febfbafb800 count 257536 datatype 7 op 0 root 0 comm 0x7febb0029670 [nranks=2] stream 0x563287834720
train-st-p3dn24xlarge-1:86946:87854 [0] NCCL INFO Send: opCount 0 sendbuff (nil) recvbuff 0x7ff818efb800 count 257536 datatype 7 op 0 root 1 comm 0x7ff73c00d690 [nranks=2] stream 0x7ff8e08f07d0
train-st-p3dn24xlarge-1:86946:87540 [0] NCCL INFO Send: opCount 1 sendbuff (nil) recvbuff 0x7ff819400000 count 262144 datatype 7 op 0 root 1 comm 0x7ff73c00d690 [nranks=2] stream 0x7ff8e08f07d0
train-st-p3dn24xlarge-1:86946:87466 [0] NCCL INFO Send: opCount 0 sendbuff (nil) recvbuff 0x7ff818e00000 count 257536 datatype 7 op 0 root 1 comm 0x7ff70000a990 [nranks=2] stream 0x7ff8e0911020
[...]

PR #561, opened 2022-10-17
[PiPPy] Add TorchDynamo example to CI

kwen2501:
Per https://github.com/pytorch/tau/pull/558#issuecomment-1281230600

### DIFF
--- a/examples/TorchDynamo/pippy_dynamo.py
+++ b/examples/TorchDynamo/pippy_dynamo.py
@@ -8,7 +8,10 @@ from typing import Dict
 import torch
 import torch.autograd.profiler_legacy
 import torch.fx
-import torchdynamo
+# TorchDynamo is moved into PyTorch for PyTorch > 1.13
+# One can use TorchDynamo without installing it separately by:
+# import torch._dynamo as dynamo
+import torchdynamo as dynamo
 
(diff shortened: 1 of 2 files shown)


examples/TorchDynamo/pippy_dynamo.py
```diff
@@ -20,7 +20,10 @@
 )
 from pippy.microbatch import TensorChunkSpec
 
-import torchdynamo
+import torchdynamo as dynamo
+# TorchDynamo is moved into PyTorch for PyTorch > 1.13
```
wanchaol, line 24: Curious if there's any constraints for PiPPy to update the pinned version of pytorch and use the latest nightly?

kwen2501: I did not update the pinned version of PyTorch in this case because:
Dynamo is not a dependency of PiPPy (at least not right now).
It is just that this example uses Dynamo and PiPPy at the same time.

So, we update the required PyTorch version only when a module PiPPy depends on need an upgrade.

kwen2501:
The CI failure is unrelated.

wanchaol (approved):
lgtm

pbelevich (approved):

PR #566, opened 2022-10-19
[dtensor] add missing API documentation for DTensor class

wanchaol:
* #567
* __->__ #566


spmd/tensor/api.py
```diff
@@ -294,13 +294,38 @@ def from_local(
         )
 
     def to_local(self) -> torch.Tensor:
+        """
+        Get the local tensor of this DTensor on its current rank. For sharding it returns
+        a local shard of the logical tensor view, for replication it returns the replica on
+        its current rank.
```
fduwjj, line 300: Do we also want to mention Partial?

fduwjj (approved):
lgtm, just one nit.

Thanks!

spmd/tensor/api.py
wanchaol: We don't want to return partial directly to the end user I think, so I actually change this in the above PR to add a flag to `to_local` directly.

PR #567, opened 2022-10-19
[dtensor] add keep_partial option to `to_local` API

wanchaol:
* __->__ #567
* #566

This PR added a option `keep_partial` to the `to_local` API, which
we by default turn partial to replicate before returning the local tensor.

This way we don't need to introduce a logical/global tensor API and
keep the partial concept be hidden unless user want to explictly
turn it on.

Another thing I do in this PR is introducing a `assertDTensorEqual` so
that test cases can use that directly without writing redistribute code.


spmd/tensor/api.py
```diff
@@ -304,6 +304,22 @@ def to_local(self) -> torch.Tensor:
         """
         return ToTorchTensor.apply(self)  # pyre-ignore[16]: autograd func
 
+    def to_logical_tensor(self) -> torch.Tensor:
```
fduwjj, line 307: Can we call it "global tensor" instead?

fduwjj (commented):
Just a n00b question for this, what's the use case for this API or it's just used by test or internal use case for now?

spmd/tensor/api.py
anj-s, line 307: why do we need an API for this?

wanchaol: I was thinking about this more, and I choose to delete that API, and instead add a option to `to_local`, i think that might more sense to end user.

wanchaol: This API was intended mainly for testing usage, because there're a lot of boilerplate code that just transforms a DTensor to replication then do the comparison. I was thinking about this more and I think it might not make sense to introduce this API yet, so I instead add a helper function in DistTensorTestBase to make testing easier.

wanchaol:
closing as source of truth now lives in pytorch, will submit PR there.

PR #568, opened 2022-10-20
[spmd/TP] export MLP sharding as a public API

XilunWu:
- `shard_mlp` is introduced in #477 to enable easy module sharding (Tensor Parallelism) on MLP layers. This enables Tensor Parallelism on ViT model.
- This PR allows user to pass `shard_mlp` (renamed to `tp_shard_mlp`) as an argument to a more general API `distribute_module`. We believe this fashion (having all module-specific TP logics separate from the general TP logic) is more modular. 
- [WIP] Current `tp_shard_mlp` assumes the MLP layer has structure "Linear layer 1 -> ReLU layer -> Linear layer 2".
- [WIP] Test code need refactor.


XilunWu:
welcome comments

spmd/tensor/parallel/api.py
```diff
@@ -76,12 +76,69 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+            sharded_weight = nn.Parameter(
+                distribute_tensor(m.weight, device_mesh, col_wise_sharding)
+            )
+            sharded_bias = nn.Parameter(
+                distribute_tensor(m.bias, device_mesh, col_wise_sharding)
+            )
+            m.register_parameter("weight", sharded_weight)
+            m.register_parameter("bias", sharded_bias)
+            # note: this hook enables access to m.weight._local_tensor.grad
+            m.weight.register_hook(functools.partial(_gradient_hook, m.weight))
```
wanchaol, line 114: since this is added in the first linear layer, we better add this to the second layer as well?

spmd/tensor/parallel/api.py
```diff
@@ -76,12 +76,69 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+        and isinstance(submodules[2], nn.Linear)
+    )
+
+
+def _gradient_hook(param: DTensor, grad: DTensor) -> None:
+    param._local_tensor.grad = grad._local_tensor
+
+
+def _shard_mlp(name: str, module: nn.Module, device_mesh: DeviceMesh) -> None:
+    if _isMLP(module):
```
wanchaol, line 96: If it's not a MLP module, maybe we raise a warning here? so that user know that they are not using this API properly.

wanchaol (approved):
lgtm, have two suggestions inlined.

fduwjj (changes requested):
Thanks for working on this. I feel we need to polish a little bit before merging this to master.

spmd/tensor/parallel/api.py
```diff
@@ -76,12 +76,69 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
                 module.register_module(n, tp_multi_head_attention)
 
 
+def _isMLP(module: nn.Module) -> bool:
+    # We assume that the structure of MLP to shard defined in object way as below:
+    # Linear -> ReLU -> Linear
+    submodules = list(module.children())
+    return (
```
fduwjj, line 83: Isn't this too strong? We really don't care what's in the middle. At end of the day, users can have many submodules, we just want to verify that there are two Linear or even numbers of Linear. This might have drawback, for example, what if user call layer norm after the first layer, but that's fine for now and DTensor does cover these ops already.

spmd/tensor/parallel/api.py
fduwjj: Shall we move the hook registration out? It's ok for this PR, but down the road, we want to move it to 2D (+FSDP) specific, right?

spmd/tensor/parallel/api.py
```diff
@@ -76,12 +76,69 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+            sharded_bias = nn.Parameter(
+                distribute_tensor(m.bias, device_mesh, col_wise_sharding)
+            )
+            m.register_parameter("weight", sharded_weight)
+            m.register_parameter("bias", sharded_bias)
+            # note: this hook enables access to m.weight._local_tensor.grad
+            m.weight.register_hook(functools.partial(_gradient_hook, m.weight))
+
+        # shard Linear layer 2
+        m = submodules[2]
```
fduwjj, line 117: It's hacky to assume that the 3rd submodule to be Linear. We can just iterate through all submodules and get its index.

spmd/tensor/parallel/api.py
XilunWu: @wanchaol, I agree with @fduwjj that we get hook registration out of `tp_shard_mlp` and move it to `fsdp_integration_2d_prep`(T135842755).

spmd/tensor/parallel/api.py
XilunWu: @wanchaol We may need to raise a warning from a different place because this `_shard_mlp` function will be applied to all nodes in the module tree. One question about the usage of `distribute_module()`: 

1. Does user write `distribute_module(model, parallel_fn=tp_shard_mlp)` to parallelize all MLP submodules in model, or 
2. User write `distribute_module(MLP, parallel_fn=tp_shard_mlp)` to parallelize one MLP layer.

I think it is the first case, then we should not raise warning from `_shard_mlp` but some function at top-level node of the tree.

spmd/tensor/parallel/api.py
```diff
@@ -76,12 +76,75 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+    # We assume that the structure of MLP to shard defined in object way as below:
+    #   * -> (Linear -> * -> Linear)+ -> *
+    # positive even number of Linear layers interleaved by activation/norm/dropout
+    linear_submodules = list(
+        filter(lambda x: isinstance(x, nn.Linear), module.children())
+    )
+    return len(linear_submodules) > 0 and len(linear_submodules) % 2 == 0
+
+
+def _shard_mlp(name: str, module: nn.Module, device_mesh: DeviceMesh) -> None:
```
fduwjj, line 89: Maybe let's just directly change the name to _parallelize_mlp.

spmd/tensor/parallel/api.py
```diff
@@ -76,12 +76,75 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
                 module.register_module(n, tp_multi_head_attention)
 
 
+def _is_mlp(module: nn.Module) -> bool:
```
fduwjj, line 79: The logic here looks good to me. Actually after a second thought. I think let's do something like this:

1. _get_linear_module_count():
2. For Pairwise parallel style, we will throw errors if the count == 0 or the count % 2 != 0.

Because at the end of day, MLP literally means multiple layers of Linear. And we just need to ensure that it's flattened and error out if it is a nested version. WDT?

spmd/tensor/parallel/api.py
```diff
@@ -76,12 +76,75 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+                        col_wise_sharding,
+                    )
+                )
+                sharded_bias = nn.Parameter(
+                    distribute_tensor(
+                        cast(torch.Tensor, m.bias),
+                        device_mesh,
+                        col_wise_sharding,
+                    )
+                )
```
fduwjj, line 114: can we make the common logic into a util func? Thanks!

spmd/tensor/parallel/api.py
```diff
@@ -76,12 +76,75 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+                replicated_bias = nn.Parameter(
+                    distribute_tensor(
+                        cast(torch.Tensor, m.bias), device_mesh, replicate
+                    )
+                )
+                m.register_parameter("weight", sharded_weight)
+                m.register_parameter("bias", replicated_bias)
+
+
+# Public APIs
```
fduwjj, line 135: We should make these internal API all private except parallelize_module.

test/spmd/tensor/parallel/test_tp_examples.py
```diff
@@ -36,78 +34,6 @@ def forward(self, x):
         return self.net2(self.relu(self.net1(x)))
 
 
-def _aggregate_local_tensor(module: torch.nn.Module) -> torch.nn.Module:
```
fduwjj, line 39: Can you also update test_2d_parallel.py? Not needed to be in the same PR. Thanks!

fduwjj (commented):
Thanks for addressing the comment and it indeed looks in the right direction. Left some comments and LMK if these sound make sense or not.

spmd/tensor/parallel/api.py
XilunWu: What do you mean by "a nested version"? For simplicity, we assume that the low-level API _parallelize_mlp() is only called on a bottom-level MLP sub-module where we simply check if the parity of linear layers' count matches with the parallel style being passed.

spmd/tensor/parallel/api.py
XilunWu: or we can just call `_parallelize_linear` instead?

spmd/tensor/parallel/api.py
fduwjj: Right, but I guess what I am trying to say is that:
1. Nested I mean the module is not flat, like Sequence(Linear(), LayerNorm(Linear()), Linear()) something like this. Because children only returns all direct children and it's not recursive.
2. We should not assume MLP only has even-number of layers. Instead, this constraint should go with pairwise parallel style.

spmd/tensor/parallel/api.py
fduwjj: That will work.

XilunWu:
closed because #623 gets landed!!

PR #570, opened 2022-10-22
Fix the FSDP extension to make load_state_dict works for 2D.

fegin:
FSDP should not see the global ShardedTensor but only the local or inner ShardedTensor. This PR fixes the issue and makes 2D parallelism checkpoint work.

### DIFF
--- a/spmd/tensor/parallel/fsdp.py
+++ b/spmd/tensor/parallel/fsdp.py
@@ -309,6 +309,7 @@ def _pre_load_state_dict(
     if len(shards) == 1 and type(shards[0].tensor) is ShardedTensor:
         inner_tensor = cast(ShardedTensor, shards[0].tensor)
         shards = inner_tensor.local_shards()
+        tensor = inner_tensor
 
     return (tensor, shards if len(shards) > 0 else [])
 


wanchaol (approved):
lgtm, thanks for the quick fix!

fduwjj (approved):
Thanks for the fix, does this mean the 2D test now pass?

wz337 (approved):
Thanks for the quick fix. LGTM!

spmd/tensor/parallel/fsdp.py
```diff
@@ -309,6 +309,7 @@ def _pre_load_state_dict(
     if len(shards) == 1 and type(shards[0].tensor) is ShardedTensor:
         inner_tensor = cast(ShardedTensor, shards[0].tensor)
         shards = inner_tensor.local_shards()
+        tensor = inner_tensor
```
gnadathur, line 312: Thanks for the quick fix. A couple of questions:
1. Instead of modifying `tensor`, should we return `inner_tensor` back and let FSDP use it. In other words, should returning the inner tensor be an explicit requirement of `_pre_load_state_dict` interface ?
2. Should we create a follow up task for test coverage ?

wz337: For No2., from what I am seeing, test_2d_parallel.py and tests from this PR just merged (https://fburl.com/jm668k0k) should be able to provide good test coverage.

fegin: @gnadathur, for Q1, I think the explicit requirement is `_pre_load_state_dict` should return what FSDP internal parallelism should see. In other world, for 2D case, `_pre_load_state_dict` should make sure the global one is not seen by FSDP so inner_tensor should be returned. However, it is possible that there is no inner_tensor. For example, for a particular weight, users only apply FSDP but not tensor parallelism. In such a case, FSDP should see the whole tensor shape.

gnadathur: > For No2., from what I am seeing, test_2d_parallel.py and tests from this PR just merged (https://fburl.com/jm668k0k) should be able to provide good test coverage.

Thanks Iris. I was wondering if it is possible to write unit tests to test the hooks instead of relying on E2E tests like `test_2d_parallel`.

gnadathur: > @gnadathur, for Q1, I think the explicit requirement is `_pre_load_state_dict` should return what FSDP internal parallelism should see. In other world, for 2D case, `_pre_load_state_dict` should make sure the global one is not seen by FSDP so inner_tensor should be returned. However, it is possible that there is no inner_tensor. For example, for a particular weight, users only apply FSDP but not tensor parallelism. In such a case, FSDP should see the whole tensor shape.

Thanks for the explanation. As discussed offline, your suggestion of adding validations on the FSDP side of the hook on the returned tensor would be a reasonable aid to catching such issuse.

PR #575, opened 2022-10-24
Fix 2d parallel tests

wz337:
Need to wait for https://github.com/pytorch/tau/pull/570/  to land. 
Fix test/spmd/checkpoint/test_dedup_tensors.py and test/spmd/checkpoint/test_fsdp_optim_state.py.

Fix lint in spmd/checkpoint/optimizer.py and spmd/checkpoint/traverse.py.


test/spmd/checkpoint/test_fsdp_optim_state.py
```diff
@@ -14,13 +14,13 @@
...
 from spmd.testing.checkpoint_utils import with_temp_dir
 from spmd.testing.common_utils import (
-    TEST_GPU_NUM,
     DistTensorTestBase,
     with_comms,
 )
 from torch.testing._internal.common_distributed import skip_if_lt_x_gpu
 from torch.testing._internal.common_utils import run_tests
 
+TEST_GPU_NUM = 4
```
fduwjj, line 23: Looks like there is a decorator named "skip_unless_torch_gpu", maybe we can directly use that?

fduwjj (commented):
Thanks for the fix. Just leave a nit here.

PR #578, opened 2022-10-26
[dtensor] support torch.mean/var and refactor reduction rule to support custom reduction

wanchaol:
This PR addresses multiple issues:

1. move the partial reduction logic to the placement directly, this could enable custom reduction in the future if needed.
2. add reduction_linear flag to deal with linear reduction rules and make the reduction dim calculation explicit.
3. supports torch.mean/var with proper reduction rule
4. Add some missing pointwise ops (eq.Scalar/out, etc.)


aazzolini (approved):
Looks great, thanks for doing it . I think there's a few test failures to go through first.

PR #581, opened 2022-10-28
[dtensor] use RedOpType and unpin pt version

wanchaol:
* __->__ #581
* #582

use RedOpType instead to avoid deepcopy failure of c10d.ReduceOp


fegin (approved):
Thanks for the fix!

aazzolini (approved):

PR #582, opened 2022-10-28
[dtensor] fix a corner case in pointwise rule

wanchaol:
* #581
* __->__ #582

This fixes a corner case in the pointwise rule when handling the
singleton dims, where if it should not be "b, ,11->ab", and instead
it should be "b, ,11->1b"


aazzolini (approved):
could you add a test case?

PR #583, opened 2022-10-28
Adding t5 inference to the example 

HamidShojanazeri:
Starting to explore Pippy as distributed inference solution for PyTorch. This PR will work through this T5 example to get it working with the auto-split(model, segments) API and benchmark it for inference.


examples/hf/t5/pippy_t5.py
```diff
@@ -115,6 +123,70 @@ def resolve_pg_per_stage(pp_rank):
 
 
 def run_master(pp_ranks, args):
+
+    def train_mode(model,split_policy, MULTI_USE_PARAM_CONFIG , concrete_args):
```
anj-s, line 127: Can these be functions outside run_master? They seem independent enough.

anj-s, line 127: nit: space after "," and no space after params.

examples/hf/t5/pippy_t5.py
```diff
@@ -115,6 +123,70 @@ def resolve_pg_per_stage(pp_rank):
...
+
+        output_loss_value_spec = {'loss': True, 'logits': False,
+                                # 'past_key_values': False,
```
anj-s, line 130: what is "past_key_values" for?

examples/hf/t5/pippy_t5.py
```diff
@@ -154,52 +227,27 @@ def run_master(pp_ranks, args):
...
-    t5_input_dict = {'input_ids': inp, 'decoder_input_ids': inp,
-                     'labels': torch.empty(bs, seq_length, dtype=torch.long, device=device).random_(
-                         t5.config.vocab_size - 1)}
+
+    if args.mode == "train":
+        t5_input_dict = {'input_ids': inp, 'decoder_input_ids': inp,
+                        'labels': torch.empty(bs, seq_length, dtype=torch.long, device=device).random_(
+                            t5.config.vocab_size - 1)}
+    elif args.mode == "inference":
+        t5_input_dict = {'input_ids': inp, 'decoder_input_ids': inp}
```
anj-s, line 236: where do we load model checkpoints for inference?

examples/hf/t5/pippy_t5.py
```diff
@@ -37,6 +37,14 @@
 
 pippy.fx.Tracer.proxy_buffer_attributes = True
 
+def model_size(model_pipe):
+
+    total_params = 0
+    for i, sm in enumerate(model_pipe.split_gm.children()):
+        params = get_number_of_params(sm)
+        print(f"submod_{i} {params // 10 ** 6}M params")
```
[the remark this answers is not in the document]
HamidShojanazeri: I think as this is a simple example, prints could help to keep it educational.

examples/hf/t5/pippy_t5.py
HamidShojanazeri: They are both using the same checkpoints, in this case its not a real checkpoint, it is mostly to demonstrate functionality other than that there is no real data going toward training/ inference.

examples/hf/t5/pippy_t5.py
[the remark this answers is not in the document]
HamidShojanazeri: Regarding the prints, It was from there from the original implementation, I think as the purpose of the example is just educational that might help to keep track of what is happening, but for sure we can change them to be part of the loggings.

kwen2501:
- Further merging functions `train_mode` and `inference_mode` into one function (named `transform_into_pipeline`)
- `args.mode` changed to `args.train = 0|1`

kwen2501:
Inference:
`python pippy_t5.py --train 0`

Training:
`python pippy_t5.py --train 1`

kwen2501 (approved):

PR #589, opened 2022-10-31
[Checkpointing] Fix linter for PR 516(2D checkpointing support) and resubmit

wz337:
This PR includes:

1. changes from @kumpera (https://github.com/pytorch/tau/pull/516) adding 2D parallel support for checkpointing.
2. Fix linter and resubmit. 

```
**PR516 Summary:**

This is a huge collection of changes to DCP. In no particular order:

Two kitchen-sync Planners that effectively replace the default ones from PTD and provide the following set of features:

1)nested state dict handling.
This is a quite common scenario and is essential to enable any optimizer state checkpointing.

2)nested ShardedTensor flattening
This is the core feature that enables 2D checkpointing as FSDP will produce them in 2d mode

dedup replicated tensors
This happens when using DT with Replicated placement.
FSDP-based optimizer loading.
load_sharded_optimizer_state_dict handles 1D and 2D loading transparently.
It won't work when FSDP is not used so that needs to be addressed.

Finally, it's a tech demo of how to compose multiple checkpoint transformations.
```


spmd/checkpoint/traverse.py
```diff
@@ -0,0 +1,180 @@
...
+    Tuple,
+    TypeVar,
+    Union,
+    cast,
+)
+from torch.distributed._shard.checkpoint.metadata import (
+    STATE_DICT_TYPE,
+)
+from torch.distributed._shard.sharded_tensor.api import ShardedTensor
+from spmd import DTensor as DT
```
wanchaol, line 20: nit: `from spmd.tensor import DTensor as DT`, we don't plan to offer DTensor as public in spmd package directly

spmd/checkpoint/optimizer.py
```diff
@@ -0,0 +1,324 @@
...
+    STATE_DICT_TYPE,
+    TensorStorageMetadata,
+)
+from torch.distributed._shard.checkpoint.planner_helpers import (
+    _create_sharded_read_items,
+    _create_read_items,
+)
+from torch.distributed.remote_device import _remote_device
+
+from spmd import DTensor as DT
```
wanchaol, line 29: nit: `from spmd.tensor import ..`

wanchaol (approved):
lgtm, only have two nits regards to import. Thanks for fixing the lint and failures!

PR #587, opened 2022-10-31
[dtensor] changes required before moving to core

wanchaol:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #587

PR #591, opened 2022-11-02
make mypy linting be more consistent with core and fix lints

wanchaol:
* #587
* __->__ #591


kwen2501 (approved):
Thanks a thousand for aligning Tau with PyTorch!
I submitted a PR https://github.com/pytorch/tau/pull/594 to fix the final lint other than those silenced ones.

kwen2501:
@wanchaol can I land this one? Any concern?

wanchaol:
> @wanchaol can I land this one? Any concern?

There're a bunch of ci failures, I need to resolve them before landing, do you need this soon? If so, I'll try to fix them soon and land it

kwen2501:
I see, no hurry, please take your time. 
Just FYI -- the mypy errors are silenced, we can have more time to fix those.

check.sh
```diff
@@ -65,7 +65,7 @@ flake8 pippy spmd test/spmd
 
 # mypy spmd test/spmd
 echo; echo "Running mypy ..."
-mypy spmd pippy test examples
+mypy spmd pippy examples
```
wanchaol, line 68: @kwen2501 after restricting more mypy linting with pytorch, there emerges tons of failures in tests, imo we shouldn't require strong type annotation in tests as that will slow down the dev efficiency, let me know if this not look good to you

wanchaol: This might also stay true for examples? I am changing mypy to only run under `pippy` and `spmd`

PR #595, opened 2022-11-03
Switch to the standard reusable setup-nvidia GitHub action

huydhn:
I'm currently seeing some NVIDIA installation failure from PyTorch 

* https://github.com/pytorch/pytorch/actions/runs/3384224730/jobs/5621261503
* https://github.com/pytorch/pytorch/actions/runs/3383111924/jobs/5618889208
* https://github.com/pytorch/pytorch/actions/runs/3382298003/jobs/5617245371

where installing NVIDIA driver fails when 510.60.02 is installed instead of the standard 515.57 from PyTorch. Both use the same share runner now as part of project Nova.  So, all PyTorch repo should use the standard reusable [setup-nvidia](https://github.com/pytorch/test-infra/blob/main/.github/actions/setup-nvidia/action.yml) from test-infra repo instead of copying `install_nvidia_utils_linux.sh`


kwen2501 (approved):
Looks fantastic. Thank you so much for moving our CI to the mainstream style.

PR #605, opened 2022-11-10
isort imports, add _is_partial_dtensor(), add make_boxed_func 

lessw2020:
start to move current work to main.
This PR is minor:
a - isort imports to api.py
b - wrap return gm with make_boxed_func to keep aot happy.
c - adds _is_partial_dtensor() as refactored function to modularize check for DTensor with _Partial placement.
d - adds _build_dummy_add_graph, which is a refactored function and used to build the dummy add graph for wait_tensor insertion.  (this was added in part to fix a pyre issue where the obj reference was untyped in _convert_output).


fegin:
LGTM

fegin (approved):

spmd/compiler/api.py
```diff
@@ -45,6 +43,14 @@ class Schema:
     placements: List[Placement]
 
 
+def _is_partial_dtensor(obj: object) -> bool:
+    """check if object is 1) DTensor and  2) with placement of _Partial"""
+    is_partial = isinstance(obj, DTensor) and isinstance(
+        obj.placements[0], _Partial
```
wanchaol, line 49: this will only check if the first placement is partial? I think we should check every placement and return true if there's one partial in the placements?

spmd/compiler/api.py
```diff
@@ -149,35 +187,16 @@ def _convert_output(
...
             continue
         obj = node_to_obj[argument]
-        if not (
-            isinstance(obj, DTensor) and isinstance(obj.placements[0], _Partial)
-        ):
+
+        if not _is_partial_dtensor(obj):
             continue
 
         has_partial = True
```
wanchaol, line 194: nit: has_partial = _is_partial_dtensor(obj)?

lessw2020: good call, will fix. thanks!

spmd/compiler/api.py
lessw2020: Thanks - changed to walk entire placement list for the _Partial check now.

spmd/compiler/api.py
```diff
@@ -137,6 +149,38 @@ def unwrap_local(e: object) -> object:
     return make_fx(dispatch)(tree_map(unwrap_local, args))
 
 
+def _build_dummy_add_graph(
+    dt: DTensor, node_to_obj: Dict[fx.Node, object]
+) -> fx.GraphModule:
+    """creates a graph for a dummy add function from a partial DTensor.
+    This dummy add is then used for expansion into a CommTensor expanded graph"""
```
wanchaol, line 156: nit: the dummy add is used for triggering `all_reduce` on Partial DTensor during the DTensor expansion of the traced graph.

wanchaol (approved):
lgtm, thanks for addressing comments!

spmd/compiler/api.py
lessw2020: Thanks - will update this with the next PR.

PR #608, opened 2022-11-11
add graph_utils, part 2 

lessw2020:
updating graph_utils with some minor fixes.

### DIFF
--- a/spmd/compiler/graph_utils.py
+++ b/spmd/compiler/graph_utils.py
@@ -20,9 +20,9 @@ class OP(str, Enum):
     PLACEHOLDER = "placeholder"
 
 
-def get_node_tensor_size(node: fx.Node, debug: bool = False) -> int:
+def get_node_tensor_size(node: fx.Node, debug: bool = False) -> Optional[int]:
     """takes an fx node, and if tensor data available, optionally displays and returns numel"""
-    size: int = 0
+    size = None
     tdata = node.meta.get("tensor_meta")
     if tdata is None:
         # assert tdata is not None, f"failed to locate metadata for node {node}"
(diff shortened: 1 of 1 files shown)


spmd/compiler/graph_utils.py
```diff
@@ -24,7 +20,7 @@ class OP(str, Enum):
     PLACEHOLDER = "placeholder"
 
 
-def get_node_tensor_size(node: fx.Node, debug=False) -> int:
+def get_node_tensor_size(node: fx.Node, debug: bool = False) -> Optional[int]:
```
fegin, line 23: This debug is not necessary if it is just used to decide whether to print out the message. Instead, `rank0_debug` should be used instead of `rank0_info`.

spmd/compiler/graph_utils.py
```diff
@@ -41,12 +37,12 @@ def get_node_tensor_size(node: fx.Node, debug=False) -> int:
     return size
```
fegin, line 37: Cannot review the accidentally committed line. So I comment here. Instead of manually calculating the size, you should be able to use `size = tdata.shape.numel()`.

spmd/compiler/graph_utils.py
fegin: Also, we should change the function name to `get_node_tensor_numel` to match the PyTorch convention.

fegin (changes requested):

spmd/compiler/graph_utils.py
lessw2020: This is interacting with the node.meta dict, rather than an actual tensor though. 
Thus, we have a dict holding shape as a tuple, dtype, stride and memory format...thus have to manually calc as it's just two ints from a dict rather than a live tensor.

spmd/compiler/graph_utils.py
lessw2020: Great points, thanks!

fegin (approved):

PR #609, opened 2022-11-12
Add --chunks option to T5 example

kwen2501:
To support inference of one chunk only, for example.


HamidShojanazeri (approved):
Thanks @kwen2501 for the prompt PR.

PR #611, opened 2022-11-14
Add fusion.py with GraphInfo, FusionElement, CommType,  _scan_graph, _insert_buffer

lessw2020:
This is first part of the fusion code.  
Currently the run_fusion will populate a master graphinfo object (which accounts for the lack of tail in graph.py), 
and then does a _scan_graph to populate a list of FusionElements, including the parameter size of each element (from node meta data). 
FusionElements represent the current communication sections of the graph that will then be fused based on the optimizing algorithm.
The detection is flexible in that it uses a section pattern match with adjustable comm type.  
Assumptions are that all communication insertions are based around the 'dummy_add' function in api.py. 

The file is not currently 'live' in terms of being called by api.py yet.

### DIFF
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -27,7 +27,7 @@ class Comm_Type(str, Enum):
 @dataclass
 class FusionElement:
     comm_type: Optional[Comm_Type] = None
-    node_list: List = field(default_factory=lambda: [])
+    node_list: List = field(default_factory=lambda: [])  # type: ignore
     size: int = 0
     prev_node: Optional[fx.Node] = None  # node that was before start of section
     next_node: Optional[fx.Node] = None  # node that was after end of section
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -1,14 +1,14 @@
 from copy import deepcopy
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Optional
+from typing import List, Optional
 
 import graph_utils as gu
 import torch
 import torch.distributed as dist
 import torch.fx as fx
-
 from graph_utils import OP
+
 
 # enum for the supported fusion comm types
 class Comm_Type(str, Enum):
@@ -21,25 +21,25 @@ class Comm_Type(str, Enum):
 
 @dataclass
 class FusionElement:
-    comm_type: None
-    node_list: list = field(default_factory=lambda: [])
+    comm_type: Optional[Comm_Type] = None
+    node_list: List = field(default_factory=lambda: [])
     size: int = 0
-    prev_node: fx.Node = None  # node that was before start of section
-    next_node: fx.Node = None  # node that was after end of section
+    prev_node: Optional[fx.Node] = None  # node that was before start of section
+    next_node: Optional[fx.Node] = None  # node that was after end of section
     processed: bool = False
     output_name: str = ""
-    clone_node: fx.Node = None
-    comm_node: fx.Node = None
-    wait_node: fx.Node = None
+    clone_node: Optional[fx.Node] = None
+    comm_node: Optional[fx.Node] = None
+    wait_node: Optional[fx.Node] = None
 
 
 @dataclass
 class GraphInfo:
     len: int = 0
-    global_buffer: fx.Node = None
+    global_buffer: Optional[fx.Node] = None
     global_buffer_size: int = 0
-    output: fx.Node = None
-    first: fx.Node = None
+    output: Optional[fx.Node] = None
+    first: Optional[fx.Node] = None
 
     def update_info(self, gm, debug=True):
         """get the len, input and output nodes"""
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -22,16 +22,27 @@ from spmd.tensor.dispatch import operator_dispatch, propagate_input_sharding
 from spmd.tensor.placement_types import Placement, Replicate, Shard, _Partial
 from spmd.tensor.redistribute import _redistribute_with_local_tensor
 
+# enum for the supported fusion comm types
+class Comm_Type(str, Enum):
+    allreduce = "allreduce_"
+    allgather = "allgather_"
+    broadcast = "broadcast_"
+    reducescatter = "reduce_scatter_"
+    scatter = "scatter_"
 
-# for prepping for allreduce fusion
+
 @dataclass
-class FusionCandidateMap:
-    name: str  # TODO - make this a field with auto fill
-    node: fx.Node
-    dtensor: DTensor
-    size: int  # TODO - numel or bytes?
-    id: float  # first element of local_tensor
+class FusionElement:
+    comm_type: None
+    node_list: list = field(default_factory=lambda: [])
+    size: int = 0
+    prev_node: fx.Node = None  # node that was before start of section
+    next_node: fx.Node = None  # node that was after end of section
     processed: bool = False
+    output_name: str = ""
+    clone_node: fx.Node = None
+    comm_node: fx.Node = None
+    wait_node: fx.Node = None
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+
+        if debug:
+            print(
+                f"updated info - len = {self.len} input = {self.first}, output = {self.output}"
+            )
+
+
+def _insert_buffer_node(
+    gm: fx.GraphModule, insert_before_node: fx.Node, buffer_size: Iterable[int]
+) -> fx.Node:
```
fegin, line 76: `buffer` here is not very informative. Can we either elaborate it in the comment more or simply change it to another name (or remove the prefix)?

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+        "comm_result",
+        "getitem",
+        "getitem",
+        "wait_comm",
+    ]
+
+    fe_size = len(fe_pattern) - 1
+    curr_count = 0
+    curr_node_list = []
+    rank = dist.get_rank()
```
fegin, line 114: This is not used.

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+                fe.prev_node = fe.node_list[0].prev
+                fe.next_node = node.next
+
+                fe.output_name = node.name
+
+                fe.clone_node = curr_node_list[0]  # fe.node_list[0]
+                fe.comm_node = fe.node_list[3]
+                fe.wait_node = node
+
+                # compute size of this fe
```
fegin, line 151: The function name is meaningful enough. This comment is not needed.

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+            if node.name.startswith(pattern):
+                curr_node_list.append(node)
+                curr_count += 1
+                continue
+            else:
+                curr_count == 0
+                curr_node_list.clear()
+                continue
+
+        if curr_count == fe_size:
```
fegin, line 131: This can be `elif` which will make the code more consistent.

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+                fe.comm_node = fe.node_list[3]
+                fe.wait_node = node
+
+                # compute size of this fe
+                fe.size = get_node_tensor_numel(fe.clone_node)  # type: ignore
+                element_list.append(fe)
+
+            curr_count = 0
+            curr_node_list.clear()
+            continue
```
fegin, line 157: This continue is not needed.

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+def _scan_graph_for_fusion_elements(
+    gm: fx.GraphModule,
+    comm_type: Comm_Type = Comm_Type.allreduce,
+) -> Optional[List[FusionElement]]:
+    """scan entire graph for matching sections of CommTensor style expansions
+    returns list of FusionElements that match comm_type"""
+
+    element_list = []
+
+    fe_pattern = [
```
fegin, line 100: Can we rename this to `fe_sequence` since the order does matter?

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+        "_tensor_constant",
+        comm_type,
+        "comm_result",
+        "getitem",
+        "getitem",
+        "wait_comm",
+    ]
+
+    fe_size = len(fe_pattern) - 1
+    curr_count = 0
```
fegin, line 112: `Index` seems to be more appropriate since we are move pointer over the sequence above and decide which string to match.

fegin (approved):
LG, most comments just to keep the coding style consistent.

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+    OP,
+    get_node_tensor_numel,
+    get_output_node,
+    graph_cleanup,
+    pretty_print_graph,
+)
+
+
+# enum for the supported fusion comm types
+class Comm_Type(str, Enum):
```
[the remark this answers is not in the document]
lessw2020: agree, almost did it initially actually.

spmd/compiler/fusion.py
lessw2020: removed

spmd/compiler/fusion.py
lessw2020: fixed

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+                element_list.append(fe)
+
+            curr_count = 0
+            curr_node_list.clear()
+            continue
+
+    return element_list
+
+
+def run_fusion(main_gm: fx.GraphModule) -> bool:
```
wanchaol, line 162: shall we call this comm_fusion or collective fusion? imo we might have different lowering passes later combined (i.e. op fusion)

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+class Comm_Type(str, Enum):
+    allreduce = "allreduce_"
+    allgather = "allgather_"
+    broadcast = "broadcast_"
+    reducescatter = "reduce_scatter_"
+    scatter = "scatter_"
+
+
+@dataclass
+class FusionElement:
```
wanchaol, line 28: maybe add a docstr to explain what's a FusionElement?

spmd/compiler/fusion.py
lessw2020: updated to index

spmd/compiler/fusion.py
lessw2020: yes great point.
updated to "run_comm_fusion"

spmd/compiler/fusion.py
lessw2020: good point, added.

spmd/compiler/fusion.py
lessw2020: You are correct.  I'd like to keep it though for clarity and a tiny bit for efficiency.  (i.e. why fall down and eval various if's below it that won't be fired, when we know we want to continue here).

spmd/compiler/fusion.py
lessw2020: great point.

spmd/compiler/fusion.py
lessw2020: agree.  I've updated it to _insert_fusion_buffer_node since it will be used for the fusion operations.

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+
+@dataclass
+class GraphInfo:
+    len: int = 0
+    global_buffer: Optional[fx.Node] = None
+    global_buffer_size: int = 0
+    output: Optional[fx.Node] = None
+    first: Optional[fx.Node] = None
+
+    def update_info(self, gm: fx.GraphModule, debug: bool = True) -> None:
```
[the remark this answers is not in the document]
lessw2020: updated

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+    allgather = "allgather_"
+    broadcast = "broadcast_"
+    reducescatter = "reduce_scatter_"
+    scatter = "scatter_"
+
+
+@dataclass
+class FusionElement:
+    comm_type: Optional[Comm_Type] = None
+    node_list: List = field(default_factory=lambda: [])  # type: ignore
```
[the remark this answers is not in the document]
lessw2020: Unfortunately not. It will complain that object does not have property .prev etc. (node properties).
  I also tried Optional[List[fx.Node]] but  ultimately had to go back to # type: ignore b/c it doesn't otherwise understand that its a list of nodes, that may or may not be populated.

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+
+def _insert_fusion_buffer_node(
+    gm: fx.GraphModule, insert_before_node: fx.Node, buffer_size: Iterable[int]
+) -> fx.Node:
+    """insert a torch.empty node in front of insert_before_node"""
+    with gm.graph.inserting_before(insert_before_node):
+        new_buffer_node = gm.graph.create_node(
+            OP.CALL_FUNCTION,
+            target=torch.empty,
+            args=tuple(buffer_size),  # TODO - need device
```
anj-s, line 88: Can you add the TODO comment on top of the LOC and have more details?

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+    ), f"failed to create buffer node, size={buffer_size}"
+
+    return new_buffer_node
+
+
+def _scan_graph_for_fusion_elements(
+    gm: fx.GraphModule,
+    comm_type: CommType = CommType.allreduce,
+) -> Optional[List[FusionElement]]:
+    """scan entire graph for matching sections of CommTensor style expansions
```
anj-s, line 101: What is CommTensor style expansions?

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+    returns list of FusionElements that match CommType"""
+
+    element_list = []
+
+    fe_sequence = [
```
anj-s, line 106: It looks like this particular sequence order is important. Is that right? Or is it sufficient that the wait_comm is the last node that we look for. 

I am assuming this is to support the current allreduce fusion for DDP only? or does it support other variations?

spmd/compiler/fusion.py
```diff
@@ -0,0 +1,182 @@
...
+    allreduce = "allreduce_"
+    allgather = "allgather_"
+    broadcast = "broadcast_"
+    reducescatter = "reduce_scatter_"
+    scatter = "scatter_"
+
+
+@dataclass
+class FusionElement:
+    """This class tracks the nodes for a DTensor expanded communication collective in the graph"""
```
anj-s, line 32: Does this track all expanded collectives for the subgraphs as well? From the fe expansion list of ops, the pattern matching seemed to be for the dummy_add tensor alone.

anj-s:
Thanks @lessw2020 for breaking up the PR. Looks great :)

spmd/compiler/fusion.py
lessw2020: yes - moved comment location to above, and expanded comment to note that we need to get the device from the underlying DTensor to ensure the buffer is placed on same device as DTensor.

spmd/compiler/fusion.py
lessw2020: I believe it is important that we match the entire sequence, rather than just the wait_comm node.  
The reason is we know from the entire sequence that this is the dummy_add insertion/expansion, as opposed to a possible future insertion of only a wait_comm for other reasons. 
Longer term we may be able to revert back to the original approach of simply doing this work during the original DTensor expansion pass, but for now we are doing post expansion.  
Assuming that other collectives are inserted using the same dummy_add approach, then this sequence match will also work for them as well. The actual comm node is dynamic in the sequence, and setup to support all the ones that comm_tensor supports.

spmd/compiler/fusion.py
lessw2020: yes, at this point this only handles the dummy_add expansion, but is setup to be dynamic for any supported collective that uses the dummy_add insertion technique.

PR #612, opened 2022-11-15
[TensorParallel] Add ParallelStyle base class and input preparation APIs

XilunWu:
Goal:

- This PR serves as a starting point of adding Junjie's TensorParallelism low-level and high-level APIs for enabling composable 2D parallelism with DTensor.

Items:

- added ParallelStyle base class
- added input preparation APIs `shard_input_1d` and `replicate_input_1d`.
- added unit tests for `shard_input_1d` and `replicate_input_1d`.


spmd/tensor/parallel/api.py
```diff
@@ -14,6 +14,176 @@
...
+class ParallelStyle(object):
+    """
+    The parallel style user wants the module or submodule to be sharded.
+    Users can extend this class to build their own parallel style with customized input/output preparations.
+    """
+
+    _prepare_input: Callable[
+        [Union[torch.Tensor, DTensor], Optional[DeviceMesh]], DTensor
+    ]
+    _prepare_output: Callable[[DTensor], Union[Tensor, DTensor]]
```
fduwjj, line 92: For output, we might also need a DeviceMesh, let's make it Optional. Also, since we need to pass in the dimension, can we also updated the signature here? I have updated the design already.

spmd/tensor/parallel/api.py
```diff
@@ -14,6 +14,176 @@
...
+    return replicate_output(tensor)
+
+
+def MakeOutputReplicated() -> functools.partial[torch.Tensor]:
+    # This function generates input handler for 1-D mesh device only
+    return functools.partial(_replicate_output_1d)
+
+
+@dataclass
+class ParallelStyle(object):
```
fduwjj, line 83: I feel we can move this into another file named parallel_style.py? WDY?

spmd/tensor/parallel/api.py
```diff
@@ -14,6 +14,176 @@
 from spmd.tensor.parallel import TensorParallelMultiheadAttention
 
 
+def _shard_input_1d(
```
fduwjj, line 17: We need to add comments for this API?

spmd/tensor/parallel/api.py
```diff
@@ -14,6 +14,176 @@
...
+
+    def __init__(self):
+        super().__init__(MakeInputReplicated(), MakeOutputReplicated())
+
+
+def _parallelize_linear_1d(
+    module: nn.Module,
+    parallel_style: ParallelStyle = ColwiseParallel(),
+    device_mesh: DeviceMesh = None,
+    tp_mesh_dim: int = 0,  # TODO: remove this param???
```
fduwjj, line 127: This is still needed for 2D. We should still keep it.

fduwjj (commented):
Please add test for each API we added here. Thanks!

spmd/tensor/parallel/api.py
XilunWu: it makes sense. Also put `MakeInputShard` etc. in parallel_style.py` maybe?

spmd/tensor/parallel/api.py
```diff
@@ -14,6 +14,176 @@
...
+) -> None:
+    if not isinstance(nn.Linear, module):
```
wanchaol, line 129: This is wrong, can you check how `isinstance` should be used?

spmd/tensor/parallel/api.py
```diff
@@ -14,6 +14,176 @@
...
+        raise RuntimeError(
+            f"{module} is not a torch.nn.Linear module but _parallelize_linear was called!"
```
wanchaol, line 131: we shouldn't disclose the function information in error message as we don't expose it to the user.  it should just be expecting torch.nn.Linear but found ...

spmd/tensor/parallel/api.py
```diff
@@ -14,6 +14,176 @@
...
+        )
+
+    if not isinstance(ParallelStyle, parallel_style):
+        raise RuntimeError(
+            f"parallel_style passed to _parallelize_linear is not a ParallelStyle object but {parallel_style}!"
```
wanchaol, line 136: same comment as above.

spmd/tensor/parallel/api.py
```diff
@@ -14,6 +14,176 @@
...
+        )
+
+    col_wise_sharding: Sequence[Placement] = [Shard(0)]
+    row_wise_sharding: Sequence[Placement] = [Shard(1)]
+    replicate: Sequence[Placement] = [Replicate()]
+
+    def linear_module_placements(
+        parallel_style: ParallelStyle,
+    ) -> Tuple[Sequence[Placement], Sequence[Placement]]:
+        if isinstance(RowwiseParallel, parallel_style):
```
wanchaol, line 146: all of `isinstance` check are wrong. Could you please add some tests for this API?

wanchaol (changes requested):
Please add some tests to ensure the API works

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+    assert device_mesh.ndim == 1, "replicate_input_1d: device mesh is not 1D"
+
+    replicate = [Replicate()]
+    if isinstance(tensor, DTensor):
+        return tensor.redistribute(device_mesh, replicate)
+    elif isinstance(tensor, torch.Tensor):
+        return DTensor.from_local(tensor, device_mesh, replicate)
+    else:
+        raise RuntimeError(
+            "Input to replicate_input_1d must be either torch.Tensor or DTensor!"
```
fduwjj, line 81: For better debuggabilty, can we say something like "Input to tensor parallel module is {} not torch.Tensor nor DTensor!".format(type(input)) What do you think?

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+    elif isinstance(tensor, torch.Tensor):
+        return DTensor.from_local(tensor, device_mesh, shard_spec)
+    else:
+        raise RuntimeError(
+            "Input to shard_input_1d must be either torch.Tensor or DTensor!"
+        )
+
+
+def replicate_input_1d(
+    tensor: Union[torch.Tensor, DTensor],
```
fduwjj, line 61: Can we call it input? Up to you.

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+) -> DTensor:
+    # This function replicates input tensor over an 1-D device_mesh
+    if device_mesh is None:
+        if isinstance(tensor, DTensor):
+            device_mesh = tensor.device_mesh
+
+    assert (
+        device_mesh is not None
+    ), "device_mesh is not passed nor can be inferred in replicate_input_1d"
+    assert device_mesh.ndim == 1, "replicate_input_1d: device mesh is not 1D"
```
fduwjj, line 72: ditto.

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+from spmd.tensor import (
+    DTensor,
+    Shard,
+    Replicate,
+    DeviceMesh,
+)
+
+
+@dataclass
+class ParallelStyle(object):
```
fduwjj, line 14: Ideally you can split this one (abstract class) into a separate PR. But this is fine this time. 

Also can we extend ABC instead? (https://docs.python.org/3/library/abc.html)

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    The parallel style user wants the module or submodule to be sharded.
```
fduwjj, line 16: Nit: to be parallelized.

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+    assert device_mesh.ndim == 1, "shard_input_1d: device mesh is not 1D"
+
+    shard_spec = [Shard(dim)]
+    if isinstance(tensor, DTensor):
+        return tensor.redistribute(device_mesh, shard_spec)
+    elif isinstance(tensor, torch.Tensor):
+        return DTensor.from_local(tensor, device_mesh, shard_spec)
+    else:
+        raise RuntimeError(
+            "Input to shard_input_1d must be either torch.Tensor or DTensor!"
```
fduwjj, line 56: See my comment in L81.

test/spmd/tensor/parallel/test_tp_examples.py
```diff
@@ -120,6 +121,56 @@ def forward(self, query, key, value):
 
 
 class DistTensorParallelExampleTest(DistTensorTestBase):
+    @with_comms
+    def test_replicate_input_1d(self):
```
fduwjj, line 125: Nit: you can separate error case and normal cases into two tests. But it's up to you.

test/spmd/tensor/parallel/test_tp_examples.py
```diff
@@ -120,6 +121,56 @@ def forward(self, query, key, value):
...
+        device_mesh = DeviceMesh(self.device_type, [[0, 1], [2, 3]])
+        with self.assertRaisesRegex(
+            AssertionError, "shard_input_1d: device mesh is not 1D"
+        ):
+            dtensor = tp.style.shard_input_1d(tensor, device_mesh)
+
+        device_mesh = DeviceMesh(self.device_type, list(range(NUM_DEVICES)))
+        # test 1
+        dtensor = tp.style.shard_input_1d(tensor, device_mesh)
+        self.assertEqual(tensor, dtensor.to_local())
```
fduwjj, line 166: Are you sure this condition will succeed? The size of tensor is [8, 16] while local tensor size has size of [2, 16]. Or am I missing something?

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+            [DTensor, Optional[DeviceMesh], Optional[int]],
+            Union[torch.Tensor, DTensor],
+        ]
+    ]
+
+
+def shard_input_1d(
+    tensor: Union[torch.Tensor, DTensor],
+    device_mesh: Optional[DeviceMesh] = None,
+    dim: int = 0,
```
fduwjj, line 37: Now I see the problem. Making dim parameter Optional does not make the param len variable. Now I see the disadvantage of this design. Let's remove Optional[int] and pass it in ParalellStyle. What do you think?

fduwjj (commented):
Thanks for working on this and leave some comments. LMK if you want to chat more on this. Thanks!

fduwjj:
Also please rebase. Thanks!

spmd/tensor/parallel/__init__.py
```diff
@@ -8,3 +8,5 @@
     replicate_input,
     replicate_output,
 )
+
+import spmd.tensor.parallel.style
```
fduwjj, line 12: Let's use from ... import ... so that we only expose certain APIs. Thansk!

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+    ]
+    _prepare_output: Optional[
+        Callable[
+            [DTensor, Optional[DeviceMesh], Optional[int]],
+            Union[torch.Tensor, DTensor],
+        ]
+    ]
+
+
+def shard_input_1d(
```
fduwjj, line 34: Also per PyTorch coding standard, you need to add comment to this function with the spec of input/output.

fduwjj: One example for this: https://github.com/pytorch/pytorch/blob/e2f0648750f2d0d0ac648728ce4c514db178cfa1/torch/distributed/_shard/sharded_tensor/api.py#L785.

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+    assert (
+        device_mesh is not None
+    ), "device_mesh is not passed nor can be inferred in shard_input_1d"
+    assert device_mesh.ndim == 1, "shard_input_1d: device mesh is not 1D"
+
+    shard_spec = [Shard(dim)]
+    if isinstance(tensor, DTensor):
+        return tensor.redistribute(device_mesh, shard_spec)
+    elif isinstance(tensor, torch.Tensor):
+        return DTensor.from_local(tensor, device_mesh, shard_spec)
```
fduwjj, line 53: Please use run_check=False for now. Thanks!

test/spmd/tensor/parallel/test_tp_examples.py
fduwjj, line 125: Also, can we move the test to a separate file? like test_tp_style.py?

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+    else:
+        raise RuntimeError(
+            "Input to shard_input_1d must be either torch.Tensor or DTensor!"
+        )
+
+
+def replicate_input_1d(
```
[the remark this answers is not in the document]
fduwjj: After a second thought, I think we can keep it for now since we now only use 1D DeviceMesh for TP. Our design is generic enough for future extension.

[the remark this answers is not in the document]
XilunWu: Agree. When there're use cases of DeviceMesh more than 1D, we can add corresponding shard/replicate functions.

spmd/tensor/parallel/style.py
XilunWu: TIL Python's Abstract Base Classes.

spmd/tensor/parallel/style.py
XilunWu: Is it only for saving comms?

spmd/tensor/parallel/style.py
fduwjj, line 34: Also can we make the name: make_input_shard_1d? The reason is that, for _prepare_output, there is no verb for converting to tensor. Let's all use the verb make to unify the component. Thanks!

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+@dataclass
+class ParallelStyle(object):
+    """
+    The parallel style user wants the module or submodule to be sharded.
+    Users can extend this class to build their own parallel style with customized input/output preparations.
+    """
+
+    _prepare_input: Optional[
+        Callable[
+            [Union[torch.Tensor, DTensor], Optional[DeviceMesh], Optional[int]],
```
fduwjj, line 22: Upon checking, this is fine.

Optional[X] is equivalent to X | None (or Union[X, None]). So let's keep Optional[int].

https://docs.python.org/3/library/typing.html

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,82 @@
...
+    tensor: Union[torch.Tensor, DTensor],
+    device_mesh: Optional[DeviceMesh] = None,
+    dim: int = 0,
+) -> DTensor:
+    # This function shards input tensor on `dim` over an 1-D device_mesh
+    if device_mesh is None:
+        if isinstance(tensor, DTensor):
+            device_mesh = tensor.device_mesh
+
+    assert (
```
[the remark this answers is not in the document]
XilunWu: The reason why I put the assertion below is to make sure that `dtensor.device_mesh` won't be None. But this seems not likely I guess. Then I can simply raise RuntimeError above.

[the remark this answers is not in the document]
wanchaol: DTensor can't construct without a device_mesh, so we don't need to worry about this :)

fduwjj:
Let's not merge for now and please first focus on your own part. https://github.com/pytorch/tau/pull/615 is not overlapping with your change and there should not be too many merge conflict. Thanks!

test/spmd/tensor/parallel/test_tp_style.py
```diff
@@ -84,3 +147,7 @@ def test_prepare_output_error(self):
         self._test_prepare_output_error(make_output_shard_1d)
         self._test_prepare_output_error(make_output_replicate_1d)
         self._test_prepare_output_error(make_output_tensor)
+
+
+if __name__ == "__main__":
```
fduwjj, line 152: Thanks for adding this.

spmd/tensor/parallel/style.py
```diff
@@ -1,10 +1,127 @@
...
 
 
+@dataclass
+class ParallelStyle(ABC):
+    """
+    The parallel style user wants the module or submodule to be parallelized.
+    Users can extend this class to build their own parallel style with customized input/output preparations.
+    """
+
+    _prepare_input: Optional[
```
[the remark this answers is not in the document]
XilunWu: Since `ParallelStyle` is a public API, I don't think we should have type alias (at least a private type alias) in it. What do you think?

fduwjj (approved):
To unblock, but keep the records of three things:
1. Type of Alias.
2. Make abstraction logic into a decorator.
3. Add more test cases if you think are necessary.

For every item if you don't think you can complete in this PR, please create a task and put your name here. Let's document everything. Thanks!

spmd/tensor/parallel/style.py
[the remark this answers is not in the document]
fduwjj: The API itself is a private API and why do we think type alias is not working here? If you want the func signature to be user facing, you can make the type public right?

PR #615, opened 2022-11-16
[PT-D] Add Tensor Parallel _prepare_output component functions

fduwjj:
This is part of efforts trying to standardize and make Tensor Parallel(TP) API Beta Release. 

Since we will use DTensor for all TP internal interactions, we need to convert the output to the state(DTensor / Tensor) which user wants.

### DIFF
--- a/spmd/tensor/parallel/style.py
+++ b/spmd/tensor/parallel/style.py
@@ -1,11 +1,11 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
 import torch
-from typing import Union, Optional
+from typing import Optional
 from spmd.tensor import DTensor, Shard, Replicate, DeviceMesh
 from spmd.tensor.parallel.utils import _prepare_output_validate
 
 
-@_prepare_output_validate  # type: ignore[arg-type] # pyre-ignore[1]
+@_prepare_output_validate  # type: ignore[arg-type] # pyre-ignore[56]
 def make_output_shard_1d(
     output: DTensor, device_mesh: Optional[DeviceMesh] = None, dim: int = 0
 ) -> DTensor:
@@ -25,7 +25,7 @@ def make_output_shard_1d(
     return output.redistribute(device_mesh, [Shard(dim)])
 
 
-@_prepare_output_validate  # type: ignore[arg-type] # pyre-ignore[1]
+@_prepare_output_validate  # type: ignore[arg-type] # pyre-ignore[56]
 def make_output_replicate_1d(
     output: DTensor, device_mesh: Optional[DeviceMesh] = None,
 ) -> DTensor:
@@ -44,7 +44,7 @@ def make_output_replicate_1d(
     return output.redistribute(device_mesh, [Replicate()])
 
 
-@_prepare_output_validate  # type: ignore[arg-type] # pyre-ignore[1]
+@_prepare_output_validate  # type: ignore[arg-type] # pyre-ignore[56]
 def make_output_tensor(
     output: DTensor, device_mesh: Optional[DeviceMesh] = None
 ) -> torch.Tensor:
--- a/spmd/tensor/parallel/style.py
+++ b/spmd/tensor/parallel/style.py
@@ -1,0 +1,64 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates
+import torch
+from typing import Union, Optional
+from spmd.tensor import DTensor, Shard, Replicate, DeviceMesh
+from spmd.tensor.parallel.utils import _prepare_output_validate
+
+
+@_prepare_output_validate
+def make_output_shard_1d(
+    output: DTensor, device_mesh: Optional[DeviceMesh] = None, dim: int = 0
+) -> DTensor:
+    """
+    Convert Output DTensor to a sharded DTensor. This will be used in ParallelStyle.
+    Args:
+        output (DTensor): output of module to be converted.
+        device_mesh (Optional[DeviceMesh]): DeviceMesh to shard the output DTensor. 
+            This needs to be a 1D DeviceMesh and we will throw exceptions if a 
+            non-1D DeviceMesh is passed in. If no DeviceMesh is passed in, we will 
+            reuse the one from output DTensor.
+        dim (int): Sharding dim for output DTensor.
+    Return:
+        (DTensor): A DTensor sharded on the given dim.
+    """
+
+    return output.redistribute(device_mesh, [Shard(dim)])
+
+
+@_prepare_output_validate
+def make_output_replicate_1d(
+    output: Union[torch.Tensor, DTensor],
+    device_mesh: Optional[DeviceMesh] = None,
+) -> DTensor:
+    """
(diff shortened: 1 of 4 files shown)


spmd/tensor/parallel/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        device_mesh = args[1]
+        assert isinstance(
+            output, DTensor
+        ), f"output of Tensor Parallel is actually {type(output)} not DTensor."
+        if device_mesh is None:
+            device_mesh = output.device_mesh
+
+        assert (
+            device_mesh.ndim == 1
+        ), f"{_prepare_output_func.__name__}: device mesh is not 1D"
```
XilunWu, line 48: I'm thinking that the function name part "{_prepare_output_func.__name__}" is not necessary because it will also be displayed in the error trace. What I have now is:

f"device_mesh dim is {device_mesh.ndim} but expcted to be 1"

What do you think?

spmd/tensor/parallel/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    Args:
+        _prepare_output_func (Callable): The func we want to inject the 
+            validation into.
+    Return:
+        func (Callable): Same input func with validation logic added.
+    """
+
+    @functools.wraps(_prepare_output_func)
+    def wrapper(*args, **kwargs):  # pyre-ignore[2, 3]
+        assert len(args) >= 2, "_prepare_output need at least two args."
```
XilunWu, line 37: Question:
If the call is `make_output_replicate(tensor)` (i.e. device_mesh argument passed is None), will `len(args)` be just 1?

XilunWu (approved):
Thx for working on this! I'll adjust my PR to keep a uniform style. Left some comments.

spmd/tensor/parallel/utils.py
fduwjj: yep.

spmd/tensor/parallel/utils.py
fduwjj: OK, maybe we still want to say it's input or output. Let me remove the func name. Thanks!

spmd/tensor/parallel/style.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Convert Output DTensor to a replicated DTensor. This will be used in ParallelStyle.
+    Args:
+        output (DTensor): output of module to be converted.
+        device_mesh (Optional[DeviceMesh]): DeviceMesh to replicate the output DTensor.
+            This needs to be a 1D DeviceMesh and we will throw exceptions if a non-1D
+            DeviceMesh is passed in. If no DeviceMesh is passed in, we will reuse the
+            one from output DTensor.
+    Return:
+        (DTensor): A DTensor made repliacted.
```
wanchaol, line 41: typo: replicate

spmd/tensor/parallel/utils.py
```diff
@@ -0,0 +1,51 @@
+import functools
+
+import torch
+from spmd.tensor import DeviceMesh, DTensor
+from typing import Callable, Optional, Union
+
+
+def _prepare_output_validate(
+    _prepare_output_func: Callable[
+        [DTensor, Optional[DeviceMesh], Optional[int]],
```
wanchaol, line 10: can we make this long callable type a type alias?

spmd/tensor/parallel/utils.py
wanchaol: if we need at least two args, let's just make device mesh be non-optional?

spmd/tensor/parallel/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    @functools.wraps(_prepare_output_func)
+    def wrapper(*args, **kwargs):  # pyre-ignore[2, 3]
+        assert len(args) >= 2, "_prepare_output need at least two args."
+        output = args[0]
+        device_mesh = args[1]
+        assert isinstance(
+            output, DTensor
+        ), f"output of Tensor Parallel is actually {type(output)} not DTensor."
```
wanchaol, line 42: nit: expect output of Tensor Parallel be a DTensor, but found ...

wanchaol (commented):
looks good overall, some nits

spmd/tensor/parallel/utils.py
fduwjj: I see. Sure.

spmd/tensor/parallel/style.py
fduwjj: opps.

spmd/tensor/parallel/utils.py
fduwjj: Opps, I thought it in a wrong way. We do want to make DeviceMesh Optional.. I will change the code.

wanchaol (approved):

spmd/tensor/parallel/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    Args:
+        _prepare_output_func (Callable): The func we want to inject the
+            validation into.
+    Return:
+        func (Callable): Same input func with validation logic added.
+    """
+
+    @functools.wraps(_prepare_output_func)
+    def wrapper(*args, **kwargs):  # pyre-ignore[2, 3]
+        assert len(args) >= 1, "_prepare_output need at least one arg."
```
XilunWu, line 36: nit: need -> needs

spmd/tensor/parallel/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        output = args[0]
+        assert isinstance(
+            output, DTensor
+        ), f"Expect output of Tensor Parallel to be a DTensor, but found {type(output)}."
```
XilunWu, line 40: Sorry for the confusion but.

One question:
Upon the exception being thrown, will the trace display the location of error as in the wrapper or in the actual `_prepare_output_func`? If the former, will it still be better to print `_prepare_output_func.__name__`? Sorry I wasn't thinking about this.

fduwjj: I think with `@functools.wraps`, it should give you the func name. You can try and verify that.

spmd/tensor/parallel/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        if len(args) < 2:
+            device_mesh = output.device_mesh
+        else:
+            device_mesh = args[1]
+
+        assert (
+            device_mesh.ndim == 1
+        ), f"device_mesh has dims {device_mesh.ndim} but expcted to be 1 for output."
+        return _prepare_output_func(*args, **kwargs)
```
XilunWu, line 49: Just found a potential issue: 
In case where `len(args) < 2` (i.e. device_mesh is not passed to func), we need to append it back to `args` before calling `_prepare_output_func(*args, **kwargs)`. Our existing test case did not capture it.

fduwjj: Fixed in https://github.com/pytorch/tau/pull/619.

PR #617, opened 2022-11-16
update dt_planner to use local_offsets from DTensorSpec

wz337:
Update dt_planner to use local_offsets() from DTensorSpec.

Test:
```
python3 test/spmd/checkpoint/test_dt_planner.py
```


wanchaol (approved):
lgtm

PR #619, opened 2022-11-17
[PT-D][Tensor Parallel] Fix the corner cased in _prepare_output and add more test cases + polish comments

fduwjj:
As title.


spmd/tensor/parallel/style.py
```diff
@@ -39,19 +39,19 @@ def make_input_shard_1d(
 
     Args:
         input (Union[Tensor, DTensor]):
-            This single tensor will be sharded on dimension `dim`
-            over the 1-D :class:`DeviceMesh`.
+            This single tensor will be sharded on dimension ``dim``
+            over the 1-D :class:``DeviceMesh``.
```
XilunWu, line 43: n00b question: is there any tool to automatic format the docstring?

XilunWu (approved):
LGTM! Thx for the quick fix and polishing!

spmd/tensor/parallel/utils.py
```diff
@@ -43,17 +43,15 @@ def _prepare_input_validate(
     @functools.wraps(_prepare_input_func)
     def wrapper(*args, **kwargs):  # pyre-ignore[2, 3]
         assert len(args) >= 1, "_prepare_input needs at least one arg."
-        # args is an immutable tuple but we may need to modify it
-        args = list(args)
+        device_mesh = None
```
XilunWu, line 46: How about `device_mesh = None if len(args) < 2 else args[1]`?

spmd/tensor/parallel/style.py
fduwjj: We need to unify how we make the comment. And I think in PyTorch they do have a tool to track that. Good question, we can follow up with Dev Infra team on this.

spmd/tensor/parallel/utils.py
fduwjj: If still have lint errors I will change it, lol.

fduwjj: Changed. Thanks!

spmd/tensor/parallel/style.py
fduwjj: here is the ref for docstring used in PyTorch and looks like there is no tool for that: https://github.com/pytorch/pytorch/blob/master/CONTRIBUTING.md#docstring-type-formatting. You can build one if you want and that will be very helpful to all PyTorch engineers.

PR #620, opened 2022-11-17
[Checkpoint] Update import for spmd checkpoint 

wz337:
Update import for spmd checkpoint files, as we have moved distributed checkpointing from torch.distributed._shard.checkpoint to torch.distributed.checkpoint in PyTorch (https://github.com/pytorch/pytorch/pull/88698).

Test:
CI


wanchaol (approved):
gogogo

PR #622, opened 2022-11-18
[Tensor Parallel] Add more ParallelStyle subclass to unblock low-level module parallelize APIs

XilunWu:
Goal:

- To unblock adding low-level module parallelize APIs `_parallelize_linear` and `_parallelize_mlp`.

What's in this PR:

- added row-wise and column-wise module parallelization styles.
- added unit tests to check initialization of new styles and the correctness of their input/output preparation functions.
- increased code reuse of unit tests.

Next Step:

- `_parallelize_linear`

PR #623, opened 2022-11-18
[PT-D][Tensor Parallel] Implement Pairwise Parallel and _parallelize_mlp

fduwjj:
As title

### DIFF
--- a/spmd/tensor/parallel/api.py
+++ b/spmd/tensor/parallel/api.py
@@ -1,5 +1,4 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
-import functools
 import torch
 import torch.nn as nn
 from typing import Sequence, Tuple
@@ -112,40 +111,6 @@ def _has_even_num_linears(module: nn.Module) -> bool:
     return len(linear_submodules) > 0 and len(linear_submodules) % 2 == 0
 
 
-def _distribute_linear_module(
-    module: nn.Linear,
-    device_mesh: DeviceMesh,
-    weight_distribute_spec: Sequence[Placement],
-    bias_distribute_spec: Sequence[Placement],
-) -> None:
-    """
-    This util function parallelize the weight and bias of a nn.Linear
-    and replace parameter in place.
-
-    Args:
-        module (nn.Linear):
-            :class:``nn.Linear`` object to be parallelized.
-        device_mesh (DeviceMesh):
-            :class:``DeviceMesh`` object which contains
-            how we distribute tensor across GPUs.
-        weight_distribute_spec (Sequence[Placement]):
-            the tensor distribute spec of DTensor for weight.
-        bias_distribute_spec (Sequence[Placement]):
-            the tensor distribute spec of DTensor for bias.
-
-    Return:
-        None
-    """
-    weight = nn.Parameter(
-        distribute_tensor(module.weight, device_mesh, weight_distribute_spec)
-    )
-    bias = nn.Parameter(
-        distribute_tensor(module.bias, device_mesh, bias_distribute_spec)
-    )
-    module.register_parameter("weight", weight)
-    module.register_parameter("bias", bias)
-
-
 def _parallelize_mlp(
     module: nn.Module,
     device_mesh: DeviceMesh,
@@ -162,8 +127,8 @@ def _parallelize_mlp(
         module (nn.Module):
             :class:``nn.Module`` object to be parallelized.
         device_mesh (DeviceMesh):
-            :class:``DeviceMesh`` object which contains
-            how we distribute tensor across GPUs.
+            :class:``DeviceMesh`` object which describes the mesh topology
+            of devices for the DTensor.
         parallel_style (ParallelStyle):
             :class:``ParallelStyle`` object which contains how
             we prepare input/output for Tensor Parallelism.
@@ -181,8 +146,9 @@ def _parallelize_mlp(
     # Define partition functions needed.
     def _rowwise_parallelize_fn(name, module, device_mesh):  # pyre-ignore[2, 3]
         for name, param in module.named_parameters():
-            print(name)
-            dist_spec = [Shard(1)] if name == "weight" else [Replicate()]
+            dist_spec = (  # type: ignore[list-item]
+                [Shard(1)] if name == "weight" else [Replicate()]
+            )
             dist_param = torch.nn.Parameter(
                 distribute_tensor(param, device_mesh, dist_spec)
             )
@@ -216,7 +182,9 @@ def _parallelize_mlp(
                 m,
                 device_mesh,
                 _colwise_parallelize_fn,
-                input_fn=parallel_style._prepare_input if i == 0 else None,
+                input_fn=parallel_style._prepare_input  # pyre-ignore[6]  # type: ignore[arg-type]
+                if i == 0
+                else None,
             )
         else:
             # Row-wise Parallelize the linear layer
@@ -224,7 +192,7 @@ def _parallelize_mlp(
                 m,
                 device_mesh,
                 _rowwise_parallelize_fn,
-                output_fn=parallel_style._prepare_output
+                output_fn=parallel_style._prepare_output  # pyre-ignore[6]  # type: ignore[arg-type]
                 if i == (len(linear_submodules) - 1)
                 else None,
             )
--- a/spmd/tensor/parallel/api.py
+++ b/spmd/tensor/parallel/api.py
@@ -190,7 +190,7 @@ def _parallelize_mlp(
         )
 
     if not _has_even_num_linears(module):
-        raise RuntimeError("We only support even number of Linear for MLP")
+        raise RuntimeError("We only support even number of Linear for MLP.")
 
     if device_mesh.ndim > 1:
         device_mesh = _create_1d_device_mesh(device_mesh, tp_mesh_dim)
--- a/spmd/tensor/parallel/api.py
+++ b/spmd/tensor/parallel/api.py
@@ -116,7 +116,7 @@ def _distribute_linear_module(
     device_mesh: DeviceMesh,
     weight_distribute_spec: Sequence[Placement],
     bias_distribute_spec: Sequence[Placement],
-) -> nn.Module:
+) -> None:
     """
     This util function parallelize the weight and bias of a nn.Linear
     and replace parameter in place.
@@ -133,8 +133,7 @@ def _distribute_linear_module(
             the tensor distribute spec of DTensor for bias.
 
     Return:
-        A :class:``nn.Linear`` object with both weight and bias
-            parallelized based on the given distribute spec.
+        None
     """
     weight = nn.Parameter(
         distribute_tensor(module.weight, device_mesh, weight_distribute_spec)
@@ -151,7 +150,7 @@ def _parallelize_mlp(
     device_mesh: DeviceMesh,
     parallel_style: ParallelStyle = PairwiseParallel(),
     tp_mesh_dim: int = 0,
-) -> None:  # pyre-ignore[7]
+) -> None:
     """
     This function assumes the input module is a sequence of nn.Linear
     and we parallelize the module based on the given parallel style.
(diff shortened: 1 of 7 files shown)


spmd/tensor/parallel/style.py
```diff
@@ -27,6 +21,29 @@ class ParallelStyle(ABC):
     _prepare_input: Optional[_Prepare_Input_Func_Type]
```
wanchaol, line 21: should the prepare_input/output be non-optional type? I think every parallel style should have them specified? This also matches with the `__init__` function where every style must pass those two function?

spmd/tensor/parallel/utils.py
```diff
@@ -104,3 +103,42 @@ def wrapper(*args, **kwargs):  # pyre-ignore[2, 3]
...
+def _create_1d_device_mesh(
+    device_mesh: DeviceMesh, tp_mesh_dim: int = 0
+) -> DeviceMesh:
+    """
+    This function converts a N-D ``device_mesh`` into a 1D ``device_mesh``
+    for 1D Tensor Parallelism.
+
+    Args:
+        device_mesh (DeviceMesh): :class:``DeviceMesh`` object which contains
+            how we distribute tensor across GPUs.
```
wanchaol, line 117: nit: device_mesh which describes the mesh topology of devices for the DTensor

spmd/tensor/parallel/utils.py
```diff
@@ -104,3 +103,42 @@ def wrapper(*args, **kwargs):  # pyre-ignore[2, 3]
...
+        tp_mesh_dim (int): the dimension of ``device_mesh`` where we perform
+            Tensor Parallelism on.
+
+    Return:
+        device_mesh (DeviceMesh): :class:``DeviceMesh`` object which contains
```
wanchaol, line 122: sth like: a 1-D :class:``DeviceMesh`` that Tensor Parallelism works on

spmd/tensor/parallel/api.py
```diff
@@ -86,3 +87,144 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+    ):
+        if i % 2 == 0:
+            # Col-wise Parallelize the linear layer
+            _distribute_linear_module(
+                m,  # pyre-ignore[6]  # type: ignore[arg-type]
+                device_mesh,
+                [Shard(0)],
+                [Shard(0)],
+            )
+            m.register_forward_pre_hook(
```
wanchaol, line 209: shouldn't we only register pre_hook on the first layer, and register post hook on the last layer? If `_parallel_mlp` works on 8 linear layers, we should only register hook for module 0 and module 7?

Also, I'm wondering if we could directly use `distribute_module` on MLPModule, as that'll only register prehook and posthook once, and maybe it also makes easier for pre/post hook?

spmd/tensor/parallel/api.py
```diff
@@ -86,3 +87,144 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+        )
+
+    if not _has_even_num_linears(module):
+        raise RuntimeError("We only support even number of Linear for MLP.")
+
+    if device_mesh.ndim > 1:
+        device_mesh = _create_1d_device_mesh(device_mesh, tp_mesh_dim)
+
+    for i, m in enumerate(
+        filter(lambda x: isinstance(x, nn.Linear), module.children())
```
wanchaol, line 199: I'm wondering if we should ensure sth weird not happen: i.e. `linear0 -> emb -> linear1` will be treated as a legit one but it's actually not, maybe we should check the layers in the middle and ensure they don't have parameters?

spmd/tensor/parallel/style.py
```diff
@@ -27,6 +21,29 @@ class ParallelStyle(ABC):
...
+        self._prepare_input = _prepare_input
+        self._prepare_output = _prepare_output
+
+
+class PairwiseParallel(ParallelStyle):
+    """
+    We concatenate colwise and rowwise styles as a fixed pair like
+    what Megatron-LM(https://arxiv.org/abs/1909.08053) is doing.
+    We assume both input and output needs to a replicate DTensor.
+    We now only support this style for Multihead Attention, MLP and
```
wanchaol, line 35: this line is kinda duplicate with the warning? Maybe we should only keep one.

spmd/tensor/parallel/style.py
```diff
@@ -27,6 +21,29 @@ class ParallelStyle(ABC):
...
+    transformer model.
+
+    .. warning::
+        We need to assume the module to be distributed as a
```
wanchaol, line 39: Maybe it's not something we "assume", it's something this ParallelStyle requires?

spmd/tensor/parallel/style.py
```diff
@@ -27,6 +21,29 @@ class ParallelStyle(ABC):
...
 
+    @abstractmethod
+    def __init__(self, _prepare_input, _prepare_output) -> None:
+        self._prepare_input = _prepare_input
+        self._prepare_output = _prepare_output
+
+
+class PairwiseParallel(ParallelStyle):
+    """
+    We concatenate colwise and rowwise styles as a fixed pair like
```
wanchaol, line 32: "PairwiseParallel style concatenates colwise ..."

test/spmd/tensor/parallel/test_parallelize_api.py
```diff
@@ -0,0 +1,121 @@
...
+            .reshape(
+                self.world_size // (self.world_size // tp_size),
+                self.world_size // tp_size,
+            )
+            .to(torch.int)
+        )
+        mesh = DeviceMesh(self.device_type, mesh_shape)
+        one_dimention_mesh_shape = mesh_shape[self.rank // tp_size, :]
+        pg = mesh.get_dim_groups()[1]
+        new_mesh = _create_1d_device_mesh(mesh, 1)
```
wanchaol, line 43: can we also add a case to test tp_dim=0?

wanchaol (commented):
Awesome work! Mostly some suggestions about comment, and have some questions about parallel_mlp works on some weird module pattern and hook registration, please see comments inlined.

spmd/tensor/parallel/api.py
```diff
@@ -86,3 +87,144 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+            we prepare input/output for Tensor Parallelism.
+        tp_mesh_dim (int):
+            the dimension of ``device_mesh`` where we perform
+            Tensor Parallelism on.
+
+    Return:
+        None
+
+    .. warning::
+        We now only support ``PairwiseParallel`` for now.
```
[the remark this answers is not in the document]
fduwjj: You are right. But we might have other style for MLP parallelization in the future, right?

spmd/tensor/parallel/api.py
```diff
@@ -86,3 +87,144 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+    """
+
+    # Define hook functions needed for preparing Input/Output.
+    def _module_forward_pre_hook(*args):  # pyre-ignore[2, 3]
```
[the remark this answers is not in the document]
fduwjj: yes...

spmd/tensor/parallel/style.py
fduwjj: Actually that's a good point, let me remove it.

spmd/tensor/parallel/api.py
fduwjj: But softmax and layerNorm also have parameters. It's hard to differentiate it from embedding in a generic way.

spmd/tensor/parallel/style.py
fduwjj: Oh sure.

test/spmd/tensor/parallel/test_parallelize_api.py
fduwjj: Sure.

spmd/tensor/parallel/api.py
fduwjj: Hmmm I don't have strong opinion for whether to use `distribute_module` or not. But the thing is that how are we going to write `partition_fn`? Unless we also do this layer by layer.

spmd/tensor/parallel/api.py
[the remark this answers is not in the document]
fduwjj: Use `distribute_module`

spmd/tensor/parallel/style.py
fduwjj: NVM has a solution here.

spmd/tensor/parallel/api.py
fduwjj: NVM, found a solution.

spmd/tensor/parallel/api.py
```diff
@@ -86,3 +87,112 @@ def _shard_self_attn_params(name: str, module: nn.Module) -> None:
...
+    Args:
+        module (nn.Module):
+            :class:``nn.Module`` object to be traversed and counted.
+
+    Return:
+        A boolean object which specifies whether the module contains
+        event-number of Linears in its children.
+
+    .. warning::
+        The traversal is not recursive for now.
```
wanchaol, line 106: qq: why we need this function to be recursive?

spmd/tensor/parallel/style.py
wanchaol: I see that we didn't remove the `Optional`, so I assume there're still cases when a parallel style have `_prepare_input` or `_prepare_output` be None? Could you share more about those cases?

spmd/tensor/parallel/utils.py
```diff
@@ -44,6 +43,9 @@ def _prepare_input_validate(
     def wrapper(*args, **kwargs):  # pyre-ignore[2, 3]
         assert len(args) >= 1, "_prepare_input needs at least one arg."
         input = args[0]
+        if isinstance(input, list) or isinstance(input, tuple):
+            input = input[0]
+            args = (input, *args[1:])
```
wanchaol, line 48: so if the first arg is a list of tuple, we only convert the first item in the list/tuple to DTensor? What if the first arg is sth like `(tensor, tensor, tensor)`?

spmd/tensor/parallel/style.py
```diff
@@ -27,6 +21,27 @@ class ParallelStyle(ABC):
...
+
+
+class PairwiseParallel(ParallelStyle):
+    """
+    PairwiseParallel concatenate colwise and rowwise styles as a fixed
+    pair like what Megatron-LM(https://arxiv.org/abs/1909.08053) is doing.
+    We assume both input and output needs to a replicate DTensor.
+
+    .. warning::
+        We need to the module to be distributed as a
```
wanchaol, line 37: looks like the some typos, i'm wondering if this would make it more clear: "PairwiseParallel only supports ``nn.Multihead Attention``, ``nn.Transformer`` or MLP Layers (i.e. even number ``nn.Linear`` layers) for now"

spmd/tensor/parallel/api.py
fduwjj: Because what if you have a sequence of a sequence of linear that kind of thing?

spmd/tensor/parallel/utils.py
fduwjj: For now we can skip this case right? Or you feel this is important to handle?

spmd/tensor/parallel/style.py
fduwjj: Because if you see ColwiseParallel, the prepare_output is None.

spmd/tensor/parallel/style.py
fduwjj: Changed the comment here. But I would not use the world MLP since L means random layers of perceptrons.

spmd/tensor/parallel/utils.py
fduwjj: Addressed offline.

spmd/tensor/parallel/style.py
fduwjj: Removed.

wanchaol (approved):
lgtm, thanks for addressing the comments!

XilunWu:
landing of #622 causes conflict. sorry for that.

Issue #625, opened 2022-11-18
[SPMD] Clone operation added during DTensor comm collective insertion slows perf


lessw2020:
After expansion of DTensor communication operations, fx is inserting a clone operation to clone the gradient tensor. 
This operation will slow down the perf and add memory, but is technically not needed. 

I have verified that we can rewrite the graph to remove it by directly updating to use the grad tensor. 

Thus we need to either:
a - modify the comm subgraph such that fx does not insert the clone operation or
b - have an optimization pass that will obviate the clone nodes via relinking the comm primitives.

PR #626, opened 2022-11-18
add _remove_gradient_tensor_clones

lessw2020:
This adds an initial optimization pass in fusion.py to remove any duplicate gradient tensor nodes created from the DTensor comm collective insertion.

It is a sequence matcher that locates the DTensor expansion signature, delinks the clone node, and remaps the comm collective tensor argument to the actual gradient tensor node.
After the pass, it then recompiles the graph to remove the duplicate gradient nodes and return an optimized graph with no redundant gradient clone nodes and a count of clone nodes removed.

This PR resolves this tracking issue:
https://github.com/pytorch/tau/issues/625

### DIFF
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -2,7 +2,7 @@ import logging
 from copy import deepcopy
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Iterable, List, Optional
+from typing import Iterable, List, Optional, Union
 
 import torch
 import torch.fx as fx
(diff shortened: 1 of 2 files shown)


lessw2020:
actually I have a new PR coming that will remove the clone node near the source, where we generate the clone node and before it goes into the main graph.  
Thus, will submit that one as well and can decide optimal approach.

spmd/compiler/fusion.py
```diff
@@ -162,22 +165,82 @@ def _scan_graph_for_fusion_elements(
...
+        elif (
+            index == len_sequence
+            and node.op == OP.CALL_FUNCTION
+            and node.name.startswith(pattern)
+        ):
+            # found matching clone/comm sequence...
+            grad_tensor_node = clone_node.args[0]
+            comm_node = node
+
+            comm_node.update_arg(0, [grad_tensor_node])
```
mrshenli, line 221: I'd vote for the more generic solution as in #631, as this fusion seems more closely bound with aotautograd internal behavior, e.g., how it handles reduce ops and process group args. 

Let me add more comments on the other PR.

PR #628, opened 2022-11-19
Fix MyPy error from PT CI

fduwjj:
Got Lint errors from PyTorch CI, so we will use this PR to fix it.

Ref: https://github.com/pytorch/pytorch/actions/runs/3500271587/jobs/5862763116

Also open MyPy check for TAU so that we can fix it early on.


wanchaol (approved):

fduwjj:
Fix tests so that we can pass PyTorch CI and avoid situation when # of GPUs are not 4 or even number.

PR #631, opened 2022-11-19
Feature remove clones source

lessw2020:
This is a second PR for removing the redundant gradient clone tensors, and approaches it differently than the first PR.   
(first one is https://github.com/pytorch/tau/pull/626)

This one effects the change closer to the source, by rewriting the subgraph produced from  _get_dtensor_dispatch_graph and inside dummy_add_graph.

It does this by adding a new function, _remove_clone_tensor() which rewrites the subgraph to bypass the clone node and then removes the clone. 
Thus, the graphs produced from SPMD do not have the redundant clones to start.
The one negative here is that each subgraph rewrite now requires it own recompile to cement the change, vs one final recompile in the earlier PR.  Not sure if recompiling multiple subgraphs is any more or less expensive vs one time full graph recompile. 
By contrast, the earlier PR cleaned the graphs by a direct pass through the entire finished graph. 

So two different approaches to resolving the same issue.
I think resolving closer to source is preferred, hence this second PR.

### DIFF
--- a/spmd/compiler/api.py
+++ b/spmd/compiler/api.py
@@ -185,8 +185,8 @@ def _remove_clone_tensor(subgm: fx.GraphModule) -> fx.GraphModule:
     """rewrite dummy add graph to remove clone of grad tensor"""
     nodemap = create_graph_node_map(subgm)
 
-    clone_node = nodemap["clone"]
-    comm_node = nodemap["allreduce__default"]
+    clone_node = nodemap.get("clone")
+    comm_node = nodemap.get("allreduce__default")
 
     assert comm_node is not None, "expected all_reduce comm node not present."
     assert clone_node is not None, "expected clone node not present."
--- a/spmd/compiler/api.py
+++ b/spmd/compiler/api.py
@@ -185,6 +185,11 @@ def _remove_clone_tensor(subgm: fx.GraphModule) -> fx.GraphModule:
     """rewrite dummy add graph to remove clone of grad tensor"""
     nodemap = create_graph_node_map(subgm)
 
+    _debug(f"188, submap type {type(subgm)}\n ")
+    _debug(f"189, {subgm.graph.print_tabular()} ")
+    _debug(f"189, {print(subgm.graph)}")
+    _debug(f"190, {nodemap=}\n")
+
     clone_node = nodemap.get("clone")
     comm_node = nodemap.get("allreduce__default")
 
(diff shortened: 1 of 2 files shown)


spmd/compiler/api.py
```diff
@@ -171,11 +173,34 @@ def dummy_add(grad: torch.Tensor, zero: torch.Tensor) -> torch.Tensor:
...
+    nodemap = create_graph_node_map(subgm)
+
+    clone_node = nodemap.get("clone")
+    comm_node = nodemap.get("allreduce__default")
+
+    assert comm_node is not None, "expected all_reduce comm node not present."
+    assert clone_node is not None, "expected clone node not present."
+
+    grad_tensor_node = clone_node.args[0]
+    comm_node.update_arg(0, [grad_tensor_node])
```
mrshenli, line 195: this looks very specific. Do we need to verify the following few conditions?

1. `comm_node` indeed takes `clone_node` as argument
2. there is no other places using `clone_node`.

spmd/compiler/api.py
```diff
@@ -171,11 +173,34 @@ def dummy_add(grad: torch.Tensor, zero: torch.Tensor) -> torch.Tensor:
...
 
+    # rewrite graph to remove redundant clone
+    traced_dispatch = _remove_clone_tensor(traced_dispatch)
+
     traced_dispatch.graph.lint()
 
     return traced_dispatch
 
 
+def _remove_clone_tensor(subgm: fx.GraphModule) -> fx.GraphModule:
```
mrshenli, line 184: A few things we can do here to make it more generic:

0. make sure the original node (clone's arg) are not called by any inplace ops after clone and before used in allreduce. Otherwise, linking allreduce to the original node might be wrong. 
1. relax the comm node from allreduce to other types of node as well. The tricky thing here is how to deal with inplace ops. If there are ops using the original node after allreduce (or other inplace ops), removing clone will make the behavior different.  
2. call `eliminate_dead_code()` to automatically remove the clone node.

mrshenli: Not sure if the following is correct:

For allreduce (or other ops) to replace its clone arg with the original arg: there can be any inplace op before allreduce and after clone

For subsequent logic to be correct: there can be any op (inplace or not inplace) after allreduce using the original node. Otherwise, allreduce would have modified value.

anj-s: couple of questions:
1. For the 0th point above, given that we are adding the dummy_add operation is there any danger of other ops in between the clone and allreaduce? If yes, what is an example of that.
2. A noob question: what does this optimization give us in terms of performance or usability?

spmd/compiler/api.py
lessw2020: Good points - I've added a check and two asserts to ensure the items above (comm_node takes clone_node as arg) and no other places are using clone node:
comm_node_user_check

spmd/compiler/api.py
lessw2020: For item 0:
I've added a user check to confirm there is only one use of the grad_tensor in the comm section subgraph. 
grad_tensor_one_user

lessw2020: Related - for @anj-s  item 1 - the check/assert added above confirms that there will only be 1 user in the subgraph, so that should remove this issue of concern re: other users of the clone.

For item 2 - what perf or usability gains?
Main gain is in perf - the clone nodes are duplicating every grad tensor in the backward graph on cpu.  There is a small hit for each cpu alloc perf wise, and esp for for larger models (think 170B) we could end up blowing out or at least consuming a large amount of cpu memory trying to clone every tensor in the graph as it currently does (not sure how fast gc would remove these vs. creation).  
I don't think there is any usability gain other than we have a cleaner graph to work with.  The main gain imo is that we are not spinning up a large part of an entire model clone on the users cpu and assuming that is a blocking mem alloc, should improve perf a bit relative to the size of the clones being created.

lessw2020: related - additionally for perf, the gradients would need to flow from the clone back to the source tensor, so that is additional overhead involved as well as the aforementioned memory alloc time and overhead.  

Final item - I'm unclear why using a cpu based clone does not cause an error since this is using NCCL. But it is another reason we want to remove these imo and use the gpu gradient tensors directly.

PR #634, opened 2022-11-21
[PT-D][Tensor Parallelism] Implement parallelize_module API and refactor _parallelize_multihead_attn

fduwjj:
We are building a new unified high-level API so that user can directly use it for tensor parallelism. Our tensor parallelism is built on top of DTensor (RFC: https://github.com/pytorch/pytorch/issues/88838).

We have already built some APIs for Multihead_attention parallelism and we now just refactor it to make sure it is working with our overall design for tensor parallelism.

### DIFF
--- a/spmd/tensor/parallel/api.py
+++ b/spmd/tensor/parallel/api.py
@@ -1,106 +1,128 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
 import torch
 import torch.nn as nn
-from typing import Sequence, Tuple
+from typing import Union, Dict
 from spmd.tensor import (
     distribute_module,
     distribute_tensor,
-    DTensor,
     Shard,
     Replicate,
     DeviceMesh,
-    Placement,
 )
 from spmd.tensor.parallel import TensorParallelMultiheadAttention
-from spmd.tensor.parallel.style import ParallelStyle, PairwiseParallel
+from spmd.tensor.parallel.style import PairwiseParallel, ParallelStyle
 from spmd.tensor.parallel.utils import _create_1d_device_mesh
 
 
-def replicate_input(
-    inputs: Sequence[torch.Tensor], device_mesh: DeviceMesh
-) -> Tuple[DTensor, ...]:
-    replicate = [Replicate()] * device_mesh.ndim
-    return tuple(
-        DTensor.from_local(tensor, device_mesh, replicate) for tensor in inputs
-    )
+__all__ = [
+    "parallelize_module",
+]
 
 
-def replicate_output(output: DTensor, device_mesh: DeviceMesh) -> torch.Tensor:
-    if isinstance(output, DTensor):
-        replicate = [Replicate()] * output.device_mesh.ndim
-        # TODO: can the output be left incontiguous?
-        return (
-            output.redistribute(output.device_mesh, replicate)
-            .to_local()
-            .contiguous()
+def parallelize_module(
+    module: nn.Module,
+    device_mesh: DeviceMesh,
+    parallelize_plan: Union[ParallelStyle, Dict[str, ParallelStyle]],
+    tp_mesh_dim: int = 0,
+) -> nn.Module:
+    """
+    The major API for tensor parallelism in PyTorch. We parallelize module
+    or sub_modules based on a parallelize_plan which contains the parallel_style
+    which indicates how user want the module or sub_module to be parallelized.
+    User can also specify different parallel_style per FQN. The API supports 2D
+    parallelism natively by accepting an n-dimension device_mesh and users just
+    need to specify the dimension where we perform tensor parallelism on.
+
+    Args:
+        module (nn.Module):
+            :class:``nn.Module`` object to be parallelized.
+        device_mesh (DeviceMesh):
+            :class:``DeviceMesh`` object which describes the mesh topology
+            of devices for the DTensor.
+        parallelize_plan (Union[ParallelStyle, Dict[str, ParallelStyle]]):
+            The plan used to parallelize the module. It can be either a
+            :class:``ParallelStyle`` object which contains how
+            we prepare input/output for Tensor Parallelism or it can be a
+            dict of FQN and its corresponding :class:``ParallelStyle`` object.
+        tp_mesh_dim (int):
+            the dimension of ``device_mesh`` where we perform
+            Tensor Parallelism on.
+
+    Return:
+        A :class:``nn.Module`` object parallelized.
+
+    Example::
+        >>> # xdoctest: +SKIP("distributed")
+        >>> from from spmd.tensor.parallel import parallelize_module, PairwiseParallel
+        >>>
+        >>> # Define the module.
+        >>> m = Model(...)
+        >>> m = parallelize_module(m, PairwiseParallel())
+        >>>
+
+    .. warning::
+        ``PairwiseParallel`` comes with constraints for now. If you need finer
+        granularity 
+    """
+
+    if device_mesh.ndim > 1:
+        device_mesh = _create_1d_device_mesh(device_mesh, tp_mesh_dim)
+
+    if isinstance(parallelize_plan, ParallelStyle):
+        if _is_mha_for_pairwise_parallel(module):
+            return _parallelize_multihead_attn(module, device_mesh)
+        elif _is_mlp_for_pairwise_parallel(module):
+            return _parallelize_mlp(module, device_mesh)
+        else:
+            for n, m in module.named_children():
+                module.register_module(
+                    n, parallelize_module(m, device_mesh, parallelize_plan)
+                )
+            return module
+    elif isinstance(parallelize_plan, dict):
+        for module_path, parallelize_style in parallelize_plan.items():
+            sub_module = module.get_submodule(module_path)
+            module.register_module(
+                parallelize_module(
+                    module_path, sub_module, device_mesh, parallelize_style
+                )
+            )
+            return module
+    # TODO: Add parallelize linear logic when https://github.com/pytorch/tau/pull/624/ merged.
+    else:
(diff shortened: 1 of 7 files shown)


spmd/tensor/parallel/api.py
```diff
@@ -1,122 +1,195 @@
...
+    module: nn.Module,
+    device_mesh: DeviceMesh,
+    parallelize_plan: Union[ParallelStyle, Dict[str, ParallelStyle]],
+    tp_mesh_dim: int = 0,
+) -> nn.Module:
+    """
+    The major API for tensor parallelism in PyTorch. We parallelize module
+    or sub_modules based on a parallelize_plan which contains the parallel_style
+    which indicates how user want the module or sub_module to be parallelized.
+    User can also specify different parallel_style per FQN. The API supports 2D
```
wanchaol, line 32: "per FQN" does this mean module FQN or parameter FQN? I think we might want to say per module, as users might not understand what FQN mean? Or we explicitly say module fully qualifed name.

spmd/tensor/parallel/api.py
```diff
@@ -1,122 +1,195 @@
...
-def replicate_output(output: DTensor, device_mesh: DeviceMesh) -> torch.Tensor:
-    if isinstance(output, DTensor):
-        replicate = [Replicate()] * output.device_mesh.ndim
-        # TODO: can the output be left incontiguous?
-        return (
-            output.redistribute(output.device_mesh, replicate)
-            .to_local()
-            .contiguous()
-        )
+def parallelize_module(
```
wanchaol, line 22: naming: Shall we call it `parallel_module`, or `parallelize_module`? which one do you think it's better?

spmd/tensor/parallel/api.py
```diff
@@ -1,122 +1,195 @@
...
+        granularity 
+    """
+
+    if device_mesh.ndim > 1:
+        device_mesh = _create_1d_device_mesh(device_mesh, tp_mesh_dim)
+
+    if isinstance(parallelize_plan, ParallelStyle):
+        if _is_mha_for_pairwise_parallel(module):
+            return _parallelize_multihead_attn(module, device_mesh)
+        elif _is_mlp_for_pairwise_parallel(module):
```
wanchaol, line 74: does this mean if `parallelize_plan` is ParallelStyle, we only support PairwiseParallel? what if user pass in a single RowwiseParallel or ColwiseParallel? shall we just error out in those case?

spmd/tensor/parallel/api.py
```diff
@@ -1,122 +1,195 @@
...
+    """
+    if isinstance(module, TensorParallelMultiheadAttention) or isinstance(
+        module, nn.MultiheadAttention
+    ):
+        return True
+
+
+def _is_mlp_for_pairwise_parallel(module: nn.Module) -> bool:
+    """
+    Traverse through all the children of the given module and count the
```
wanchaol, line 117: nit: all the immediate children of the given module?

spmd/tensor/parallel/api.py
```diff
@@ -137,51 +210,106 @@ def _parallelize_mlp(
...
+    if isinstance(module, nn.MultiheadAttention):
+        tp_multi_head_attention = TensorParallelMultiheadAttention(
+            module.embed_dim,
+            module.num_heads,
+            device=torch.device(device_mesh.device_type),
+            tp_size=device_mesh.size(tp_mesh_dim),
+            add_bias_kv=module.bias_k is not None,
+        )
+        tp_multi_head_attention.copy(module)
+        return _parallelize_multihead_attn(tp_multi_head_attention, device_mesh)
```
wanchaol, line 236: nit: instead of recursive call, shall we just assign module to `tp_multi_head_attention` instead?

spmd/tensor/parallel/api.py
```diff
@@ -1,122 +1,195 @@
...
+        A boolean object which specifies whether the module is MLP or not.
 
     .. warning::
         The traversal is not recursive for now.
     """
     linear_submodules = list(
         filter(lambda x: isinstance(x, nn.Linear), module.children())
     )
-    return len(linear_submodules) > 0 and len(linear_submodules) % 2 == 0
+    return len(linear_submodules) > 0
```
wanchaol, line 133: oh so we don't check if it's even number of linear layers now? what if there're 3 linear layers, how we gonna shard the third one?

test/spmd/tensor/parallel/test_2d_parallel.py
```diff
@@ -108,10 +101,10 @@ def shard_module(m, pg):
     m.net2 = _aggregate_local_tensor(m.net2)
 
 
-def _shard_wrap_module(module, module_shard, fsdp_wrap, tp_pg, fsdp_pg):
+def _shard_wrap_module(module, module_shard, fsdp_wrap, twod_mesh, fsdp_pg):
```
wanchaol, line 104: nit: 2d_mesh?

spmd/tensor/parallel/api.py
```diff
@@ -1,122 +1,195 @@
...
+        module (nn.Module):
+            :class:``nn.Module`` object to be parallelized.
+        device_mesh (DeviceMesh):
+            :class:``DeviceMesh`` object which describes the mesh topology
+            of devices for the DTensor.
+        parallelize_plan (Union[ParallelStyle, Dict[str, ParallelStyle]]):
+            The plan used to parallelize the module. It can be either a
+            :class:``ParallelStyle`` object which contains how
+            we prepare input/output for Tensor Parallelism or it can be a
+            dict of FQN and its corresponding :class:``ParallelStyle`` object.
```
wanchaol, line 46: we should say module FQN to avoid confusion with parameter FQN

wanchaol (commented):
Looking good, left some comments about questions and naming, let me know if those make sense or not.

spmd/tensor/parallel/api.py
fduwjj: Sure, let me change it to clarify that.

spmd/tensor/parallel/api.py
fduwjj: parallel to me is a noun, so I would still want to keep the verb `parallelize`. Is there any naming guidance for PyTorch?

spmd/tensor/parallel/api.py
fduwjj: Oh that is TBD since Xilun's PR has not been merged yet. Let me add a TODO here to clarify.

spmd/tensor/parallel/api.py
fduwjj: Oh I see your point, ok. Let me change it.

spmd/tensor/parallel/api.py
fduwjj: We don't do anything for the third layer. The tricky thing is that if we only limit to even-number of linears, we cannot use pairwise for 3-layer of Linears and users have to specify parallel_style per path. That's why I want to change it.

spmd/tensor/parallel/api.py
fduwjj: sure.

test/spmd/tensor/parallel/test_2d_parallel.py
fduwjj: Looks this is invalid syntax for python... maybe let's call it mesh_2d.

fduwjj:
The one left mypy error is from PiPPy and @kwen2501 is on it.

spmd/tensor/parallel/api.py
wanchaol: This make sense, let's keep it as `parallelize`, i don't think there's a naming guidance from pytorch, so we need to rationalize API by ourselves. Do you think we shall call `parallel_plan` as `parallel` is a noun/adj?

spmd/tensor/parallel/api.py
wanchaol: I see, so pair-wise only parallel the first two linear layers in this case, and leave the last one untouched, that make sense, I guess for MLP we don't really need to care about the third one

spmd/tensor/parallel/api.py
```diff
@@ -1,122 +1,194 @@
...
-            .contiguous()
-        )
+def parallelize_module(  # type: ignore[return]
+    module: nn.Module,
+    device_mesh: DeviceMesh,
+    parallelize_plan: Union[ParallelStyle, Dict[str, ParallelStyle]],
+    tp_mesh_dim: int = 0,
+) -> nn.Module:
+    """
+    The major API for tensor parallelism in PyTorch. We parallelize module
```
wanchaol, line 29: This is the only API we plan to expose to the user right? Let's just call it "The API to apply Tensor Parallelism"?

spmd/tensor/parallel/api.py
```diff
@@ -1,122 +1,194 @@
...
+        A boolean object which specifies whether the module is MLP or not.
 
     .. warning::
         The traversal is not recursive for now.
     """
     linear_submodules = list(
         filter(lambda x: isinstance(x, nn.Linear), module.children())
     )
-    return len(linear_submodules) > 0 and len(linear_submodules) % 2 == 0
+    return len(linear_submodules) > 0
```
wanchaol, line 132: On a second thought, shouldn't we return when `len(linear_submodules) > 1` instead? It also matches with the error throwing below.

wanchaol (approved):
lgtm, a few more comments.

spmd/tensor/parallel/api.py
fduwjj: Yep, sure, let me change it.

spmd/tensor/parallel/api.py
fduwjj: I don't have a strong opinion on this one. Let me think more on this.

PR #639, opened 2022-11-22
Fix for shape mismatch during SPMD tracing.

aazzolini:
I'll explain tomorrow.

### DIFF
--- a/spmd/compiler/aot_function_patch.py
+++ b/spmd/compiler/aot_function_patch.py
@@ -1,0 +1,144 @@
+from functorch._src.aot_autograd import (
+    default_partition,
+    AOTConfig,
+    PytreeThunk,
+    create_aot_dispatcher_function,
+    KNOWN_TYPES,
+)
+from functools import wraps
+from typing import Callable, Dict, Optional, Tuple
+import torch.utils._pytree as pytree
+
+
+def patched_aot_function(
+    fn: Callable,
+    fw_compiler: Callable,
+    bw_compiler: Optional[Callable] = None,
+    partition_fn: Callable = default_partition,
+    decompositions: Optional[Dict] = None,
+    num_params_buffers: int = 0,
+    hasher_type=None,  # deprecated
+    static_argnums: Optional[Tuple[int]] = None,  # deprecated
+    pre_compile_fn: Optional[Callable] = None,
+) -> Callable:
+    """
+    Traces the forward and backward graph of :attr:`fn` using torch dispatch
+    mechanism, and then compiles the generated forward and backward graphs
+    through :attr:`fw_compiler` and :attr:`bw_compiler`.
+
+    :func:`aot_function` traces the forward and backward graph ahead of time,
+    and generates a joint forward and backward graph.  :attr:`partition_fn` is
+    then used to separate out forward and backward graphs. The partitioner
+    function can be used to perform optimizations such as recomputation. One can
+    set `decompositions` dictionary to decompose the operators into a sequence
+    of core or simpler operators supported by the backend compilers.
+
+    :func:`aot_function` uses a compilation cache, based on input tensor
+    properties, to detect when there is a need of recompilation.
+
+    .. warning::
+        This API is experimental and likely to change.
+
+    Args:
+        fn (Callable): A Python function that takes one ore more arguments. Must
+            return one or more Tensors.
+        fw_compiler (Callable): A Python function that accepts an Fx graph with
+            Aten ops and input args, and returns a Callable that semantically is
+            equivalent to the input Fx graph.
+        bw_compiler (Optional[Callable]): A Python function that accepts an
+            Fx graph with Aten ops and input args, and returns a Callable that
+            semantically is equivalent to the input Fx graph.  Default: None
+            (when None, it defaults to the :attr:`fw_compiler`)
+        partition_fn (Callable): A Python function that takes a joint forward
+            and backward graph, and partitions it into separate forward and
+            backward graphs.
+        decompositions (Dict): A dictionary to define the decomposition of
+            larger Aten ops into simpler or core Aten ops.
+
+    Returns:
+        Returns a ``Callable`` that retains the eager behavior of the original
+        :attr:`fn`, but with forward and backward graph compiled via
+        :attr:`fw_compile` and :attr:`bw_compile`.
+
+    A simple example usage of :func:`aot_function` is as follows. This example
+    will print the forward and backward graphs of the function ``fn``
+
+        >>> fn = lambda x : x.sin().cos()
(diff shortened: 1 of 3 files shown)


spmd/compiler/aot_function_patch.py
```diff
@@ -0,0 +1,144 @@
...
+    PytreeThunk,
+    create_aot_dispatcher_function,
+    KNOWN_TYPES,
+)
+from functools import wraps
+from typing import Callable, Dict, Optional, Tuple
+import torch.utils._pytree as pytree
+
+
+def patched_aot_function(
```
fegin, line 13: Can you add @no_type_check to avoid the breakage of mypy?

wanchaol, line 13: what's this patched aot_function for? can we add a comment about why we need this?

spmd/compiler/api.py
```diff
@@ -358,10 +392,34 @@ def forward(
...
+                return compile_inps
+
             self._compiled_m = aot_module(
                 self._local_module,
                 partial(self._compile, TrainingPhase.FORWARD),
                 partial(self._compile, TrainingPhase.BACKWARD),
+                pre_compile_fn=gather_inputs_for_compilation,
             )
+            global_args = [
+                DTensor.from_local(arg, self._schema.mesh, [Shard(0)])
```
fegin, line 415: Add `type: ignore[arg-type]` to suppress the mypy error.

fegin (commented):
We should fix the test breakage (mypy) before landing this PR.

spmd/compiler/api.py
```diff
@@ -298,9 +317,24 @@ def _convert_to_distributed(
...
+                    schemas[i].placements,
+                )
+            else:
+                # But in the backward pass we got "real" sharded inputs
+                # so we have to actually make DTensors out of them
+                assert training_phase == TrainingPhase.BACKWARD
+                node_to_obj[node] = DTensor.from_local(
+                    inps[i],
+                    schemas[i].mesh,
+                    schemas[i].placements,
```
wanchaol, line 335: nit: we should use `run_check=False` to avoid potential communication?

spmd/compiler/api.py
```diff
@@ -358,10 +392,34 @@ def forward(
...
+                    .to_local()
+                    for x in inps
+                ]
+                return compile_inps
+
             self._compiled_m = aot_module(
                 self._local_module,
                 partial(self._compile, TrainingPhase.FORWARD),
                 partial(self._compile, TrainingPhase.BACKWARD),
+                pre_compile_fn=gather_inputs_for_compilation,
```
wanchaol, line 412: oh so this is where we need to patch the aot_function?

spmd/compiler/api.py
```diff
@@ -358,10 +392,34 @@ def forward(
         self, *args: Tuple[object], **kwargs: Dict[str, object]
     ) -> object:
         if self._compiled_m is None:
+            input_set = set(args)
+
+            def gather_inputs_for_compilation(inps):
+                compile_inps = [
+                    x
+                    if not isinstance(x, torch.Tensor) or x not in input_set
+                    else DTensor.from_local(x, self._schema.mesh, [Shard(0)])
```
wanchaol, line 401: what this is actually doing is basically create a sharded DTensor then convert it to replicate, then to local in order to capture the correct shape?

wanchaol (approved):
stamp to unblock, I think we might need some follow ups on cleaning up this hack a bit.

spmd/compiler/api.py
```diff
@@ -298,9 +317,24 @@ def _convert_to_distributed(
             assert i < len(
                 inps
             ), f"got more placeholer nodes ({i + 1}) than inputs ({len(inps)})"
-            node_to_obj[node] = DTensor.from_local(
-                inps[i], schemas[i].mesh, schemas[i].placements
-            )
+            if training_phase == TrainingPhase.FORWARD:
+                # in the forward phase we start with the full "global" ensors.
```
mrshenli, line 321: nit: enstors -> tensors

spmd/compiler/api.py
```diff
@@ -298,9 +317,24 @@ def _convert_to_distributed(
...
+                # we needed this because we needed to capture the original graph.
+                node_to_obj[node] = distribute_tensor(  # DTensor.from_local(
+                    inps[i],
+                    schemas[i].mesh,
+                    schemas[i].placements,
+                )
+            else:
+                # But in the backward pass we got "real" sharded inputs
+                # so we have to actually make DTensors out of them
+                assert training_phase == TrainingPhase.BACKWARD
```
mrshenli, line 331: I didn't fully get this. I thought for both forward and backward, we first get graphs of local tensors and expand that. Did the plan change?

mrshenli (commented):
Stamp to unblock. It will definitely be helpful to have some more description to the changes in `_convert_to_distributed` function.

mrshenli (approved):

spmd/compiler/aot_function_patch.py
aazzolini: Added a comment .

spmd/compiler/api.py
aazzolini: It's alright as this should happen only once at compile time.

spmd/compiler/api.py
aazzolini: yeah i added a comment explaining -- the issue is that we want to present "full" logical inputs at capture time, but then we want to pass shraded input at run-time (because we transformed the graph in between).

spmd/compiler/api.py
aazzolini: Yep, the SPMD API assumes that the inputs to forward function are implicitly sharded. So I "undo" that in the first iteration, so that we can capture a single-device graph that contains the "full / logical" shapes baked in , and then proceed to the parallelization afterwards.

spmd/compiler/api.py
aazzolini: The changes in _convert_to_distributed are basically meant to fix a bug where we were not picking up changes in non-tensor args.

One of the things that propagate_input_sharding implementations are allowed to do is to modify non-tensor args. Specifically one of the things that we do specially for view ops is to modify the shape input. 

If you look at the implementation for the eager mode propragatrion (here: https://github.com/pytorch/tau/blob/5e4f98b49dcf26254300e66917369049cda84198/spmd/tensor/dispatch.py#L129-L141)

you'll notice that for non-tensor args, it actually uses the args that were modified by the sharding propgataion function implementation.


So what this does in this particular failing example , is that for torch.expand, it changes the input shape from global shape to local shape in the transformed graph.

aazzolini:
gpu tests failing , fixing now.

PR #640, opened 2022-11-22
Replace large-cpu init mode

kwen2501:
- Migrate GPT2 example to use the GSPMD based deferred init mode
- Remove "device" argument from PipelineDriver
- Add `pippy.utils.pp_group_barrier()` to avoid race causing "is not deferred" issue


HamidShojanazeri:
Thanks @kwen2501 , LGTM, and successfully worked on T5 model as well.

HamidShojanazeri (approved):

PR #644, opened 2022-11-22
[PT-D][Tensor Parallelism] Fix parallelize_linear, complete parallelize_module and code refactor

fduwjj:
1. The current rowwise linear is wrong so we need to fix it. The input should be shard(-1)
2. Add _parallelize_linear into parallelize_module
3. Refactor test.


spmd/tensor/__init__.py
```diff
@@ -168,7 +169,11 @@ def replicate_module_params_buffers(m: nn.Module, mesh: DeviceMesh) -> None:
 
     # register input_fn as module forward pre hook
     if input_fn is not None:
-        module.register_forward_pre_hook(lambda _, inputs: input_fn(inputs, device_mesh))  # type: ignore[misc]
+        dim = kwargs.get("input_shard_dim", None)
```
wanchaol, line 172: what's the reason we need this? shouldn't this goes to the make_input_shard/replicate functions directly?

spmd/tensor/parallel/api.py
```diff
@@ -74,6 +74,10 @@ def parallelize_module(  # type: ignore[return]
         device_mesh = _create_1d_device_mesh(device_mesh, tp_mesh_dim)
 
     if isinstance(parallelize_plan, ParallelStyle):
+        # RowwiseParallel or ColwiseParallel
+        if not isinstance(parallelize_plan, PairwiseParallel):
```
wanchaol, line 78: We should do explicitly check IMO for Rowwise and Colwise

spmd/tensor/parallel/api.py
```diff
@@ -251,6 +254,7 @@ def _parallelize_linear(
             _rowwise_parallelize_linear_fn,
             input_fn=parallel_style._prepare_input,  # type: ignore[arg-type, misc] # pyre-ignore[6]
             output_fn=parallel_style._prepare_output,  # type: ignore[arg-type, misc] # pyre-ignore[6]
+            input_shard_dim=-1,
```
wanchaol, line 257: I think this should goes into prepare_input of rowwise parallel style directly?

spmd/tensor/__init__.py
fduwjj: The function itself is stateless, the dim is not stored there.

spmd/tensor/parallel/api.py
fduwjj: OK

spmd/tensor/parallel/api.py
fduwjj: Is there a way for function to store a state?

spmd/tensor/__init__.py
wanchaol: I see, but we have a default argument for dim in make_input_shard? should we put the -1 there?

also, maybe we can use functools.partial to generate a partial function that already have the state?

spmd/tensor/parallel/api.py
awgu: You _could_ store an attribute on the function itself. 
```
>>> def f():
>>>     pass
>>> f.a = 1
>>> f.a
1
```
I am not sure if this is recommended, but there _is_ precedent for this in [old FSDP code](https://github.com/pytorch/pytorch/blob/77d7f2c65945438e0292b270998cea07c0d9d3d8/torch/distributed/fsdp/wrap.py#L226-L227):
```
size_based_auto_wrap_policy.EXCLUDE_WRAP_MODULES = {nn.ModuleList, nn.ModuleDict}  # type: ignore[attr-defined]
size_based_auto_wrap_policy.FORCE_LEAF_MODULES = {nn.MultiheadAttention}  # type: ignore[attr-defined]
```
where `size_based_auto_wrap_policy` is a function.

fduwjj: We will create a new prepare_input func instead.

spmd/tensor/parallel/style.py
```diff
@@ -101,6 +101,30 @@ def make_input_shard_1d(
         )
 
 
+def make_input_shard_1d_dim_last(
```
XilunWu, line 104: thx for fixing the input sharding func

spmd/tensor/parallel/style.py
```diff
@@ -144,14 +168,14 @@ def make_output_shard_1d(
     Convert Output DTensor to a sharded DTensor. This will be used in ParallelStyle.
     Args:
         output (DTensor): output of module to be converted.
-        device_mesh (Optional[DeviceMesh]): :class:``DeviceMesh`` object needed to
+        device_mesh (Optional[DeviceMesh]): :class:`DeviceMesh` object needed to
```
XilunWu, line 171: (Optional[DeviceMesh]) -> (:class:`DeviceMesh`, optional) 
https://github.com/pytorch/pytorch/blob/master/CONTRIBUTING.md#docstring-type-formatting

spmd/tensor/parallel/style.py
```diff
@@ -165,13 +189,13 @@ def make_output_replicate_1d(
     Convert Output DTensor to a replicated DTensor. This will be used in ParallelStyle.
     Args:
         output (DTensor): output of module to be converted.
-        device_mesh (Optional[DeviceMesh]): :class:``DeviceMesh`` object needed to
+        device_mesh (Optional[DeviceMesh]): :class:`DeviceMesh` object needed to
```
XilunWu, line 192: ditto

spmd/tensor/parallel/style.py
```diff
@@ -185,13 +209,13 @@ def make_output_tensor(
     Convert Output DTensor to a replicated DTensor first and then convert it to Tensor.
     Args:
         output (DTensor): output of module to be converted.
-        device_mesh (Optional[DeviceMesh]): :class:``DeviceMesh`` object needed to
+        device_mesh (Optional[DeviceMesh]): :class:`DeviceMesh` object needed to
```
XilunWu, line 212: ditto

spmd/tensor/parallel/style.py
```diff
@@ -185,13 +209,13 @@ def make_output_tensor(
...
             replicate the output and it needs to be a 1D ``device_mesh`` and we will
             throw exceptions if a non-1D ``device_mesh`` is passed in. If no
             ``device_mesh`` is passed in, we will reuse the one from output.
             Default: ``None``
     Return:
-        A :class:``torch.Tensor`` object converted from output DTensor.
+        A :class:`torch.Tensor` object converted from output DTensor.
```
XilunWu, line 218: DTensor -> :class:`DTensor` maybe? Not strictly necessary however.

test/spmd/tensor/parallel/test_parallelize_api.py
```diff
@@ -79,59 +79,64 @@ def test_creat_1d_device_mesh_error(self):
         ):
             _create_1d_device_mesh(mesh, 3)
 
-    def _check_module(
-        self, local_module, dist_module, inp_size, local_input_fn, device_mesh
+    def _compare_params(
```
XilunWu, line 82: thx for refactoring this function!

spmd/tensor/__init__.py
```diff
@@ -1,5 +1,5 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
-from typing import Optional, Sequence, Callable, cast
+from typing import Any, Optional, Sequence, Callable, cast
```
wanchaol, line 2: nit: remove the Any import to avoid lint failure?

spmd/tensor/parallel/api.py
```diff
@@ -278,19 +284,19 @@ def _parallelize_multihead_attn(
 
     Args:
         module (nn.Module):
```
wanchaol, line 286: can we change the all the parameter comment, i.e. :class:`nn.Module` inside the "(:class:`nn.Module`) instead? that would match more to the pytorch comment style.

test/spmd/tensor/parallel/test_parallelize_api.py
```diff
@@ -79,59 +79,64 @@ def test_creat_1d_device_mesh_error(self):
...
-                    ).to_local(),
-                )
+        # check backward and ensure gradients are same
+        self._compare_params(local_module, dist_module, rowwise, True)
 
-            local_optim.step()
-            dist_optim.step()
+        local_optim.step()
+        dist_optim.step()
+        self._compare_params(local_module, dist_module, rowwise, True)
```
XilunWu, line 135: True -> False since we want to check correctness of param update?

XilunWu (commented):
Nice work! left some comments.

wanchaol (approved):
looks good, one suggestion about comment style, thanks for the fix!

spmd/tensor/parallel/api.py
fduwjj: sure. Let me do that.

test/spmd/tensor/parallel/test_parallelize_api.py
fduwjj: opps.

spmd/tensor/parallel/style.py
fduwjj: DTensor part will be removed.

PR #647, opened 2022-11-23
Fix tracer for dict and list inputs

aazzolini:
We need to flatten input arguments first so that we can add all tensor inputs to the list.


aazzolini:
Screen Shot 2022-11-23 at 10 15 03 AM

PR #648, opened 2022-11-23
Update README.md

kwen2501:


README.md
```diff
@@ -492,7 +498,7 @@ We see the model train, memorizing the 512 examples in our input batch:
 
 ## PiPPy on CUDA
 
-When using PiPPy on CUDA devices, the model must be on a CUDA device before being passed to PiPPy, for example:
+When using PiPPy on CUDA devices, the model can be on a CUDA device before being passed to PiPPy, for example:
```
fduwjj, line 501: Does this mean PiPPy now supports non-CUDA model? WoW

fduwjj (approved):
LGTM

README.md
kwen2501: Yes. The model can be in CPU state while being split into stages, then each stage can be materialized on the corresponding GPU only.

PR #649, opened 2022-11-23
[dtensor] remove dtensor from tau as it now lives in core

wanchaol:
we will reuse spmd/tensor folder for extended ops and experimental features.

Note that test_dtensor_ops.py is still there, this is to ensure we could run op db tests in tau until the pytorch version of op db enabled


wz337 (commented):
LGTM!

wz337 (approved):

fduwjj (approved):
Nice, can we also make sure linter is happy so that CI all passed? Thanks!

PR #650, opened 2022-11-28
[PT-D] Add more test cases when we use use_orig_params for FSDP wrapping

fduwjj:
We need to use flag `use_orig_params` in FSDP for ViT model convergence. So let's add more unit tests to ensure it is indeed working as expected.


wz337:
LGTM. Thanks for cleaning up and improving the test!

wz337 (approved):

wanchaol:
PT side PR https://github.com/pytorch/pytorch/pull/89779

Given that now TP/DTensor source of truth lives in pytorch instead of tau after the merge of https://github.com/pytorch/tau/pull/649, closing this in favor of the PT side PR.

PR #651, opened 2022-11-28
Feature dynamic fusion - fuses communication calls based on fusion policy

lessw2020:
This PR implements communication fusion, where they level of fusion is implemented via a simple integer policy (2 = fuse every 2 comm calls into 1), with the idea of later plugging in various fusion policies. 

It uses a single, global comm buffer to optimize for speed and minimal memory overhead.   In effect it is a cascading series of graph rewrites to optimize the communication overhead. 

This PR is dependent upon the status of the clone node removal PR (pending), so for now this adds the impl but is not linked to run via SPMD api in this PR.

### DIFF
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -38,8 +38,8 @@ class FusionElement:
 
     in_graph: bool = False
     processed: bool = False
-    size: int = 0
-    shape: Optional[List[int]] = field(default_factory=lambda: [])  # type: ignore
+    size: Optional[int] = 0
+    shape: Optional[Iterable[[int], [int]]] = field(default_factory=lambda: [])  # type: ignore
     comm_type: Optional[CommType] = None
     node_list: Optional[List[fx.Node]] = field(default_factory=lambda: [])  # type: ignore
     prev_node: Optional[fx.Node] = None  # node that was before start of section
@@ -263,7 +263,7 @@ def _copy_fe_to_buffer(
 
     # create placeholder remapping
     pl_map: Dict[fx.Node, fx.Node] = {}
-    pl_map[pl_list[0]] = gi.global_buffer_node
+    pl_map[pl_list[0]] = gi.global_buffer_node  # type: ignore
 
     for i in range(num_fusion_elements):
         pl_map[pl_list[i + 1]] = in_fe_list[i].grad_tensor_node
@@ -455,7 +455,7 @@ def _update_new_copy_nodes_users(value_remap: Dict[fx.Node, fx.Node]) -> None:
             )
             # if len(node.users) == 0:
             user = node.args[0]
-            node.users[user] = ""
+            node.users[user] = ""  # type: ignore
             node_user_len = len(node.users)
             assert node_user_len, f"failed to update users for node {node.name}"
 
@@ -463,8 +463,8 @@ def _update_new_copy_nodes_users(value_remap: Dict[fx.Node, fx.Node]) -> None:
 def _update_node_tensor_metadata(
     node: fx.Node,
     new_shape: torch.Size,
-    dtype: Optional[torch.dtype] = None,
-    memory_format: Optional[torch.memory_format] = None,
+    in_dtype: Optional[torch.dtype] = None,
+    in_memory_format: Optional[torch.memory_format] = None,
 ) -> TensorMetadata:
     """update a node's metadata to the the new shape, dtype and/or memory format"""
     curr = node.meta.get("tensor_meta")
@@ -473,23 +473,28 @@ def _update_node_tensor_metadata(
     ), f"failed to obtain tensor meta data on node {node.name}"
 
     shape = curr.shape
-    dtype = curr.dtype
+    curr_dtype = curr.dtype
     requires_grad = curr.requires_grad
     stride = curr.stride
 
-    memory_format = curr.memory_format
+    curr_memory_format = curr.memory_format
     is_quantized = curr.is_quantized
     qparams = curr.qparams
+
+    updated_dtype = in_dtype if in_dtype is not None else curr_dtype
+    updated_memory_format = (
+        in_memory_format if in_memory_format is not None else curr_memory_format
+    )
 
     # tempt = torch.empty(new_shape)
     # new_shape = tempt.shape
 
     new_metadata = TensorMetadata(
         new_shape,
-        dtype,
+        updated_dtype,
         requires_grad,
         stride,
-        memory_format,
+        updated_memory_format,
         is_quantized,
         qparams,
     )
@@ -522,7 +527,7 @@ def _finalize_output_node(
     for item in fe_list:
         grad_node = item.grad_tensor_node
         wait_node = item.wait_node
-        replacement_mapping[wait_node] = grad_node
+        replacement_mapping[wait_node] = grad_node  # type: ignore
 
     # we have fused a subset, only update that subset within the larger output node args
     # TODO - this assumes that all gradient tensors are comm handled.
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -21,7 +21,7 @@ from .graph_utils import (
 from .log_utils import rank0_debug
 
 logger: logging.Logger = logging.getLogger(__name__)
-_debug = partial(rank0_debug, logger)
+_debug = partial(rank0_debug, logger)  # type: ignore
 
 # enum for the supported fusion comm types
 class CommType(str, Enum):
@@ -50,7 +50,7 @@ class FusionElement:
 
     def _get_next_node(
         self,
-    ):
+    ) -> fx.Node:
         """get the next node after this FE section"""
         next_node = self.node_list[-1].next  # type: ignore
         # _debug(f"57, next node name is {next_node.name}")
@@ -225,7 +225,9 @@ def _copy_fe_to_buffer(
 
     num_fusion_elements = len(copy_list)
 
-    def copy_to_buffer(buffer, tensor_list):
+    def copy_to_buffer(
+        buffer: torch.Tensor, tensor_list: List[torch.Tensor]
+    ) -> torch.Tensor:
         offset = 0
         for t in tensor_list:
             size = t.numel()
@@ -296,7 +298,9 @@ def _copy_fe_to_buffer(
     buffer_comm_node.update_arg(0, [buffer_node])  # type: ignore
 
 
-def _build_buffer_comm_graph(gm, gi) -> fx.GraphModule:
+def _build_buffer_comm_graph(
+    gi: GraphInfo, gm: fx.GraphModule
+) -> fx.GraphModule:
     """have to make our own all_reduce and wait subgraph for buffer"""
     # from torch.distributed._spmd.comm_tensor import CommTensor
     from torch.distributed.distributed_c10d import (  # ProcessGroup,; Work,; all_reduce,
@@ -337,7 +341,9 @@ def _scatter_results_from_buffer(
     scatter_list = fe_list
     num_fe_items = len(scatter_list)
 
-    def scatter_from_buffer(buffer, scatter_list):
+    def scatter_from_buffer(
+        buffer: torch.Tensor, scatter_list: List[torch.Tensor]
+    ) -> torch.Tensor:
         offset = 0
         for t in scatter_list:
             numel = t.numel()
@@ -350,7 +356,8 @@ def _scatter_results_from_buffer(
     # it is just a proxy for the global buffer node.
     # consider simply saving and reusing a single dummy buffer
 
-    buffer = torch.empty(buffer_size)
+    buffer = gi.tracing_buffer
+    assert buffer is not None, f" missing global tracing buffer in {gi}"
     buffer_shape = buffer.shape
 
     tlist = []
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -2,14 +2,20 @@ import logging
 from copy import deepcopy
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Iterable, List, Optional
+from functools import partial
+from typing import Dict, Iterable, List, Optional
 
 import torch
+import torch.distributed as dist
 import torch.fx as fx
+from torch.fx.experimental.proxy_tensor import make_fx, proxy_slot
+from torch.fx.passes.shape_prop import TensorMetadata
 
 from .graph_utils import (
     OP,
-    get_node_tensor_numel,
+    get_all_nodes_of_type,
+    create_graph_node_map,
+    get_node_tensor_numel_shape,
     get_output_node,
     graph_cleanup,
     pretty_print_graph,
@@ -17,6 +23,7 @@ from .graph_utils import (
 from .log_utils import rank0_debug
 
 logger: logging.Logger = logging.getLogger(__name__)
+_debug = partial(rank0_debug, logger)
 
 # enum for the supported fusion comm types
 class CommType(str, Enum):
@@ -32,25 +39,45 @@ class FusionElement:
     """This class tracks the nodes for a DTensor expanded communication collective in the graph"""
 
     in_graph: bool = False
+    processed: bool = False
+    size: int = 0
+    shape: Optional[List[int]] = field(default_factory=lambda: [])  # type: ignore
     comm_type: Optional[CommType] = None
     node_list: Optional[List[fx.Node]] = field(default_factory=lambda: [])  # type: ignore
-    size: int = 0
     prev_node: Optional[fx.Node] = None  # node that was before start of section
-    next_node: Optional[fx.Node] = None  # node that was after end of section
-    processed: bool = False
     output_name: str = ""
-    clone_node: Optional[fx.Node] = None
     comm_node: Optional[fx.Node] = None
     wait_node: Optional[fx.Node] = None
+    grad_tensor_node: Optional[fx.Node] = None
+
+    def _get_next_node(
+        self,
+    ):
+        """get the next node after this FE section"""
+        next_node = self.node_list[-1].next
+        # _debug(f"57, next node name is {next_node.name}")
+        assert (
+            next_node is not None
+        ), f"failed to get valid next node after {self.node_list[-1].name}"
+        return next_node
 
 
 @dataclass
 class GraphInfo:
+    """provides a home for global aspects of this graph.
+    Currently tracks first and last node, len of the graph and
+    the location and size of the global buffer node
+    """
+
     len: int = 0
-    global_buffer: Optional[fx.Node] = None
+    num_starting_fe: int = 0
+    fe_list: list = None
+    peak_memory_required: int = 0
+    global_buffer_node: Optional[fx.Node] = None
     global_buffer_size: int = 0
-    output: Optional[fx.Node] = None
+    tracing_buffer: Optional[torch.Tensor] = None
     first: Optional[fx.Node] = None
+    output: Optional[fx.Node] = None
 
     def update_info(self, gm: fx.GraphModule) -> None:
         """get the len, input and output nodes"""
@@ -78,19 +105,41 @@ class GraphInfo:
 
 
 def _insert_fusion_buffer_node(
-    gm: fx.GraphModule, insert_before_node: fx.Node, buffer_size: Iterable[int]
+    gm: fx.GraphModule, buffer_size: Iterable[int], gi: GraphInfo = None
 ) -> fx.Node:
-    """insert a torch.empty node in front of insert_before_node"""
+    """insert a torch.empty node for the global buffer.
+    defaults to first node after placeholder nodes.
+    appends to GlobalInfo if passed in"""
+
+    # default to inserting just after last placeholder node
+    for node in gm.graph.nodes:
+        if node.op == OP.PLACEHOLDER:
+            continue
+        insert_before_node = node
+        break
+
+    # TODO - fix with correct rank - needs to match with higher DTensor device
+
+    rank = dist.get_rank()
+    if torch.distributed.is_initialized():
+        torch.cuda.set_device(rank)
+    rank_device = torch.cuda.current_device()
+
     with gm.graph.inserting_before(insert_before_node):
         new_buffer_node = gm.graph.create_node(
             OP.CALL_FUNCTION,
             target=torch.empty,
             # TODO - need device from DTensor to put buffer on gpu
-            args=tuple(buffer_size),
+            args=(buffer_size,),
+            kwargs={"device": rank_device},
         )
     assert (
         new_buffer_node is not None
     ), f"failed to create buffer node, size={buffer_size}"
+
+    if gi is not None:
+        gi.global_buffer_node = new_buffer_node
+        gi.global_buffer_size = buffer_size
 
     return new_buffer_node
 
@@ -105,10 +154,10 @@ def _scan_graph_for_fusion_elements(
     element_list = []
 
     fe_sequence = [
-        "clone",
+        # "clone",
         "_tensor_constant",
         "_tensor_constant",
-        CommType,
+        comm_type,
         "comm_result",
         "getitem",
         "getitem",
@@ -144,15 +193,20 @@ def _scan_graph_for_fusion_elements(
                 # need to fully populate this fe...
                 # we will be removing/rewriting the node list so we save prev and next
                 fe.prev_node = curr_node_list[0].prev
+                _debug(f"prev node = {fe.prev_node}")
                 fe.next_node = node.next
 
                 fe.output_name = node.name
                 fe.wait_node = node
+                fe.comm_node = curr_node_list[2]
 
-                fe.clone_node = curr_node_list[0]
-                fe.comm_node = curr_node_list[3]
+                fe.grad_tensor_node = fe.comm_node.args[0][0]
 
-                fe.size = get_node_tensor_numel(fe.clone_node)  # type: ignore
+                size, shape = get_node_tensor_numel_shape(fe.grad_tensor_node)  # type: ignore
+                fe.size = size
+                fe.shape = shape
+                _debug(f"\nfe list size shape {size=}, {shape=}\n")
+
                 element_list.append(fe)
 
             index = 0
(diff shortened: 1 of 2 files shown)


spmd/compiler/fusion.py
```diff
@@ -78,20 +103,44 @@ def update_info(self, gm: fx.GraphModule) -> None:
 
 
 def _insert_fusion_buffer_node(
-    gm: fx.GraphModule, insert_before_node: fx.Node, buffer_size: Iterable[int]
+    gm: fx.GraphModule,
+    buffer_size: int,
+    gi: Optional[GraphInfo] = None,
 ) -> fx.Node:
-    """insert a torch.empty node in front of insert_before_node"""
+    """insert a torch.empty node for the global buffer.
```
aazzolini, line 110: qq should this come right before the fused op (instead of right after the placeholder) -- to avoid memory overhead? (just a qq, not blocker for this pR) >

spmd/compiler/fusion.py
```diff
@@ -32,25 +37,45 @@ class FusionElement:
...
-    comm_node: Optional[fx.Node] = None
-    wait_node: Optional[fx.Node] = None
+    comm_node: Optional[fx.Node] = None  # type: ignore
+    wait_node: Optional[fx.Node] = None  # type: ignore
+    grad_tensor_node: Optional[fx.Node] = None  # type: ignore
+
+    def _get_next_node(
+        self,
+    ) -> fx.Node:
+        """get the next node after this FE section"""
```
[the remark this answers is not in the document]
aazzolini: Ohh ok now i understand what it is , not clear for someone reading it for the first time , maybe just change comment to FusionElement directly .

spmd/compiler/fusion.py
```diff
@@ -32,25 +37,45 @@ class FusionElement:
...
+        assert (
+            next_node is not None
+        ), f"failed to get valid next node after {self.node_list[-1].name}"  # type: ignore
+        return next_node
 
 
 @dataclass
 class GraphInfo:
+    """provides a home for global aspects of this graph.
+    Currently tracks first and last node, len of the graph and
```
aazzolini, line 66: What does "first" and "last" node mean?

spmd/compiler/fusion.py
```diff
@@ -144,15 +193,18 @@ def _scan_graph_for_fusion_elements(
                 # need to fully populate this fe...
                 # we will be removing/rewriting the node list so we save prev and next
                 fe.prev_node = curr_node_list[0].prev
-                fe.next_node = node.next
+                # fe.next_node = node.next
```
aazzolini, line 196: why is this commented out?

spmd/compiler/fusion.py
```diff
@@ -105,10 +154,10 @@ def _scan_graph_for_fusion_elements(
     element_list = []
 
     fe_sequence = [
-        "clone",
+        # "clone",
```
aazzolini, line 157: why is this commented out?

aazzolini (commented):
This overall looks in good shape. Some comments / questions for this PR:

1) Could we add unit tests to make sure that we check transformation behavior?
2) Could you explain what are the current limitations of the approach and what are the next steps?

spmd/compiler/fusion.py
```diff
@@ -32,25 +37,45 @@ class FusionElement:
     """This class tracks the nodes for a DTensor expanded communication collective in the graph"""
```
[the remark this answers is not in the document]
lessw2020: right now it is DTensor specific in that the DTensor expansion will wrapper the all_reduce (or whatever comm) in a CommTensor, which adds the get_item and wait_comm nodes. 
I reuse the wait comm node and we search for it atm, so in that sense it is DTensor expansion specific.
It can be updated/adapted to any comm (from DTensor or not) but would need to add a check whether the signature wait_comm was present or not, and add it if needed.

spmd/compiler/fusion.py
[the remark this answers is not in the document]
lessw2020: yes, good point!  Have updated to FusionElement instead of just FE.

spmd/compiler/fusion.py
lessw2020: great point - I was thinking of forwards pass and it would not matter re: memory pressure, but you are right.  
For backwards activations are being freed as it progresses, so we should hold off on creating the global buffer until first use to reduce peak mem pressure.
Not hard to change, but will require some reordering as we have to do it after determining fusion plan...thus prefer to get this PR in as is, and then update so I have made this tracking issue for it:
https://github.com/pytorch/tau/issues/654

spmd/compiler/fusion.py
lessw2020: this is the very first and very last node of the graph.  
Atm, fx.graph does not keep/expose a tail, so I wanted to be able to store and access front and back since we have to do a full graph scan to start.  (you can reverse iter but it's not clear that internally it isn't running the entire graph to do that first, which could be slow for larger graphs).

spmd/compiler/fusion.py
lessw2020: Originally this was holding the 'next' out of comm section node (i.e we hold a prev node which is the last node before the start of comms, and a next that is the start of the regular graph (after comms).  However, next is not a reliable marker as we started doing fusions since that next node will either a) become irrelevant since this comm call was removed as part of being fused later or b)the fusion itself inserts the buffer scatter nodes here meaning that the whole comm section has expanded.  
I replaced that static pointer with the dynamic function  _get_next_node() for FusionElements. 
I have gone ahead and remove the above line so it's clear we no longer use it.

spmd/compiler/fusion.py
lessw2020: AOT inserts clone nodes in front of the DTensor comm expansion by default (per Sanket it has something to do with assuming people will not be adjusting/using(?) the backwards graph and it's some kind of generic debugging thing).
They are not needed for our use, and it's wasteful memory and perf wise (each clone is a CPU copy of the tensor) and also complicates the graph.  Thus I wanted to eliminate them. 
This is done in this PR https://github.com/pytorch/tau/pull/631 but I haven't finished the review for it. 
For now I commented out  the clone node reference assuming the 631 PR will land shortly and the graph will not have clones...but in case 631 does not land, then wanted to note where we are tracking clone nodes.  (will remove it instead of comment once the clone PR is finalized).

lessw2020:
Regarding the general questions:

1 - Could we add unit tests to make sure that we check transformation behavior?

Yes, but I'd like to do it separately vs this PR as this PR is already dependent on another PR (the clone node PR).  That's in part why fusion is not hooked in for this PR (adds the code, but not live/linked to SPMD api yet. 
 Once the clone PR is finalized, then can do the unit tests.  Have opened this issue to track:
https://github.com/pytorch/tau/issues/655

2 - Could you explain what are the current limitations of the approach and what are the next steps?
Sure - current limits:
a - we expect the DTensor signature for comms as we reuse the existing wait_comm nodes for where fusion is implemented.  Thus need to add ability to generate our own wait_comm nodes if we want to have it handle any comm calls for fusion. 
b - right now we assume that all calls use the same process group (and thus any call is eligible for fusion with any other call).  That is not correct for all cases and so we need to add inspection of PG and then only fuse same PG calls in the future. 
(added tracking issue: https://github.com/pytorch/tau/issues/657) 
c - similarly, we assume we need an FP32 buffer but we should inspect the dtype and then enable appropriate dtype buffer.  
(added tracking issue: **https://github.com/pytorch/tau/issues/656**)

lessw2020:
Next steps:
a - get clone situation finalized
b - add fusion policy based on requested comm size (currently impl as merging of number of calls to merge)
c - add unit tests
d - integrate reader function that can read the policies expressed by variable bucket size policy (where we will have erratically spaced fusions).  This seems to be the most performant per the simulation so need to get a real graph and add a reader to interpret that. 
e - add reader and functions for for capuchin policy (dynamic activation checkpointing) where we will checkpoint only some tensors and swap others.  This is arguably the other 'big' improvement for perf in the immediate pipeline. 
f - add shifting support where we move some comms into forward pass to improve net backwards pass time.

spmd/compiler/fusion.py
```diff
@@ -2,21 +2,26 @@
...
 import torch
+import torch.distributed as dist
 import torch.fx as fx
+from torch.fx.experimental.proxy_tensor import make_fx
+from torch.fx.passes.shape_prop import TensorMetadata
 
 from .graph_utils import (
-    OP,
-    get_node_tensor_numel,
+    OP,  # get_all_nodes_of_type,; pretty_print_graph,
```
anj-s, line 15: what is this comment about?

spmd/compiler/fusion.py
```diff
@@ -32,25 +37,45 @@ class FusionElement:
     """This class tracks the nodes for a DTensor expanded communication collective in the graph"""
 
     in_graph: bool = False
+    processed: bool = False
```
anj-s, line 40: Can we add a comment for what each of these properties represent?

spmd/compiler/fusion.py
anj-s, line 54: nit: Can you capitalize the first letter in your doc comments?

spmd/compiler/fusion.py
```diff
@@ -32,25 +37,45 @@ class FusionElement:
...
+    comm_node: Optional[fx.Node] = None  # type: ignore
+    wait_node: Optional[fx.Node] = None  # type: ignore
+    grad_tensor_node: Optional[fx.Node] = None  # type: ignore
+
+    def _get_next_node(
+        self,
+    ) -> fx.Node:
+        """get the next node after this FE section"""
+        next_node = self.node_list[-1].next  # type: ignore
+        # _debug(f"57, next node name is {next_node.name}")
```
anj-s, line 56: remove?

spmd/compiler/fusion.py
anj-s: I had the same question in the doc. Can you mention this in your comment as well?

spmd/compiler/fusion.py
```diff
@@ -32,25 +37,45 @@ class FusionElement:
...
 @dataclass
 class GraphInfo:
+    """provides a home for global aspects of this graph.
+    Currently tracks first and last node, len of the graph and
+    the location and size of the global buffer node
+    """
+
     len: int = 0
-    global_buffer: Optional[fx.Node] = None
+    num_starting_fe: int = 0
```
anj-s, line 71: same comment as above for adding comments for each of the properties.

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+    #    if node.op == OP.OUTPUT:
+    #       new_output = node
+    #       break
+    # _debug(f"703, updated output node args {new_output.args=}\n")
+
     # final review print
     graph_cleanup(gm)
 
-    pretty_print_graph(gm, "final version, fusion pass")
+    # TODO - remove this.  This is purely a final check to verify all meta data
```
anj-s, line 648: remove?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+
+    return peak_memory
+
+
+def run_comm_fusion(gm: fx.GraphModule) -> fx.GraphModule:
     """main entry into remapping graph for all_reduce fusion"""
 
-    result = False
+    # result = False  TODO - do we return a success code or the graph,
+    # since graph is modified directly in place?
```
anj-s, line 575: Can we skip returning anything? Any errors should be immediately thrown and in place graph modifications mean we don't need to return anything right?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+
+    # first recompile to make sure we have coherent graph
```
anj-s, line 577: Why do we need this? When would this fail?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+    gm.recompile()
 
     # get our main graph info
-    graph_info = GraphInfo()
-    graph_info.update_info(gm)
+    gi = GraphInfo()
```
anj-s, line 581: nit: lets rename to graph_info for readability

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
 
     # scan graph for all comm sections (fusion elements)
     fe_list = _scan_graph_for_fusion_elements(gm, comm_type=CommType.allreduce)
 
+    gi.num_starting_fe = len(fe_list)  # type: ignore
+    gi.fe_list = fe_list
+
+    # simple fusion policy where int = num buckets to fuse...start with 2,
+    # meaning every 2 comms are fused into 1
+    fusion_policy: int = 2
```
anj-s, line 594: I had a question about this in the doc. Shouldn't the fusion policy be driven by the bucket size that we want?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+    # _debug(f"537 - updated output args = {new_output_args}\n")
+
+
+def _determine_peak_memory(gi: GraphInfo, fusion_policy: int) -> int:
+    """
+    scans fe list to determine max memory required across all fusion instances.
+    this result is used to allocate the global buffer for fusion, where we
+    re-use a global buffer to avoid repeated allocations per fusion.
+    """
+    peak_memory = 0  # currently measured in numel
```
anj-s, line 553: This will depend on the precision used for grad reduction right? Should we figure out a way to incorporate that policy here?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+    curr_memory = 0
+    fast_index = 0
```
anj-s, line 555: nit:rename this cur_fusion_index?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+    # determine peak memory using fusion policy
+    peak_memory_required = _determine_peak_memory(gi, fusion_policy)
+
+    buffer_node = _insert_fusion_buffer_node(gm, peak_memory_required, gi)
+
+    # Main process loop - iterate all fusion elements, apply fusion to subsets
+
+    offset = 0
+    count = 0
+    # new_output_args=[]
```
anj-s, line 605: remove?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
 
-def run_comm_fusion(gm: fx.GraphModule) -> bool:
+def _copy_fe_to_buffer(
+    gi: GraphInfo, gm: fx.GraphModule, in_fe_list: list[FusionElement]
+) -> None:
+    """first half of fusion - move desired items to buffer and create graph"""
+    buffer_node = gi.global_buffer_node
+    buffer_size = gi.global_buffer_size
+
+    copy_list = in_fe_list
```
anj-s, line 224: Should we explicitly copy the list? What is tlist below doing?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+        buffer = buffer = torch.empty(buffer_size)
+
+    tlist = []
+    for item in copy_list:
+        a = torch.zeros((item.shape[0], item.shape[1]))  # type: ignore
+        tlist.append(a)
+
+    load_gm = make_fx(copy_to_buffer)(buffer, tlist)
+
+    subnodemap = create_graph_node_map(load_gm)
```
anj-s, line 253: can we rename this to reflect that this is a mapping from node.name to node?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+
+    # update load loop to use main graph items
+    fn_list = []
```
anj-s, line 256: can we use full names for var names here?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+            pl_list.append(node)
+        elif node.op == OP.CALL_FUNCTION:
+            fn_list.append(node)
+
+    # create placeholder remapping
+    pl_map: Dict[fx.Node, fx.Node] = {}
+    pl_map[pl_list[0]] = gi.global_buffer_node  # type: ignore
+
+    for i in range(num_fusion_elements):
+        pl_map[pl_list[i + 1]] = in_fe_list[i].grad_tensor_node  # type: ignore
```
anj-s, line 269: can we add a comment here that mentions the mapping from the buffer nodes for the gradients.

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+
+    def remap_copy_args(in_node: fx.Node) -> fx.Node:
+        out_node = in_node
+        if in_node in pl_map:
+            out_node = pl_map[in_node]  # type: ignore
+        elif in_node in value_remap:
+            out_node = value_remap[in_node]
+        return out_node
+
+    value_remap = {}
```
anj-s, line 281: should this be defined before the above function?

anj-s:
> Regarding the general questions:
> 
> 1 - Could we add unit tests to make sure that we check transformation behavior?
> 
> Yes, but I'd like to do it separately vs this PR as this PR is already dependent on another PR (the clone node PR). That's in part why fusion is not hooked in for this PR (adds the code, but not live/linked to SPMD api yet. Once the clone PR is finalized, then can do the unit tests. Have opened this issue to track: #655
> 
> 2 - Could you explain what are the current limitations of the approach and what are the next steps? Sure - current limits: a - we expect the DTensor signature for comms as we reuse the existing wait_comm nodes for where fusion is implemented. Thus need to add ability to generate our own wait_comm nodes if we want to have it handle any comm calls for fusion. b - right now we assume that all calls use the same process group (and thus any call is eligible for fusion with any other call). That is not correct for all cases and so we need to add inspection of PG and then only fuse same PG calls in the future. (added tracking issue: #657) c - similarly, we assume we need an FP32 buffer but we should inspect the dtype and then enable appropriate dtype buffer. (added tracking issue: **#656**)

Thanks Less! Why do we need the clone node PR for adding unit tests? Can we add a unit test without the SPMD linking?

spmd/compiler/fusion.py
wconstab, line 157: it might be worth considering using the FX pattern matcher engine instead of your own loop.
`pytorch/torch/fx/passes/utils/matcher_utils.py`

Probably takes some getting used to and not sure if it'll save you a lot of LOC here though...

Also want to point out some helpers in
`pytorch/torch/fx/passes/graph_manipulation.py`
which might be useful given some of the things you have below
(get_size_of_node, replace_target_nodes_with, ...)

spmd/compiler/fusion.py
```diff
@@ -32,25 +37,45 @@ class FusionElement:
...
+        next_node = self.node_list[-1].next  # type: ignore
+        # _debug(f"57, next node name is {next_node.name}")
+        assert (
+            next_node is not None
+        ), f"failed to get valid next node after {self.node_list[-1].name}"  # type: ignore
+        return next_node
 
 
 @dataclass
 class GraphInfo:
```
wconstab, line 64: wonder if it makes sense to just stick this info onto the GraphModule you are modifying rather than having a second object to pass around?

spmd/compiler/fusion.py
```diff
@@ -78,20 +103,44 @@ def update_info(self, gm: fx.GraphModule) -> None:
...
+
+    rank = dist.get_rank()
+    if torch.distributed.is_initialized():
+        torch.cuda.set_device(rank)
+    rank_device = torch.cuda.current_device()
+
     with gm.graph.inserting_before(insert_before_node):
         new_buffer_node = gm.graph.create_node(
             OP.CALL_FUNCTION,
             target=torch.empty,
```
wconstab, line 131: did you consider the tradeoff between calling torch.empty every iteration, vs allocating the memory at compile time and stashing the reusable buffer as an attribute of the graphmodule?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+    for i, item in enumerate(gi.fe_list):  # type: ignore
+        fast_index += 1
+        curr_memory += item.size  # type: ignore
+
+        if fast_index == fusion_policy:
+            peak_memory = max(peak_memory, curr_memory)
+            fast_index = 0
+            curr_memory = 0
+
+    # _debug(f"574, peak memory determined to be {peak_memory}")
```
wconstab, line 565: better to leave these lines uncommented and use `log.debug` so the info is controlled by log levels

spmd/compiler/fusion.py
wconstab, line 605: cleanup

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+            offset += size
+        return buffer
+
+    # setup dummy vars
+    buffer = None
+    if gi.tracing_buffer is None:
+        buffer = torch.empty(buffer_size)
+        gi.tracing_buffer = buffer
+    elif gi.tracing_buffer:
+        buffer = buffer = torch.empty(buffer_size)
```
wconstab, line 244: s/elif/else? 

remove `buffer = buffer = ...`?  or is it supposed to be `gi.tracing_buffer = buffer = ...`?

spmd/compiler/fusion.py
wconstab, line 224: any reason for not using in_fe_list consistently?

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+
+    tlist = []
+    for item in copy_list:
+        a = torch.zeros((item.shape[0], item.shape[1]))  # type: ignore
```
wconstab, line 248: you'd want to use zeros_like  probably, in case shape is not 2-dim (or else assert it is 2-dim?)

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
     return element_list
 
 
-def run_comm_fusion(gm: fx.GraphModule) -> bool:
+def _copy_fe_to_buffer(
```
wconstab, line 217: i'm pretty confused by this whole function.  it seems overly complicated.  why the need for separate tracing buffers and tracing a copy function?  i would try to just directly construct fx nodes that interact with my global buffer.

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+    buffer_comm_node = in_fe_list[-1].comm_node
+    buffer_comm_node.update_arg(0, [buffer_node])  # type: ignore
+
+
+def _build_buffer_comm_graph(
+    gi: GraphInfo, gm: fx.GraphModule
+) -> fx.GraphModule:
+    """have to make our own all_reduce and wait subgraph for buffer"""
+    # from torch.distributed._spmd.comm_tensor import CommTensor
+    from torch.distributed.distributed_c10d import (  # ProcessGroup,; Work,; all_reduce,
```
wconstab, line 306: import at the top? or is this a circular issue

spmd/compiler/fusion.py
lessw2020: oh formatter must have mushed that up there.  I've removed completely as not needed.

spmd/compiler/fusion.py
```diff
@@ -162,22 +214,448 @@ def _scan_graph_for_fusion_elements(
...
+    )
+
+    buffer_size = gi.global_buffer_size
+
+    def dummy_add(
+        grad_buffer: torch.Tensor, zero: torch.Tensor
+    ) -> torch.Tensor:
+        return grad_buffer + zero
+
+    grad_buffer: torch.Tensor = torch.empty(buffer_size)
```
wconstab, line 318: i would probably also have just created an fx.Graph, added 2 placeholders to it with names, and then inserted a call_function(torch.add, a, b) and an output node.  Not necessary to allocate real tensors and trace a graph.  possibly also fuse this whole function inline into wherever it is called, so you don't have to iterate this new graph and copy its nodes into another graph- just create them in the right place up front?

wconstab (commented):
Overall seems like it probably works, but it's a bit complicated and hard to follow.  Might be worth refactoring with aim to reduce LOC, utilize existing FX utils.

spmd/compiler/fusion.py
lessw2020: Yes - good idea.  I've updated all doc strings for current and will continue with that style going forward.

spmd/compiler/fusion.py
lessw2020: removed

spmd/compiler/fusion.py
lessw2020: added comments for all:
updated_comments

spmd/compiler/fusion.py
lessw2020: Yes, we can remove this if we want to proceed with the basis that errors will result in stoppage of the fusion process rather than a result flag (I had some try except blocks earlier but I've removed).  
I was also thinking originally in a Rust way where we always return a Result enum which has both return value and err value to double check everything and I was planning to mimic that. 
(also you are correct - graph is in place modified, I was returning to signal modification but I guess thats known up front.)

spmd/compiler/fusion.py
lessw2020: It can fail when nodes have been silently/unexpectedly removed by fx recompile or if for some reason, it doesn't want to accept updates to a nodes args.  
I had some try/except guards earlier but have removed now that the process is more ironed out, and the enforcing user count updates on the copy nodes has largely removed this risk. 
Per the above issue, I've updated the doc string to make it clear errors will result in stoppage rather than attempting to handle, and removed the above.

spmd/compiler/fusion.py
lessw2020: sorry for the confusion here - this is a half completed function that atm is not used.  We do a full trace b/c ultimately this has to be wrapped with a CommTensor to produce and wait on the comm work handle.  
For now I am re-using existing wait comms in the graph rather than writing our own which was why this function is only half complete. 
I've updated in the comments that this is a stub to make it clearer.

spmd/compiler/fusion.py
lessw2020: atm this is a stub function so I kept the includes local to it rather than at the top.  
Also I see black has scrunched the commented out aspects in 306 above and have removed.
I've also added a comment in the doc string that this is only a stub for clarity.

spmd/compiler/fusion.py
lessw2020: The tracing allows me to leverage fx unrolling and making a series of nodes for what is otherwise a pretty complex series of node additions.  I would not want to do this by hand inserting nodes unless I'm missing something (though in thinking about this perhaps you mean taking the view/slice/copy nodes and then programmatically looping and adding everything that way?)

For context - Below is the pre and post nodes added by tracing and unrolling for a simple 2 call fusion:

annotated_final_graph

In green highlight are the pre view/slice/copy nodes added in the pre aspect and after is the slice/view/copy nodes for moving back out.
Note that the buffer is a 1D buffer and thus we reshape in and out to handle any shapes seen.  If the buffer matched the dims then this would be a lot fewer nodes and could make more sense to just insert directly. 
I think though what you mean is to build a function that directly inserts nodes based on the known view/slice/copy for in and slice/view/copy for out instead of tracing and creating subgraphs?

spmd/compiler/fusion.py
lessw2020: good point!  removed comments.

spmd/compiler/fusion.py
lessw2020: good catch, removed.

spmd/compiler/fusion.py
lessw2020: this was a mistake - updated and thanks for catching.  
this is the reuse of the global tracing buffer where we check if already created, use it, otherwise create and assign to the gi for future re-use.

spmd/compiler/fusion.py
lessw2020: good points!  I did look at the pattern matcher but it seemed a bit overkill given that I know we have a series of sequential node patterns and can succeed with a single pass vs pattern matcher is digging into the node args and reverse running the graph to deal with this with an exponential approach.
Longer term, I'd say this is the correct approach with ultimately pushing some utils I've made back into the graph codebase directly rather than maintaining our own graph_utils file.  
I've added a todo in the design doc to track.

spmd/compiler/fusion.py
lessw2020: not overall. I was using the in_fe_list before splitting the function into two different ones and this was a rename to make the code clearer in the copy side.
I've removed in_fe_list and simply work with copy_list as the input param and stay consistent from there. 
Thanks for the spot!

spmd/compiler/fusion.py
lessw2020: Actually I think you may have a good point - this could be done with a hybrid loop of node insertion to replicate the view/slice/copy and offset adjustments to replicate the loop unrolling via nodes and be done without the actual tracing. 

I've added it to the design doc and will test it out as a potential simplification. 
Thanks for the idea, gave you credit in the design doc.

spmd/compiler/fusion.py
lessw2020: definitely better to use zeros_like, thanks a bunch for this catch!
updated to zeros_like.

spmd/compiler/fusion.py
lessw2020: good  catch.  I actually just removed this as it was an artifact of when it was originally a single function (which was split into two). 
Now the incoming param is copy_list  instead of in_fe_list, and that is used continuously throughout the function.

spmd/compiler/fusion.py
lessw2020: yes removed.

spmd/compiler/fusion.py
lessw2020: good point. I updated to curr_fe_index since it represents the count of active fusion elements (fe) for this fusion.

spmd/compiler/fusion.py
lessw2020: yes - changed name to create_name_to_node_map to make it clearer.

spmd/compiler/fusion.py
lessw2020: yes correct.  I've entered a tracking issue so this can be added after the PR.  
https://github.com/pytorch/tau/issues/656

spmd/compiler/fusion.py
lessw2020: it technically works as is b/c the earlier use is an embedded function that is only invoked after this...but agree it's confusing and cleaner to have it above. 

Moved it up and also added in typing for it:
value_remap: Dict[fx.Node, fx.Node] = {}

spmd/compiler/fusion.py
lessw2020: added comment clarifying pl map usage.

spmd/compiler/fusion.py
lessw2020: yes, removed.  I think ultimately we may want to add some generic debug utils in graph_utils as we encounter complex graphs but will do that as functions, instead of the above commented out sections :)

spmd/compiler/fusion.py
lessw2020: it can and ultimately will be as one type of optimization.  I went with integer to start b/c goal was to ensure accurate fusion handling and so easier to verify by delink byte computations for bucket and simply modifying model sizes based on varying layers. 
I will add in the bucket size policy as next option, and have added this to track it:
https://github.com/pytorch/tau/issues/659

lessw2020: other aspect was bucket compute basis can result in a whole large number of fusion calls being pushed to the end if bucket size is not somewhat optimized, at which point you have minimal comm/compute overlap and have slowed things down.  
Integer based somewhat avoids that by ensuring a more even spread of fused calls throughout the graph by default for better comm/compute overlap.

spmd/compiler/fusion.py
lessw2020: updated to graph_info!

lessw2020:
> > Regarding the general questions:
> > 1 - Could we add unit tests to make sure that we check transformation behavior?
> > Yes, but I'd like to do it separately vs this PR as this PR is already dependent on another PR (the clone node PR). That's in part why fusion is not hooked in for this PR (adds the code, but not live/linked to SPMD api yet. Once the clone PR is finalized, then can do the unit tests. Have opened this issue to track: #655
> > 2 - Could you explain what are the current limitations of the approach and what are the next steps? Sure - current limits: a - we expect the DTensor signature for comms as we reuse the existing wait_comm nodes for where fusion is implemented. Thus need to add ability to generate our own wait_comm nodes if we want to have it handle any comm calls for fusion. b - right now we assume that all calls use the same process group (and thus any call is eligible for fusion with any other call). That is not correct for all cases and so we need to add inspection of PG and then only fuse same PG calls in the future. (added tracking issue: #657) c - similarly, we assume we need an FP32 buffer but we should inspect the dtype and then enable appropriate dtype buffer. (added tracking issue: **#656**)
> 
> Thanks Less! Why do we need the clone node PR for adding unit tests? Can we add a unit test without the SPMD linking?

The main issue is that depending on how the clone node PR lands (i.e. it goes in or does not, and if it goes in, which PR as two ways to fix), the handling of the graph for fusion will need to adjusted, which then flows into also changing the unit tests.
So it is not "must do first" but rather a nice to do to avoid future make work to adjust for the clone situation. 

Re: add unit test without the SPMD linking - 
we could make a simple test using a prebuilt backwards graph and just supply that for unit testing.  That would remove the need to link SPMD, but the drawback is then we are not really testing that changes in SPMD don't break fusion, which would be the main point of the unit tests.   
That's why I was thinking better to have SPMD linked in so unit testing is actually providing comprehensive tests, but we could start with a simple standalone for now.

anj-s (approved):

PR #653, opened 2022-11-28
try using aot_module_simplified

wanchaol:


spmd/compiler/api.py
```diff
@@ -433,11 +433,14 @@ def gather_inputs_for_compilation(
                 )
                 return compile_inps
 
-            self._compiled_m = aot_module(
+            self._compiled_m = aot_module_simplified(
                 self._local_module,
                 partial(self._compile, TrainingPhase.FORWARD),
                 partial(self._compile, TrainingPhase.BACKWARD),
-                pre_compile_fn=gather_inputs_for_compilation,
+                # pre_compile_fn=gather_inputs_for_compilation,
```
anj-s, line 440: Can we remove the comment here and add a comment below for why we need this? Also what are the downsides with this approach (if any)?

fegin:
Can we add a description about this PR, especially what's the difference of aot_module_simplified that can allow us to remove the patch?

wanchaol:
@anj-s @fegin sorry I should have mark this as a draft PR, copying @aazzolini investigation on this:

1) the functionality we need (passing different inputs for compilation vs. runtime) is now expossed in aot_module_simplified.
2) however, aot_module_simplified would require our inputs/outputs to be "flattened", which is not the case in SPMD contract right now.
3) Adding an additional flattening/unflattening step would increase complexity of SPMD impl. 
4) I believe for now is best to keep things as is, and Wanchao's fix to my patch should suffice for now. 
5) After prototype phase we may want to revisit the entire wrapping / aot_autograd mechanism anyways if we want it to be part of the dynamo/inductor pipeline.

Given this I will convert this as a draft PR and we can discuss later on how we should resolve it in the future

PR #658, opened 2022-11-29
[SPMD] Port support for torch.cat 

anj-s:
Port torch.cat and torch.slice_scatter.
* Added unit test 
* Account for dims < 0


aazzolini (approved):
Looks great!

PR #652, opened 2022-11-28
[SPMD] Split the SPMD flow to tracing, profiling and optimization stages

fegin:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #652

**Motivation:**
This PR splits the SPMD flow to tracing, profiling and graph optimization three stages for following reasons:
1. Insert the placeholders for profiling and graph optimization stages.
2. Can reuse profiling and graph optimization for already traced graphs.
3. Reduce the code conflict when people are developing different stages.

**What are in this PR?:**
1. Separate the tracing and distributing part to a standalone module. The api.py contains only the main flow.
2. Add DistributeGraph to harness the profiling and graph optimization stages.
3. Add GraphOptimization to chain different graph optimization (we only have 1 now).

PR #660, opened 2022-11-30
[SPMD] Split the SPMD flow to tracing, profiling and optimization stages

fegin:
Resubmit https://github.com/pytorch/tau/pull/652/ as ghstack seems to fail to merge it to the main branch.


aazzolini:
Approving since it's the same as the other PR that was already approved.

aazzolini (approved):

PR #661, opened 2022-11-30
Feature run spmd - add standalone driver program for SPMD work

lessw2020:
This PR adds a driver program, run_spmd.py, that can be used to run SPMD directly in a configurable environment with a given sample model.
Usage is simply "python run_spmd.py" from the spmd directory. 
It sets up and runs with a configurable world size and configurable models.
It then runs the model using both DDP and SPMD, and then does a comparison of the grads returned via allclose as well as a two row visual display of a grad subset. 
run_smpd_results

By default it runs a small 2 layer linear model, with a configurable layer size. 
Also in there is the permute model subset from Rich. 

I've been using it for the fusion work and found it really helpful for being able to quickly run and experiment/verify things. Thus wanted to add it directly to the repo so hopefully it is useful for others by being able to quickly spin up a given model and environment for SPMD.


fegin:
General comment, we should move this file to an example folder or testing folder but not just reside in the plain SPMD folder as this is the entry point of all the functionality APIs.

lessw2020:
> General comment, we should move this file to an example folder or testing folder but not just reside in the plain SPMD folder as this is the entry point of all the functionality APIs.

Ok moving to /examples/spmd/run_spmd.py

examples/spmd/run_spmd.py
```diff
@@ -0,0 +1,188 @@
...
+    # main work
+    work_main(rank, world_size)
+
+    # teardown
+    teardown(rank)
+
+
+if __name__ == "__main__":
+    os.environ["MASTER_ADDR"] = "localhost"
+    os.environ["MASTER_PORT"] = "29502"
```
fegin, line 184: Prefer to have a random number > 1024. Otherwise the same program executing twice may get the port being occupied message.

examples/spmd/run_spmd.py
```diff
@@ -0,0 +1,188 @@
...
+        super().__init__()
+        self.w = torch.nn.Parameter(torch.rand((5, 10)))
+        self.b = torch.nn.Parameter(torch.rand((5)))
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x_t = x.permute(0, 2, 1)
+        return torch.nn.functional.linear(x_t, self.w, self.b)
+
+
+class replicaModel(nn.Module):
```
fegin, line 91: nit: `ReplicaModel`. Python naming rule.

fegin (approved):
LGTM, have some nits.

examples/spmd/run_spmd.py
lessw2020: my work in Rust continues to show lol.
fixed and thanks.

examples/spmd/run_spmd.py
lessw2020: good idea.  added random port selection with seed. 

Went with the recommended dynamic range for port selection to be safe:

"Dynamic ports—Ports in the range 49152 to 65535 are not assigned, controlled, or registered. They are used for temporary or private ports. They are also known as private or non-reserved ports. Clients should choose ephemeral port numbers from this range..."

PR #662, opened 2022-12-01
Fix SPMD expansion and test_tracing.py

aazzolini:
Turns out that our unit tests were not properly testing SPMD because we had the same random_seed set on all ranks which means all parameters and inputs were the same. After setting different random seeds on each rank, the tests started failing, but then passed again after fixed a bug in distribute.py (we were not adding the collectives at the end because we were passing _allow_partial=False)


wanchaol (approved):
thanks for the fix!

anj-s (approved):

PR #664, opened 2022-12-01
Fix issue when backward graph returning sharded tensors

aazzolini:
When SPMD is part of a larger model, backard() could output sharded
    gradients (for input) as well as replicated gradients (for parameters).
    
    This fixes a bug that was causing for non-partial/ non-replicated
    outputs to disappear from the graph output.
    
    --
    
    Also, not necessary for this specific fix, but something that could
    trigger issues down the road -- the output_specs were not reflecring the
    Partial -> Replicated  resharding on the outputs. This fixes it so that
    _convert_to_distributed always returns the proper output sharding for
    the graph.
    (One thing we can do since we have the output schemas: return DTensors from SPMD (optionally)

### DIFF
--- a/spmd/compiler/distribute.py
+++ b/spmd/compiler/distribute.py
@@ -467,7 +467,7 @@ class _SPMD:
                     schemas.append(shard_schema)
 
         parallelized_gm, output_specs = _convert_to_distributed(
-            training_phase, gm, inps, schemas
+            training_phase, gm, inps, schemas, _allow_partial=False,
         )
         self._known_specs_by_node_name.update(output_specs)
 
--- a/spmd/compiler/distribute.py
+++ b/spmd/compiler/distribute.py
@@ -467,7 +467,11 @@ class _SPMD:
                     schemas.append(shard_schema)
 
         parallelized_gm, output_specs = _convert_to_distributed(
-            training_phase, gm, inps, schemas, _allow_partial=False,
+            training_phase,
+            gm,
+            inps,
+            schemas,
+            _allow_partial=False,
         )
         self._known_specs_by_node_name.update(output_specs)
 
--- a/spmd/compiler/distribute.py
+++ b/spmd/compiler/distribute.py
@@ -170,7 +170,7 @@ def _get_dtensor_dispatch_graph(
 
 def _build_dummy_add_graph(
     dt: DTensor, node_to_obj: Dict[fx.Node, object]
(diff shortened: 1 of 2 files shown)


spmd/compiler/distribute.py
```diff
@@ -165,22 +165,17 @@ def remap_arg(arg: object) -> object:
     def unwrap_local(e: object) -> object:
         return e._local_tensor if isinstance(e, DTensor) else e
 
-    gm = make_fx(dispatch)(local_target_args)
-    for nd in gm.graph.nodes:
-        if nd.name == "_tensor_constant0":
-            print(
-                f"WARNING: found constant while SPMD expansion for {node}. Graph {gm}. This is not supported."
-            )
-    return gm
```
aazzolini, line 174: I realized this warning is wrong, i think a constnat is always expected (It's the ProcessGroup i think).

spmd/compiler/distribute.py
```diff
@@ -206,38 +201,44 @@ def dummy_add(grad: torch.Tensor, zero: torch.Tensor) -> torch.Tensor:
...
             continue
-        obj = node_to_obj[argument]
 
-        has_partial = _is_partial_dtensor(obj)
+        obj = node_to_obj[argument]
 
-        if not has_partial:
+        if not _is_partial_dtensor(obj):
+            new_args.append(argument)
             continue
```
aazzolini, line 223: This is the main fix . There's 2 issues

1) has_partial was set to True and then back to False if a Sharded() output appeared after a Partial_() output (I think this bug was added in the cleanup).

2) new_args needs to be appended with the output even if it's not partial. Thsi was being skipped -- and this is what caused less outputs than expected.

richqyz (approved):
Thanks for fixing this!

spmd/compiler/distribute.py
anj-s: "new_args needs to be appended with the output even if it's not partial." - Can you explain this a bit more? Not sure I got this issue.

anj-s (approved):

spmd/compiler/distribute.py
aazzolini: In this loop we are iterating through all the outputs.
We need to check, for each output, if they are Partial or not. 
- If they are Partial, we add the all reduce, and then append the output from all_reduce to the output list
- If they are not Partial, we should just append them to the new output list. But we were not doing that, so here's the bug.

PR #666, opened 2022-12-01
graph_optimization: update before validate

aazzolini:
this behavior was not tested;
for some reason, when running Rich's unit tests, the backward graphs are already present -- most likely because we run forward a second time?


fegin:
I'll investigate this issue. Accept to unblock.

fegin (approved):

PR #671, opened 2022-12-02
clone node offset, fix node issue from deepcopy, add autodepth finder for comm node

lessw2020:
This PR is to update two aspects:
1 - rollback the changes that assumed clone nodes would not be present 
2 - fix a tricky bug from using deepcopy for the fe node list. 
3 - adds automatic offset finder for the comm node to replace the hardcoded offset within the FusionElement node list. 

Note - unit test is pending, but needs some further integration work for the new api/distributed breakout. Wanted to get the above fixes in directly.  
This file with changes is not live/hooked in yet to be clear.

### DIFF
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -86,6 +86,8 @@ class GraphInfo:
     first: Optional[fx.Node] = None
     # last node in graph (tail / output)
     output: Optional[fx.Node] = None
+    # offset to comm node within a FusionElement sequence
+    fe_offset_to_comm_node: Optional[int] = None
 
     def update_info(self, gm: fx.GraphModule) -> None:
         """Get the len, input and output nodes"""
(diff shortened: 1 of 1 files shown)


spmd/compiler/fusion.py
```diff
@@ -178,6 +177,8 @@ def _scan_graph_for_fusion_elements(
     fe_size = len(fe_sequence) - 1
     index = 0
     curr_node_list = []
+    # depth to reach comm_node, adjust based on clone presence
+    offset_to_comm_node = 3
```
fegin, line 181: A general question: is this always the fixed number? Seems to be brittle. Any way to figure it out dynamically?

lessw2020: It's normally fixed, but agree it's better to just auto-adapt and not have a brittle assumption.
I've added a run once autodepth finder that seeks and stores the offset, so that now code will automatically adjust for any future changes in fe_sequence. 
Thanks for the feedback!

PR #674, opened 2022-12-03
Add layernorm and layernorm_backward sharding rules

aazzolini:


test/spmd/tensor/test_experimental_ops.py
```diff
@@ -0,0 +1,60 @@
...
+from torch.distributed._tensor import DeviceMesh, distribute_tensor
+from torch.distributed._tensor.placement_types import (
+    Replicate,
+    Shard,
+    _Partial,
+)
+from spmd.tensor.experimental_ops import *  # noqa: F401 F403
+
+
+class TraceModuleTest(DTensorTestBase):
```
anj-s, line 16: Thanks for adding a new file

test/spmd/tensor/test_experimental_ops.py
```diff
@@ -0,0 +1,60 @@
...
+    @property
+    def world_size(self):
+        return 2
+
+    def test_native_layernorm(self):
+        self.init_pg(backend="nccl" if DEVICE_TYPE == "cuda" else "gloo")
```
anj-s, line 22: should we assert that it is only nccl cause we don't support gloo right now?

anj-s (approved):

test/spmd/tensor/test_experimental_ops.py
wanchaol, line 22: setting up pg is actually what `@with_comms` is doing, maybe consider using that?

spmd/tensor/experimental_ops.py
```diff
@@ -16,33 +17,76 @@
...
+    assert isinstance(weight, DTensorSpec)
+    assert isinstance(bias, DTensorSpec)
+    assert isinstance(grad_input_mask, (list, tuple))
+    assert all(isinstance(s, Replicate) for s in weight.placements)
+    assert all(isinstance(s, Replicate) for s in bias.placements)
+    # ensure sharding on dim 0, which will trigger the "Partial" output on weight and bias grads
+    assert any(
+        isinstance(s, Shard) and s.dim == 0 for s in grad.placements
+    ), f"Got {grad.placements}"
+    weight_grad = DTensorSpec(
```
wanchaol, line 67: could we add some comment about why grad sharding on dim 0 will produce partial gradients on weight and bias grad?

wanchaol (approved):
lgtm, have some questions about weight/bias partial sharding, stamp to unblock

PR #675, opened 2022-12-05
Adding a issue and PR template

HamidShojanazeri:
This PR provides a template for the issues, lets iterate on this to come up with all the required info we need to tackle an issue.


kwen2501 (approved):
Let's try it out.

PR #678, opened 2022-12-06
[SPMD] Populate primals <-> param <-> grad mapping

fegin:
**What does this PR do?**
1. Populate the mapping between primal<->pram<-> grad so that the later fusion algorithm can utilize the information.
2. Rewrite graph_utils.py and part of the fusion.py to extract the common part.
3. Introduce `get_comm_block_node()` to fetch all the nodes belong to a communication node. This replace the existing logic in `fusion.py`.

### DIFF
--- a/spmd/compiler/distribute.py
+++ b/spmd/compiler/distribute.py
@@ -28,7 +28,7 @@ from torch.utils._pytree import tree_flatten, tree_map, tree_unflatten
 
 from .aot_function_patch import patched_aot_function
 from .distributed_graph import DistributedGraph
-from .graph_utils import CommType, get_comm_block_ops, OP
+from .graph_utils import CommType, get_comm_block_nodes, OP
 from .log_utils import rank0_info
 
 # patch aot_function so that we can pass the full (non-sharded) input to capture the graph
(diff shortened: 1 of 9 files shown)


spmd/compiler/graph_utils.py
```diff
@@ -22,52 +19,96 @@ class OP(str, Enum):
...
+        node = node.prev
+
+    comm_idx = len(node_list) - comm_idx - 1
+    node_list.reverse()
+
+    return comm_idx, node_list
+
+
+def get_node_tensor_metadata(
+    node: fx.Node, must_present: bool = True
```
lessw2020, line 85: minor nit - I think "required" or "is_required" is cleaner for a param name vs "must_present".  
I initially interpreted 'must_present" as meaning "must debug show", vs what you really mean is "must_be_present".

lessw2020 (approved):
LGTM!
minor nit comment on param name, but this is a nice improvement overall to the code.

spmd/compiler/distribute.py
```diff
@@ -421,25 +422,68 @@ def __init__(
...
+            len(self._dist_graph.primal_name_to_node[0]),
+            nparams,
+        )
+
+    def _map_grad_to_param(self, gm: fx.GraphModule) -> None:
+        # This assumes that the order of gradients are in the same order
+        # as the parameters in the input. The last argument of the output
+        # is None.
+        # TODO: this is very brittle, verify how we can do better than this.
+        self._dist_graph.grad_to_primal.append({})
```
awgu, line 457: @fegin For my conceptual understanding, why is `grad_to_primal` a `List[Dict[fx.Node, fx.Node]]` and not just `Dict[fx.Node, fx.Node]`? The current implementation seems to assume that the list always has length 1 (e.g. L472).

PR #679, opened 2022-12-06
[SPMD] Add graph optimization selector

fegin:
**What does this PR do?**
This PR add the graph optimization selection interface to allow users to decide which optimization(s) to run.


lessw2020 (approved):
looks good!  
Longer term I think we'll want to have a default optimization already supplied in init for the user, but for now this is ideal so we can test all the different options via SPMD params.
Thanks for adding this feature.

PR #683, opened 2022-12-07
[SPMD] Fix debug print which expects `i` to be valid

anj-s:


spmd/compiler/fusion.py
```diff
@@ -736,9 +736,9 @@ def run_overlap_communication(gm: fx.GraphModule) -> None:
     for i, item in enumerate(fe_list[1:]):  # type: ignore
```
fegin, line 736: I think adding `i = -1` before this line is more accurate.

spmd/compiler/fusion.py
```diff
@@ -736,9 +736,9 @@ def run_overlap_communication(gm: fx.GraphModule) -> None:
...
         moved_nodes = _move_comm_section(graph_info, gm, item)  # type: ignore
 
-    _debug(
-        f"\nOptimization stats:\nOverlap communication pass has moved -* {i+1} *- communication calls\n"
-    )
+        _debug(
```
fegin, line 739: This debug message is the summary of the `_move_comm_section` so the original indentation is correct. We should initialize `i` before the for loop.

spmd/compiler/fusion.py
anj-s: makes sense. We tend not to refer to loop vars outside the loop and hence assumed we were debugging something else.

Also make more sense for us to assert that we should have optimized "something" at this point so we should not asserting on i > -1?

PR #684, opened 2022-12-07
[SPMD] Print out the graph if applying optimization

fegin:
Print out the graph before applying optimization and after applying optimization.

### DIFF
--- a/spmd/compiler/api.py
+++ b/spmd/compiler/api.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Dict, Optional, Sequence, Tuple
 
 import torch.distributed as dist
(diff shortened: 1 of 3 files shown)


lessw2020 (approved):
looks good!
I personally like to use print_tabular() so maybe we can set something to toggle the type of print displayed in the future.

spmd/compiler/api.py
```diff
@@ -19,6 +24,7 @@ def __init__(
         optimize_first_iter: bool = False,
         optimizations: Sequence[GraphOptimization] = tuple(),
         map_param_and_grad: bool = True,
+        print_graph: bool = False,
```
wconstab, line 27: instead of a flag, consider using logging levels.

log.info(my mostly useful logs)
log.debug(my graph)

then just have users configure the log level

fegin: This idea is better. Since @anj-s  is preparing another PR for logging the graph. I'll just let she do so and land this PR for now so that we can debug.

PR #685, opened 2022-12-07
[SPMD] Ensure optimized is set to True after optimizations are applied

fegin:
Right now the `optimized` flag is not appropriately set. This PR fixes the issue.


lessw2020 (approved):
lgtm!

PR #687, opened 2022-12-08
Unify argument parsing across tests and examples

### LINKED ISSUE #384: Unify argument parsing across tests and examples to avoid code duplication
Subtask of https://github.com/pytorch/PiPPy/issues/299

acforvs:
Closes #384

### DIFF
--- a/pippy/utils.py
+++ b/pippy/utils.py
@@ -2,6 +2,7 @@
 import os
 import socket
 import logging
+import argparse
 
 # Pinning process to a separate GPU if not yet done by launch script
 # Notes:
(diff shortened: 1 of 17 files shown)


pippy/utils.py
```diff
@@ -224,3 +225,53 @@ def pp_group_barrier():
...
+        "--master_port",
+        type=str,
+        default=os.getenv("MASTER_PORT", default_port),
+    )
+    if default_schedule is not None:
+        parser.add_argument(
+            "-s",
+            "--schedule",
+            type=str,
+            default=list(default_schedule)[0],
```
kwen2501, line 262: I think we can use `FillDrain` as the default schedule.

pippy/utils.py
```diff
@@ -224,3 +225,53 @@ def pp_group_barrier():
...
+    parser.add_argument(
+        "--master_addr",
+        type=str,
+        default=os.getenv("MASTER_ADDR", default_addr),
+    )
+    parser.add_argument(
+        "--master_port",
+        type=str,
+        default=os.getenv("MASTER_PORT", default_port),
+    )
```
kwen2501, line 256: I am thinking if we should leave these four args in the user program, for two reasons:
1. These args may vary from program to program, so it may be hard to unify;
2. These args may come from environment variables, so it may be good for them to be in top level so that users know which env vars correspond to them.

kwen2501 (approved):
Looks good! Thanks much for the contribution! (And sorry about the delay.)
I just added two minor comments. Let me know what you think :)

PR #689, opened 2022-12-09
Fix distributed.py when op output is a list

aazzolini:
When the output  of an op is a list we need to hack up to "recompose" multiple tensor outputs back to the original list. (because later this list will be getitem'd into by the subsequent operators).

I believe later on we should redo this in a better way, but for now this should fix our layer_norm_backward that returns a tuple of 3 tensors.

### DIFF
--- a/spmd/compiler/distribute.py
+++ b/spmd/compiler/distribute.py
@@ -278,7 +278,7 @@ def _convert_output(
 def _rebuild_graph(
     gm: fx.GraphModule,
     node_replacements: Dict[torch.fx.Node, torch.fx.GraphModule],
-) -> Dict[torch.fx.Node, torch.fx.Node]:
+) -> None:
     # replace nodes in local traced graph with DTensor's dispatch graph
     new_nodes: Dict[torch.fx.Node, torch.fx.Node] = {}
     for node in gm.graph.nodes:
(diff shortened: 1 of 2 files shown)


spmd/compiler/distribute.py
```diff
@@ -307,7 +308,34 @@ def _rebuild_graph(
...
+                    # 2. multiple outputs resulting from getitem of all elements of tuple
+                    if len(outputs) == 1:
+                        # for single output, we replace the node with the single node
+                        output = outputs[0]
+                    else:
+                        # for multiple outputs, we check that these outputs correspond
+                        # to all elements of a tuple. In that case, we replace
+                        # uses of the output directly with the original tuple
+                        source = None
+                        for i, o in enumerate(outputs):
```
lessw2020, line 323: Very minor nit: I think 'out' instead of 'o' is more informative and avoids o/0/O misreading, esp since there's o.args[0] multiple times.

lessw2020 (approved):
Looks good to me! 
very minor nit on o use but not consequential.

spmd/compiler/distribute.py
```diff
@@ -316,6 +344,7 @@ def _rebuild_graph(
     gm.graph.lint()
     gm.graph.eliminate_dead_code()
     gm.recompile()
+    return new_nodes
```
anj-s, line 347: Are we expecting a return value from this function?

PR #690, opened 2022-12-09
update fusion to run with default params, buffer zero shape and size …

lessw2020:
This PR updates default params for comm_fusion in the new pipeline style, and adjusts the buffer tensors to not assume 2D shape (support ndim) for fusion.

### DIFF
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -212,12 +212,12 @@ def _copy_fe_to_buffer(
     if gi.tracing_buffer is None:
         buffer = torch.empty(buffer_size)
         gi.tracing_buffer = buffer
-    elif gi.tracing_buffer:
+    else:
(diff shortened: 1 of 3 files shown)


spmd/compiler/fusion.py
```diff
@@ -575,7 +575,7 @@ def run_fuse_communication(gm: fx.GraphModule) -> None:
 
     # simple fusion policy where int = num buckets to fuse...start with 2,
     # meaning every 2 comms are fused into 1
-    fusion_policy: int = 2
+    assert fusion_policy > 1, f"fusion policy requires > 1 for actual fusion. "
```
fegin, line 578: Nice

lessw2020:
pyre fixes are coming in 20 mins but opening for review so we can push at 5PM.

fegin (approved):
LGTM

PR #693, opened 2022-12-10
fix fstring lint

lessw2020:
not sure how this leaked in but showed up after I had merged and deleted the branch.
This, this is a one line PR so can push tau.


wz337 (approved):
LGTM!

PR #694, opened 2022-12-10
Memory savings during SPMD expansion

aazzolini:


richqyz (approved):
Thanks a lot for the effort!

PR #695, opened 2022-12-10
[SPMD] Introduction fusion with concat

fegin:
**What does this PR do?**
This PR introduce the fusion implementation that use concat to aggregate gradient without a fixed/pre-allocated buffer. The implementation does not rely one `make_fx()` but simply use `graph.call_function()` to create nodes on-the-fly.

This PR also refactor the existing unittests.


test/spmd/compiler/test_graph_optimizations.py
```diff
@@ -67,142 +69,51 @@ def forward(
...
-            )
-
-            spmd_overlap = SPMD(
-                base_model,
+        all_spmd = []
+        optimizations = [
+            [],
+            [GraphOptimization("overlap_communication")],
+            [GraphOptimization("fuse_communication")],
+            [GraphOptimization("fuse_communication_cat")],
```
lessw2020, line 78: Nice - thanks for updating the unit test for all of these.

lessw2020 (approved):
Looks good! Let's get it in and run for some numbers.

lessw2020:
went ahead and merged so we can push into FBCode for some perf numbers.

PR #697, opened 2022-12-13
Ring fusion - adds overlapped comm fusion for SPMD using a ring buffer

lessw2020:
This PR adds overlapped fusion using a ring buffer.
It has two params:
a - fusion_policy which is the total number of comm calls to fuse (i.e. 60 = fuse 60 calls into 1)
b - buffer_policy, which controls how many unique buffers to have in the ring.  Buffers are handed out in a cycle
to each respective fusion operation. 
The reason for the cycling is to avoid overwriting the buffer - we do not mutex the buffer, and comms are asynch.
Thus we can have a situation where a new fusion starts writing into an existing buffer before the outputs from the previous call are copied out.

### DIFF
--- a/spmd/compiler/fusion.py
+++ b/spmd/compiler/fusion.py
@@ -693,9 +693,8 @@ def run_fuse_communication_ring(
     )
 
     graph_info.num_starting_fe = len(fe_list)  # type: ignore
-    _debug(f"688, len of fe_list = {len(fe_list)}\n")
+    _debug(f"len of fe_list = {len(fe_list)}\n")
 
-    # fe_list = fe_list[::-1]
     graph_info.fe_list = fe_list
 
     # determine peak memory using fusion policy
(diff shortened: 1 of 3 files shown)


spmd/compiler/fusion.py
```diff
@@ -72,10 +72,13 @@ class GraphInfo:
     fe_list: List[FusionElement] = field(default_factory=lambda: [])
     # max memory needed for fusion buffer
     peak_memory_required: int = 0
-    # node housing global buffer for fusion comms
-    global_buffer_node: Optional[fx.Node] = None
+    # list housing global buffers for fusion comms
+    _ring_buffer: Optional[List[fx.Node]] = None
     # size of the global buffer
     global_buffer_size: int = 0
+    _ring_size: int = 0
```
anj-s, line 79: nit: I would call this num_ring_buffs perhaps?

spmd/compiler/graph_optimization.py
```diff
@@ -166,18 +166,20 @@ def apply(
         return self
 
     @graph_optimization_pass()
-    def fuse_communication(
+    def fuse_communication_ring(
         self,
         bucketing_strategy: BucketingStrategy = BucketingStrategy.FIXED,
         scheduling_policy: SchedulingPolicy = SchedulingPolicy.FCFS,
-        fusion_policy: int = 10,
+        # fusion_policy: int = 2,
```
anj-s, line 173: do we need to uncomment this?

lessw2020: yes - I had it setup when running so I only needed to mod fusion.py for debugging ease, but we want it to be 'pass through' from SPMD.

spmd/compiler/fusion.py
lessw2020: good point - I've updated to _ring_num_buffers so that it's clear the variable here is the number of buffers in the ring.  (num_ring_buffs could allude to having multiple ring buffers).

anj-s (approved):

spmd/compiler/fusion.py
```diff
@@ -92,6 +95,38 @@ class GraphInfo:
...
+        self,
+    ) -> fx.Node:
+        """get the next buffer node in the ring"""
+        buffer_node = self._ring_buffer[self._ring_index]  # type: ignore
+        self._current_ring_index = self._ring_index
+        assert (
+            buffer_node is not None
+        ), f"failed to get ring buffer for index {self._ring_index}\n"
+        self._ring_index += 1
+        if self._ring_index >= self._ring_size:
```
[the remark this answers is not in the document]
lessw2020: Generically right now we don't, unless we expand the ring buffer to fully match the total count of fusion calls (i.e. 1:1).   
Per Chien-Chin, it seems the fix would be if we move the wait nodes directly up against the buffer usage to enforce a mutex type setup as the waits would then block until the call is complete. 
Per our discussion, holding off on that until we see if ring buffer will be used and then can update with this if so.

PR #701, opened 2022-12-15
fix pg_planner.py and its test

wz337:
After this PR(https://github.com/pytorch/pytorch/pull/90212) landed to Pytorch master, ProcessGroupAwareSavePlanner and DefaultSavePlanner has an overlapping variable (self.original_state_dict), which would result in the bytes value of the state_dict not being loaded correctly. 

This PR fixes the issue above and re-instate the test for pg_planner.


fduwjj (approved):
LGTM

PR #702, opened 2022-12-15
fix dt_planner and its test

wz337:
After this PR(https://github.com/pytorch/pytorch/pull/90212) landed to Pytorch master, the class variable is defined in DefaultSavePlanner. Therefore, we will need to pass the correct value to super().__init__().

This PR fixes the issue above and re-instate the test for dt_planner.


fduwjj (approved):
LGTM

PR #705, opened 2022-12-16
Fix device buffer

lessw2020:
This PR: 
1 - moves the buffer creation from using "rank" to the universal "cuda" to match DTensor and fix multi-node support. 
It also makes a naming synch between fusion and graph_optimization to be consistent with param ring_num_buffers for the total buffer count.
2 - unrelated but proactive fix - the _get_tracer() call in CommTensor no longer takes an obj.  Have thus updated param signature to match in distribute.py.

### DIFF
--- a/spmd/compiler/distribute.py
+++ b/spmd/compiler/distribute.py
@@ -94,7 +94,7 @@ def _get_dtensor_dispatch_graph(
     def remap_arg(arg: object) -> object:
         if isinstance(arg, torch.fx.Node):
             obj = node_to_obj[arg]
-            if _get_tracer(obj):
+            if _get_tracer():
                 # This is a shared arg, already has a tracer from previous
                 # tracing. Delete the tracer.
                 del cast(Dict[object, object], obj.__dict__)[proxy_slot]
@@ -435,7 +435,7 @@ def _convert_to_distributed(
                     # TODO(anj): we need this for getitem but can we be more generic?
                     if isinstance(obj, tuple):
                         return obj
-                    if _get_tracer(obj):
+                    if _get_tracer():
                         # This is a shared arg, already has a tracer from previous
                         # tracing. Delete the tracer.
                         del cast(Dict[object, object], obj.__dict__)[proxy_slot]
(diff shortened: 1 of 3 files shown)


spmd/compiler/distribute.py
```diff
@@ -435,7 +435,7 @@ def remap_arg(arg: object) -> object:
                     # TODO(anj): we need this for getitem but can we be more generic?
                     if isinstance(obj, tuple):
                         return obj
-                    if _get_tracer(obj):
+                    if _get_tracer():
```
anj-s, line 438: don't you need obj? why delete it?

spmd/compiler/fusion.py
```diff
@@ -171,29 +170,23 @@ def _create_fusion_buffers(
...
     new_buffer_node = None
     with gm.graph.inserting_before(insert_before_node):
         for i in range(ring_size):
             new_buffer_node = gm.graph.create_node(
                 OP.CALL_FUNCTION,
                 target=torch.empty,
-                # TODO - need device from DTensor to put buffer on gpu
                 args=(buffer_size,),
-                kwargs={"device": rank_device},
+                kwargs={"device": "cuda"},
```
anj-s, line 189: does this assume a set_device(rank) call? if so can we make a comment regarding that?

anj-s (approved):

spmd/compiler/fusion.py
lessw2020: have added a comment for this.  
From what I could see, DTensor uses the same exact concept of referencing "cuda", and per the docs here "cuda" is auto-expanded to "cuda:x" where X is the device id, meaning DTensor should fail earlier if this is not properly set and we can thus rely on it. 
 
"e.g., a [torch.Tensor](https://pytorch.org/docs/stable/tensors.html#torch.Tensor) constructed with device 'cuda' is equivalent to 'cuda:X' where X is the result of [torch.cuda.current_device()](https://pytorch.org/docs/stable/generated/torch.cuda.current_device.html#torch.cuda.current_device)."

spmd/compiler/distribute.py
lessw2020: ah, CommTensor has been updated so this api call no longer uses the obj to determine the slot. 
Rather it now uses get_innermost_proxy_mode() and returns None or mode.tracer. 
(link: https://github.com/pytorch/pytorch/commit/54563e6288efda566be3653a547c8be2781cc547)

PR #706, opened 2022-12-20
[SPMD] Add support for logging configs

anj-s:
1. Enable config based logging where we can specify the format for different modules. This borrows from torchdynamo's logging settings.
2. Cleaning up smaller nits.

Follow ups:
1. Delete the earlier rank0.. logging config
2. Create a new logging object for each of the modules or use a cache. I have a feeling we might be rewriting logging object given they are global values.


spmd/compiler/distribute.py
```diff
@@ -210,6 +211,7 @@ def _convert_output(
     gm: fx.GraphModule,
     node: fx.Node,
     node_to_obj: Dict[fx.Node, object],
+    logger: Optional[logging.Logger] = None,
```
wanchaol, line 214: why we need to pass the logger into the function? it's already a global variable

spmd/compiler/distribute.py
```diff
@@ -1,45 +1,46 @@
...
+
+from torch._subclasses.fake_tensor import FakeTensorMode
 
 # patch aot_function so that we can pass the full (non-sharded) input to capture the graph
 # pyre-fixme
 torch._functorch.aot_autograd.aot_function = patched_aot_function
 
-
-logger: logging.Logger = logging.getLogger(__name__)
+logger: Optional[logging.Logger] = None
```
wanchaol, line 43: can we directly get the logger here instead of getting the logger inside each function?

spmd/compiler/distribute.py
```diff
@@ -384,17 +386,19 @@ def _convert_to_distributed(
         - transformed graph module
         - map from output name to DTensorSpec
     """
+    global logger
+    logger = get_logger("spmd_exp")  # type: ignore
```
wanchaol, line 390: nit: maybe get the logger in the global scope directly

spmd/compiler/distribute.py
```diff
@@ -612,6 +613,9 @@ def distribute(
     *args: Tuple[object],
     **kwargs: Dict[str, object],
 ) -> nn.Module:
+    # Initialize logger
+    logger = get_logger("spmd_exp")
```
wanchaol, line 617: ditto

spmd/compiler/fusion.py
```diff
@@ -93,6 +89,8 @@ class GraphInfo:
     actual_grad_index_mapping: Dict[fx.Node, int] = field(
         default_factory=lambda: {}
     )
+    global logger
+    logger = get_logger("graph_opt")  # type: ignore
```
wanchaol, line 93: I think we should probably move this out of the GraphInfo data class? logging probably should not correlate with custom data structure

spmd/compiler/fusion.py
```diff
@@ -802,29 +775,23 @@ def run_overlap_communication(gm: fx.GraphModule) -> None:
         graph_info, gm, comm_type=CommType.ALLREDUCE
     )  # type:ignore[arg-type]
 
-    _debug(f"length of fe_list {len(fe_list)}")  # type: ignore
+    logger.debug(f"length of fe_list {len(fe_list)}")  # type: ignore
```
wanchaol, line 778: any reason to use debug instead of info here? do we have a rule when to use which?

spmd/compiler/log_utils.py
```diff
@@ -17,3 +20,75 @@ def rank0_info(logger: logging.Logger, *args: Any, **kwargs: Any) -> None:
 def rank0_warning(logger: logging.Logger, *args: Any, **kwargs: Any) -> None:
```
wanchaol, line 20: nit: delete the warning too?

wanchaol (approved):
lgtm, a couple of suggestions inlined.

spmd/compiler/distribute.py
anj-s: We can't get the logger in global scope because we want torch.distributed to be initialized with PGs instantiated. I have moved printing on a single rank to the logger.

spmd/compiler/log_utils.py
anj-s: This is a followup task since I want to make sure this current logger logic works. Backup for devs in case it does not.

spmd/compiler/distribute.py
kumpera, line 617: This is missing `global logger`, nothing uses the local variable.

kumpera (changes requested):

anj-s:
Lint and test errors are related to checkpoint/*_pg_planner.

spmd/compiler/fusion.py
anj-s: This is useful most likely for debugging hence the change.

spmd/compiler/fusion.py
anj-s: This is the part that I don't like about this PR but for now it allows us to configure what logging we want. Answered above as to why initializing a global python object won't work.

PR #709, opened 2022-12-20
Remove requirement for loss spec

kwen2501:
Previously, we require users to provide a one-hot map to specify which output value is the loss.
For example, for T5:
```
    output_loss_value_spec = {
        "loss": True,
        "logits": False,
        "encoder_last_hidden_state": False,
    }
```
This requires users to have a good understanding of the model.

With this PR, we try to free users from such specification. We do so by automatically searching for "loss" in the output dict. (This is the most prevalent format for Hugging Face models.) 

Previously, `output_loss_value_spec=None` has been used to indicate the inference mode. Now, to do the same, users can set the model to evaluate mode before passing it to PiPPy. That is,
```
model.eval()
pipe = Pipe.from_tracing(model, ...)
```

### DIFF
--- a/examples/hf/t5/pippy_t5.py
+++ b/examples/hf/t5/pippy_t5.py
@@ -165,18 +165,14 @@ def transform_into_pipeline(
     chunks: int = 1,
     all_worker_ranks: List[int] = None,
 ):
-    output_loss_value_spec = {
-        "loss": True,
-        "logits": False,
-        "encoder_last_hidden_state": False,
-    }
+    if not args.train:
+        model.eval()
 
     t5_pipe = Pipe.from_tracing(
         model,
         multi_use_param_spec,
         tracer=PiPPyHFTracer(),
         concrete_args=concrete_args,
-        output_loss_value_spec=output_loss_value_spec if args.train else None,
         split_policy=split_policy,
     )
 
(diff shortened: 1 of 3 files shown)


HamidShojanazeri:
Thanks very much @kwen2501 , this is a great step to simplify the API for user! I could successfully test it.

HamidShojanazeri:
As for next steps, It would be great if we use the following opportunities to add further simplifications down the road as well. 

a. making `MULTI_USE_PARAM_CONFIG = MultiUseParameterConfig.REPLICATE if args.replicate else MultiUseParameterConfig.TRANSMIT` as default setting

b. making concrete args to be handled transparently from user by taking the required args in the from_tracing init. 
`sig = inspect.signature(model.forward)
    concrete_args = {p.name: p.default for p in sig.parameters.values() if p.name not in input_names}`

c. automatically infer the "output_chunk_spec" as in this case user needs to understand the what are all the available outputs from model, this requires going back to model src code. We could bring back all the available outputs and let users choose the required one from the output dict. (not sure if there is any downside to this, would appreciate your thoughts)

`output_chunk_spec = {"logits": TensorChunkSpec(0),"encoder_last_hidden_state": TensorChunkSpec(0)}`

PR #710, opened 2022-12-26
fix barrier params

jiqing-feng:
@HamidShojanazeri Hi, Hamid. I think `torch.distributed.barrier` does not need parameters, besides, the parameter `args.pp_group` is not defined. I remove this parameter and the script runs well. Could you help me review this changes? Thanks!

HF_inference_errors

### DIFF
--- a/examples/inference/HF_inference.py
+++ b/examples/inference/HF_inference.py
@@ -120,7 +120,7 @@ def run_master(pp_ranks, args):
    
     model_pipe.defer_stage_init(args.device)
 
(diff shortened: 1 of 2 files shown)


kwen2501 (approved):
Thanks for the contribution. Approving.
In the future, we can also use `pp_group_barrier()` in pippy.utils.
Example:
https://github.com/pytorch/tau/blob/main/examples/gspmd/pippy_gspmd.py#L74

PR #712, opened 2023-01-01
Fix hf_model_tests

kwen2501:
- Update skipped tests per upstream changes
- Skip numeric check that has been done in forward tests


test/hf_test.py
```diff
@@ -917,8 +924,14 @@ def test_case(self):
             ]:
                 self.skipTest("Need to support CLIPVisionModelWithProjection")
 
+            # TODO: support SwinBackbone
```
anj-s, line 927: nit: I would refactor this check into a function.

Also seems like we have to backbones to support. what is the missing feature?

anj-s (approved):

PR #713, opened 2023-01-02
Fix PiPPy + Dynamo CI

kwen2501:
Fixing CI error encountered after Dynamo upgrade/migration:
https://github.com/pytorch/tau/actions/runs/3816032630/jobs/6491411483

Solution:
updated the way to compute the reference value -- we now move it out of the compilation loop. 

More details: 
it seems `gm` in the compilation loop comes with some Fake components, hence we cannot directly pass `example_inputs` to `gm`.

Effect:
fixed "PiPPy tests / programming_model_tests"

### DIFF
--- a/examples/TorchDynamo/pippy_dynamo.py
+++ b/examples/TorchDynamo/pippy_dynamo.py
@@ -10,6 +10,7 @@ import torch.fx
 # TorchDynamo is moved into PyTorch for PyTorch > 1.13
 # One can use TorchDynamo without installing it separately by:
 import torch._dynamo as dynamo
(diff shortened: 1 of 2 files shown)


examples/TorchDynamo/pippy_dynamo.py
```diff
@@ -10,6 +10,7 @@
 # TorchDynamo is moved into PyTorch for PyTorch > 1.13
 # One can use TorchDynamo without installing it separately by:
 # import torch._dynamo as dynamo
```
anj-s, line 12: can we use this instead?

anj-s (approved):

PR #714, opened 2023-01-10
fixing the example if conditions

HamidShojanazeri:
Fixing the conditions to pick the right model in the inference example


kwen2501 (approved):
LGTM! Thanks for the fix!

PR #718, opened 2023-01-13
Example of capturing training step and lowering to inductor

aazzolini:


aazzolini:
Closing , wrong files.

PR #719, opened 2023-01-13
Example of capturing training step and lowering to inductor

aazzolini:


experimental/inductor.py
```diff
@@ -0,0 +1,218 @@
...
+
+def make_inductor_fn(fn, example_inputs):
+    """
+    Use inductor to compile the function fn. 
+    Uses make_functional_fx to trace it passing example_inputs.
+    """
+    fnfx, attr_vals, unused_input_idx = make_functional_fx(fn)(*example_inputs)
+
+    used_args = prepare_args(attr_vals, unused_input_idx, example_inputs)
+    compiled_f = compile_fx_inner(fnfx, used_args)
```
voznesenskym, line 124: why compile_fx_inner vs compile_fx?

experimental/inductor.py
```diff
@@ -0,0 +1,218 @@
...
+        #    - We collect all concrete getattr inputs into "attr_vals"
+        from collections import defaultdict
+        attr_vals_map = dict()
+        attr_nodes = defaultdict(lambda: [])
+
+        for p in k2.graph.nodes:
+            if p.op == 'get_attr':
+                val = getattr(k2, p.target)
+                attr_vals_map[p.target] = val
+                attr_nodes[p.target].append(p)
```
voznesenskym, line 57: if this attr_nodes is > 1 `p` for the target (say, `k`), would you not also have `k` vals? Does attr_vals_map need to map to a list too?

experimental/inductor.py
```diff
@@ -0,0 +1,218 @@
...
+
+        after_last_placeholder = None
```
voznesenskym, line 59: nit - weird name/logic? - this is the last arg + 1 right? logically, you are doing "insert before the first argument after the last placeholder" which is a long way of saying "insert after the last arg" - maybe clean this up to just insert after the last placeholder?

experimental/inductor.py
```diff
@@ -0,0 +1,218 @@
...
+        #      graph to inductor. Inductor expects only tensors as inputs
+        #    - This works since all the non-tensor inputs should be unused at this point,
+        #      since in principle we should have converted all getattrs into inputs.
+        #    - We keep the indices of unused inputs in "unused_inputs_idx"
+        unused_inputs = []
+        unused_inputs_idx = []
+
+        for i, p in enumerate(k2.graph.nodes):
+            if p.op != 'placeholder':
+                break
```
voznesenskym, line 86: why this constraint? if you have unused other ops, might as well prune no?

experimental/inductor.py
aazzolini: compile_fx calls aot_autograd. in this example I bypass aot_autograd , replacing it with make_functional_fx above.

experimental/inductor.py
aazzolini: right and I do call eliminate_dead_code() at the end but it's a bit more complicated because there's some blacklisting that needs to happen and in particular there is an inplace call to copy_ that needs to be kept even though there's no user to its output.

experimental/inductor.py
aazzolini: yeah i want to insert a new placeholder after the last placeholder, which means we'll be appending args to the input of the graph . i'll change to inserting_after.

experimental/inductor.py
aazzolini: hmm i think it would be the same value for all getattrs that point to the same parameter. specifically what this means is : 

x1 = getattr('myparam')
x2 = getattr('myparam')

this will map to self.myparam in both cases, so they both map to the same value, but i need to replace both nodes with the placeholder. Does that make sense?

aazzolini (commented):
Thanks @voznesenskym  for the quick review! 
I'd really appreciate suggetsions on how this could be integrated with dynamo? Would it be possible to somehow integrate the dynamo frontend with this without causing graph breaks due to backwards or optim?

voznesenskym (approved):

PR #720, opened 2023-01-14
[WIP] Capture whole fused graph, perform SPMD expansion, run with Inductor

aazzolini:
Not yet working.


experimental/inductor.py
```diff
@@ -0,0 +1,411 @@
...
+        torch.cuda.set_device(torch.distributed.get_rank() % 8)
+
+    print('================ main_simple =================')
+    main_simple()
+
+    print('================ main with SGD =================')
+    main(torch.optim.SGD)
+
+    print('================ main with Adam =================')
+    # sharding seems to be wrong -- seems to be doing a bunch of allreduces that are not needed
```
aazzolini, line 409: See outputs here: 
https://gist.github.com/aazzolini/4e7b8580fe0b5cebdefe0a7307a9a00f

Notice multiple all_reduce being inserted after the forward pass

aazzolini:
Expose an API `make_inductor_fn` to perform SPMD expansion and inductor lowering on a function possibly containing side-effects.
In the example, the API is used to compile and run a full `training_step` (fwd, bwd and optim.step).

### Compilation passes

For now, this bypasses dynamo because dynamo leads to a graph break when doing loss.backards() . The goal here is to capture the entire function in a single graph.

In order to properly fuse and lower kernels, a functionalization pass as well as specific decompositions are required by inductor. In order to do so , we need to do `make_fx(functionalize(function))`. However, when capturing a training_step, this doesn't work because `functionalize` complains when `loss.backwards()` is called from inside the function.

The workaround is to call make_fx once to "remove" the backwards call, and then call functionalize on the lowered graph. So we do: `make_fx(functionalize(make_fx(function))`. (This is a trick that was used in the past as part of `aot_autograd` implementation, so nothing new here).

Another issue with `functionalize` is that it doesn't allow for `getattrs` to be present in the graph -- all tensors used inside of `function` must be passed as explicit arguments. In order to do so, we perform a transformation `convert_getattrs_to_inputs` after the first `make_fx` -- removing all instances of "getattr" and making them input arguments to the function. We also collect the concrete values of self.attr so that we can pass them as inputs to the function afterwards.

After converting getattrs to inputs, we are still left with some non-tensor inputs to the graph. In our training_step example, we take as inputs a `nn.Module` and a `torch.Optimizer` instances. Indiuctor will complain if we pass these as input to our compiled function. Luckily, `convert_getattrs_to_inputs` already extracted all the parameters and buffers and optimizer states and exposed them as explicit inputs to the function, so we don't need the nn.Module and the torch.Optimizer instances to be passed to the compiled function. We remove these unused arguments `remove_unused_inputs` and keep track of the removed inputs so that we can keep track of them when calling the compiled function.

Finally, it turns out that `functionalize` itself ends up adding getattrs to the graph, but our SPMD expansion doesn't support that because it was originally meant to be used as a compiler for aot_autograd. So what we do is run another pass of `getattr` removal before the SPMD expansion.

So our compilation pipeline looks like this:

```
gm = make_fx(fn)
gm2, getattr_vals = convert_getattrs_to_inputs(gm)
gm3 = make_fx(functionalize(gm2))
gm4, getattr_vals2 = convert_getattrs_to_inputs(gm3)
gm5, unused_inputs_idx = remove_unused_inputs(gm4)
gm6 = convert_to_distributed(gm5)   # SPMD expansion
compiled_f = inductor.compile_fx_inner(gm6)
```

Where `compiled_f` is a inductor-compiled function.


### Example inputs

Care has to be taken on what example inputs are passed in to each compilation step.
- the first `make_fx` takes inputs exactly as in the initial function. In regards to SPMD, these inputs are supposed to be global inputs (not local shards of the inptus). So for example if we are passing a batch of training data with global batch_size=1024, our input.size(0) needs to be 1024, regardless of the number of workers.
- After we convert_getattrs_to_inputs, the example inputs need to be modified to also contain the converted getattrs as part of the inputs. So we need to pass initial_example_inputs + getattr_vals as example inputs.
- After we remove_unused_inputs, we must remove the unused example inputs as well. In our training_step, this means we're going to remove the nn.Module and the torch.Optimizer from the input list moving forward.
- Example inputs passed to `convert_to_distributed` need to be sharded, not global inputs, so before calling this pass, we need to obtain shards somehow -- currently we're creating a random tensor that has a shape equivalent to a shard. For example in our batch_size=1024, if we have 2 GPUs, our sharded_example_inputs will have suze(0) = 512.

### Sharding annotations
- SPMD expansion needs input tensors (including parameters and buffers) to be annotated with the seried sharding spec. In order to do so, we expose an API `mark_placements` to annotate tensors/parameters/etc with the desired placement. We also pass a `default_shard_schema` to `convert_to_distributed` which is used to obtain the DeviceMesh (we only support one device mesh for an SPMD expansion) and to obtain the default placements when an input tensor is not annotated.

### Issues
- The main issue currently is that we're not using FakeTensors for the initial pass -- so for example if we have a huge parameter that we would like to shard, right now this doesn't work, because 1) we need to instantiate the concrete parameter with full size (before sharding) so that the initial passes can happen; 2) `getattr_vals` will keep a reference to the full (not-sharded) parameter, so even if we shard the parameter afterwards, when we call `compiled_f`, it will still pass the initial parameter. We need a way to declare parameters as lazy and materialize them after sharding is decided. I'll investigate this next.

aazzolini:
Status: I'm currently unable to call loss.backward() on a FakeTensor that is marked as 'cuda'. It works for cpu FakeTensors, but then optimizer.step() won't work because passing capturable=True will check that the tensor type is cuda.

aazzolini:
Status: this seems to be working for the simple model in the example. The compiled function takes DTensors in the input, and produces DTensors in the output.

Nice example is that loss is returned as a "Partial" tensor so that it avoids doing all-reduce but at the same time informs the user that this local loss needs to be sumemd to get the full batch loss.

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+        k2 = k1(*args, **kwargs)
+        print_graph('first_fx', k2)
+
+        attr_vals = convert_getattrs_to_inputs(k2)
+        print_value('attr_vals', attr_vals)
+
+        # 3. Remove unused inputs
+        #    - This is needed in order to remove non-tensor inputs before passing this
+        #      graph to inductor. Inductor expects only tensors as inputs
+        #    - This works since all the non-tensor inputs should be unused at this point,
```
wanchaol, line 177: will there be any problems for non-tensor inputs? i.e. forward that contains both a tensor input and some non-tensor args

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+def warmup_optimizer(optim):
+    # unfortunatelly, in order to initialize the optimizer states
+    # we need to call optimizer.step(). In order to do that, we will initialize
+    # parameter gradients with zero tensors just so we can run a step.
+    for group in optim.param_groups:
+        for param in group['params']:
+            param.grad = torch.zeros_like(param)
+    # run an optimizer step just to initialize the optimizer states
+    # multiple issues here:
+    # 1. this will make the optimizer think we're at the 2nd iteration
```
wanchaol, line 247: if we do `make_inductor_fn` directly on the train_step, would we still need to warm up optimizer? because the train_step itself contains foward + backward + optim.step so it should be warmed up already during the first time we `make_fx` it?

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+            tensor_materializer = partial(
+                _materialize_dtensor,
+                default_shard_schema.mesh,
+                default_shard_schema.placements,
+            )
+        # In the following lines we perform utterly black magic in order to
+        # materialize our models. It special cases models and optimizers
+        # and assumes that all parameters contained in optimziers are model
+        # parameters.
+        tracking_materializer = TrackingMaterializer(_materialize_tensor)
```
wanchaol, line 379: should pass in `tensor_materializer` instead?

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+
+	
+T = TypeVar('T')
+
+
+def _materialize_dtensor(mesh, default, tensor, param=None):
+    if isinstance(tensor, DTensor):
+        return tensor
+    else:
+        # do we have a way to make this into a DTensor?
```
wanchaol, line 290: do you mean full tensor -> DTensor? maybe use the `distribute_tensor` API?

wanchaol (commented):
First pass reviewing although it's still WIP, looks a great direction! left some q about warm up optimizer specifically.

experimental/faketest.py
```diff
@@ -0,0 +1,86 @@
...
+
+DIST = True
+inductor.DIST = DIST
+
+
+# BUG: Returning mutated input generate 2 issues:
+# 1) functionalization doesn't properly return the same initial buffer but rather the new one its copied into
+# 2) inductor doesn't emit the buffer materialization for that input.
+
+# SURPRISING BEHAVIOR: Creating DeviceMesh in FakeMode fails because mesh and rank_coordinatees are stored as tensors
```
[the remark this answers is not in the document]
aazzolini: Yes let's make them numpy arrays, I think that would be best.

experimental/test_fake_backward.py
```diff
@@ -0,0 +1,16 @@
...
+from torch._subclasses.fake_tensor import FakeTensorMode
+
+def debug_fake_backwards():
+    torch.randn(1, device='cuda')
+    # torch.set_default_device('cuda')
+    with FakeTensorMode():
+        p = torch.randn(4, 2, requires_grad=True, device='cuda')
+        x = torch.randn(8, 4, device='cuda')
+        y = torch.mm(x, p).square().sum()
+        # y.backward()
```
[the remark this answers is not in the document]
aazzolini: Yes but for now we are able to work around it by running a concrete cuda operation first.

experimental/inductor.py
aazzolini: Yes this isn't battle proof yet. Right now I didn't flex the types of things you pass in.
Also, in general, since we're bypassing dynamo, there's no guarantees that non-tensor ops will be properly captured, as they would ended up being baked into the graph.
Ultimately it would be best if dynamo could produce a single graph without breaks. Right now it's breaking on .backwards()

experimental/inductor.py
aazzolini: The first time we make_fx it, it will be on fake mode. This is required because we're initially tracing with "global shapes", which would make our device OOM if we were not in fake mode.

Then, the second time we trace it, we still didn't do SPMD expansion, so even though we could, in theory, run the train_step with eager DTensor, I thought we would prefer to avoid that, since we may want to do SPMD expansion using more sophisticated propagation techniques in the future. But it would be an option for sure.

experimental/inductor.py
aazzolini: Right. Basically the logic here is as follows:
1.1) If we have a tag on this tensor to specify its sharding, _or_ if we have a default sharding strategy then we produce a DTensor.
1.2) Otherwise produce a Tensor (not DTensor)

As for what to fill the materialized tensor with:
  2.1) If this is a Fake tensor, then just make a Tensor [1.2] (or DTensor [1.1]) containing zeros. (User could further initialize this with torch.nn.init functions)
  2.2) If this was a real already a tensor and we are producing a DTensor [1.1], then use distribute_tensor API. Otherwise, just keep the original tensor as is.

experimental/inductor.py
aazzolini: oh wait yeah this is a bug !

aazzolini: after fixing this, turns out warmup_optimizer() i not working for adam with DTensor, because it creates a torch.zeros() internally to track training step. Should we have DTensor mode for these things?

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+        sgd = optim_cls(mc.parameters(), lr=0.01, capturable=True)
+
+        # 1.3. create fake input examples
+        fake_x = torch.ones(10, 4)
+        fake_y = torch.ones(10, 2)
+
+        compiled_f = make_inductor_fn(train_step, [mc, sgd, fake_x, fake_y], default_shard_schema=None)
+
+    # 2.3. create real input examples (should be DTensors)
+    x = torch.ones(10, 4)
```
anj-s, line 642: Should we have called distribute_tensor() on x and y before running compiled_f?

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+    p = torch.randn(2, 1, requires_grad=True, device='cuda')
+
+    if DIST:
+        x = distribute_tensor(x, mesh, [Shard(0)])
+        p = distribute_tensor(p, mesh, [Replicate()])
+
+    if DIST:
+        # make sure our parameter is the same across devices initially
+        # doing this after compile because during compilation we're running the training step a few times
+        # TODO: fix this. we should run compilation with fake tensors.
```
anj-s, line 507: I am guessing this is not true anymore since you got FakeTensors to work above.

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+        # torch.distributed.all_reduce(p)
+        pass
+
+    dprint(f'after compile, param norm = {(p*p).sum()}')
+
+    with torch.no_grad():
```
anj-s, line 513: Do we always need to do this in the no_grad() context?

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+        inp_spec = DTensorSpec(mesh, tensor._placements, tensor.shape, tensor.ndim)
+        return fn(inp_spec.local_shape, device=tensor.device)
+    elif default is not None:
+        inp_spec = DTensorSpec(mesh, default, tensor.shape, tensor.ndim)
+        return fn(inp_spec.local_shape, device=tensor.device)
+    raise NotImplementedError(f'Dont know how to get a shard of tensor {tensor}')
+
+
+
+def make_inductor_fn(fn, example_inputs, default_shard_schema=None):
```
anj-s, line 356: nice work!

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+    """
+    Use inductor to compile the function fn. 
+    Uses make_functional_fx to trace it passing example_inputs.
```
anj-s, line 359: are example_inputs always going to be Fake tensors?

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+    """
+    fnfx, attr_vals, unused_input_idx = make_functional_fx(fn)(*example_inputs)
+
+    used_args = prepare_args(attr_vals, unused_input_idx, example_inputs)
+
+    # materialize models + SPMD expansion. must be outside of fake mode
```
anj-s, line 365: I think we had talked about this and also done some initial prototyping about carrying out the SPMD expansion using FakeTensors. We should still do that IMO. Did you find any reason in your prototype for us to not pursue that?

experimental/inductor.py
```diff
@@ -0,0 +1,669 @@
...
+            used_args = [
+                get_example_shard(
+                    arg,
+                    schema.mesh,
+                    schema.placements
+                ) for arg, schema in zip(used_args, schemas)
+            ]
+
+            from spmd.compiler.distribute import _convert_to_distributed, TrainingPhase
+            fnfx, _, output_schemas = _convert_to_distributed(TrainingPhase.FORWARD, fnfx, used_args, schemas, _allow_partial=True)
```
anj-s, line 415: Ideally we should only materialize before we run the model with DTensor inputs.

aazzolini:
I'm now able to run a few iterations of  "Pure DDP" as well as "mixed FSDP / DDP".
This requires the latest updates in https://github.com/pytorch/pytorch/pull/92231.

experimental/inductor.py
aazzolini: @wanchaol I implemented a simple/hacky DTensorMode to deal with this.

experimental/inductor.py
aazzolini: Yes this is actually done, maybe this is an older version of the diff?

experimental/inductor.py
aazzolini: right, in the new version, i am calling 

mc.linear.reset_parameters()

which will act inplace on DTensors and will effectively initalize them in a sharded way. 
I had to implement a DTensor.uniform_ propagation rule in order to make this work -- such that it calls uniform_ on each shard of the parameter. This is nice, however it is not invariant to resharding (we shouold ideally have a version of random ops that takes a shard-invariant random seed, as previously discussed with Wanchao, if we want to preserve invariance of the op over resharding).

experimental/inductor.py
aazzolini: Yes otherwise it tries to compute gradients of an inplace operation, namely the copy_ operation that happens at parameter update time, and it will complain that you can't do backwards for an inplace operation. This is fine because at this point, backwards is already baked into the graph.

In case you want to do backwards of backwards i think you could use functorch beforehand.

experimental/inductor.py
aazzolini: not necessraily although i didn't test that it still works if example inputs are not fake.

experimental/inductor.py
aazzolini: No fundamental reason other than it currently doesn't work :)
If we can do SPMD expansion using FakeTensors that's great.
We need to make sure that DTensor can run in eager mode with FakeTensors as well so that we can propagate sharding from parameters to optimizer states at optimizer_warmup stage.

initialize_optimizers and materialize_model may need to be slightly modified to work on FakeMode+DTensor, but  nothing fundamentally differnt (we should rename materialize_model to distribute_model since it would now take a FakeTensor() as and produce a DTensor(FakeTensor()) for parameters).

experimental/inductor.py
aazzolini: Definitely , if we can get DTensor to work wiht FakeTensors this would be possible.

experimental/inductor.py
```diff
@@ -0,0 +1,773 @@
...
+
+    from torch.distributed._tensor import DTensor, ops, distribute_tensor
+    DTensor._op_to_rules['aten.lift_fresh_copy.default'] = ops.tensor_ops.default_prop_rule
+
+    # the rule below may have issues issues:
+    #  - likely doens't cover the case when a shape constant is passed
+    DTensor._op_to_rules['aten.zero.default'] = ops.tensor_ops.prop_create_like
+
+    @register_prop_rule("aten.zeros.default")
+    def prop_aten_zeros_default(op_schema: OpSchema) -> OutputSharding:
```
wanchaol, line 46: The DTensorMode looks great! Thanks for making this happen! I have one question with regards to those tensor constructors and it's interaction with the DTensorMode:

Most of the nn.Module parameters are initialized with those tensor constructors (i.e. torch.empty), if we implement those rules and use DTensorMode to initialize a model, does it mean that all of the parameters becoming replicated by default?

experimental/inductor.py
wconstab: I think there is a convention now where dynamo always fake-ifies inputs before passing them into the downstream compiler stack.  I'm assuming that since this function is analogous to the inductor backend to dynamo, the same should apply

aazzolini: Makes sense -- I think there's no reason to not fakify example_inputs. However some of the inputs are stateful (such as models and optimizers) -- for these it's not clear if we should force-fakify them. As a matter of fact, we are actually materializing those inside this function. 


Ideally we want the materialize part to happen _after_ this compilation is done. The only reason I'm doing it beforehand right now is because SPMD expansion still doesn't work in Fake Mode.

PR #725, opened 2023-01-18
[checkpoint] Add example to show how to extend FileSystemWriter

wz337:
This is an example to show how to subclass FileSystemWriter and overriding prepare_local_plan. This would fix the issue of directory not exists when running checkpoint on multiple hosts.


fduwjj (approved):
Do you also want to add some test for this? Unblock for now.

PR #726, opened 2023-01-23
[SPMD] Cleanup - 1/N

anj-s:
1/N cleanup of unused code and print statements that can be added as part of the debug logging config.


examples/spmd/run_spmd.py
```diff
@@ -143,7 +143,7 @@ def work_main(rank: int, world_size: int) -> None:
                 ),
                 placements=[Replicate()],
             ),
-            optimize_first_iter=True,
+            force_compile=True,
```
awgu, line 146: I wanted to check that this change was intentional (since I am not sure if `run_spmd.py` runs in CI).

It looks like the `SPMD` API uses `optimize_first_iter`, not `force_compile`.

https://github.com/pytorch/tau/blob/7c71c23e6aea3f69f98859921447b6eea74285b2/spmd/compiler/api.py#L20-L28

anj-s: I had moved to rename this param but reverted as you can see. thanks for catching it!

spmd/compiler/distribute.py
```diff
@@ -603,7 +603,7 @@ def distribute(
     dist_graph: DistributedGraph,
     param_schema: Schema,
     input_schemas: Sequence[Placement],
-    force_compile: bool,
+    optimize_first_iter: bool,
```
aazzolini, line 606: I believe this will break fbcode stuff.

anj-s: This is reverting a change that was made in a previous PR by mistake I think. We are only using `optimize_first_iter` everywhere.

spmd/compiler/distribute.py
```diff
@@ -654,7 +654,7 @@ def gather_inputs_for_compilation(
     #
     # TODO(chienchin): figure out how to be compatible with optimizer state_dict
     # loading.
-    if force_compile:
+    if optimize_first_iter:
```
lessw2020, line 657: I see - this param flipped in the original code from optimize_first_iter in api.py, and auto became force_compile via positional param.
So this change simply corrects it's naming to be consistent.

lessw2020 (approved):
everything looks good to me.
(fixing the name change from force_compile to optimize_first_iter was a good catch).

PR #729, opened 2023-01-26
[checkpoint] Fix pg planner

wz337:
Fix broken CI in pg_planner:
```
spmd/checkpoint/pg_planner.py:61:9: error: "init" undefined in superclass  [misc]
spmd/checkpoint/pg_planner.py:81:9: error: "init" undefined in superclass  [misc]
```

This is changed due to https://github.com/pytorch/pytorch/pull/92829.


anj-s (approved):
Thanks for fixing this!

PR #730, opened 2023-01-27
[SPMD] Cleanup 2/N

anj-s:
* Rename functions and refactor to cleanup _convert_to_distribute
* Removed some variables that don't seem to be used
* Removed function params that were being passed around for debug logging


kumpera (approved):
LGTM

PR #735, opened 2023-02-06
[spmd] lint fixes

wanchaol:


awgu (approved):
:)

spmd/tensor/experimental_ops.py
```diff
@@ -14,9 +14,10 @@
 from torch.distributed._tensor.ops.utils import register_prop_rule
 from torch.distributed._tensor.ops.common_rules import pointwise_rule
 
-aten = torch.ops.aten
+aten = torch.ops.aten  # pyre-ignore
```
anj-s, line 17: is there any guidance on which errors we should ignore and which we should fix?

wanchaol: I think the rule of thumb is that if the lint error is easy to fix, we should fix it instead of ignoring it. Also we should fix mostly every mypy errors, pyre is in a lower priority

wanchaol: The pyre linting was introduced to make internal codebase happier as internal uses pyre. pytorch uses mypy mostly.  I think we can keep pyre if we care more about internal cases, and fixes easy pyre lint failures, as long as it's manageable with ignores.

PR #742, opened 2023-02-13
Make output chunk spec optional

kwen2501:
`output_chunk_spec` was used to specify how micro-batch outputs should be merged or reduced (if the output field is a loss).

### Previous
Previously, user would need to specify an `output_chunk_spec` to `PipelineDriver`.
For example:
```
output_chunk_spec = {'loss': CustomReducer(torch.tensor(0.0), lambda a, b: a + b),
                     'logits': TensorChunkSpec(0),
                     'encoder_last_hidden_state': TensorChunkSpec(0)}
```

### Now
We automate this process by:
1. Adding an optional `loss_reducer` argument to `PipelineDriver`, which has a default value of `sum_reducer`, i.e.
```
sum_reducer = CustomReducer(torch.tensor(0.0), lambda a, b: a + b)
```
2. If an output field is specified/identified as loss in the earlier `Pipe.from_tracing` call, we would use the provided `loss_reducer` for this field. If `loss_reducer` is not specified, we would use the default `sum_reducer`.
3. For a non-loss output field, we would use `TensorChunkSpec(0)` by default, i.e. merging along the 0 dimension.

### DIFF
--- a/example.py
+++ b/example.py
@@ -124,10 +124,10 @@ if local_rank == 0:
     driver = PipelineDriverFillDrain(
         pipe,
         64,
+        world_size=world_size,
         args_chunk_spec=args_chunk_spec,
         kwargs_chunk_spec=kwargs_chunk_spec,
         output_chunk_spec=output_chunk_spec,
-        world_size=world_size,
     )
 
     x = torch.randn(512, 512)
--- a/example_train.py
+++ b/example_train.py
@@ -137,10 +137,10 @@ if local_rank == 0:
     driver = PipelineDriverFillDrain(
         pipe,
         64,
+        world_size=world_size,
         args_chunk_spec=args_chunk_spec,
         kwargs_chunk_spec=kwargs_chunk_spec,
         output_chunk_spec=output_chunk_spec,
-        world_size=world_size,
     )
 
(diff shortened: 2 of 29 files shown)


HamidShojanazeri:
Testing the PR, see its failing on "'generated_loss_spec' referenced before assignment" in IR.py https://gist.github.com/HamidShojanazeri/5db4af143f051624ad42b7bf35b8b626

HamidShojanazeri (commented):
Thanks @kwen2501 , testing the PR, see its failing on "'generated_loss_spec' referenced before assignment" in IR.py https://gist.github.com/HamidShojanazeri/5db4af143f051624ad42b7bf35b8b626

kwen2501:
> Thanks @kwen2501 , testing the PR, see its failing on "'generated_loss_spec' referenced before assignment" in IR.py https://gist.github.com/HamidShojanazeri/5db4af143f051624ad42b7bf35b8b626

Thanks. Fixed for inference mode in [39db031](https://github.com/pytorch/tau/pull/742/commits/39db031b7972f09650bc80fa2b44be96649c7450)

HamidShojanazeri:
Thanks @kwen2501 , could successfully test it.

PR #750, opened 2023-02-16
update AOTConfig params for functionalization of inference

lessw2020:
AOT pushed an update that adds a param related to handling mutations of the graph during inference. 
Unfortunately while they updated inductor for it, they did not add a default param and so running with latest PT will err out. 
This PR just adds the new param for patched_aot_function.py so we can continue as before.


lessw2020:
Reference PR that caused this param change:  https://github.com/pytorch/pytorch/pull/92857

fegin (approved):

PR #759, opened 2023-03-01
Inference example compile update

HamidShojanazeri:
integrating pippy.all_compile() changes to inference example.

### DIFF
--- a/examples/inference/HF_inference.py
+++ b/examples/inference/HF_inference.py
@@ -1,7 +1,6 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
 import argparse
 import inspect
-import logging
 import os
 import time
 
@@ -9,9 +8,6 @@ import torch
 import pippy
 import pippy.fx
 from pippy import run_pippy
-from pippy.IR import MultiUseParameterConfig, Pipe
-from pippy.PipelineDriver import PipelineDriverFillDrain, PipelineDriver1F1B, PipelineDriverInterleaved1F1B, \
-    PipelineDriverBase
 from pippy.hf import PiPPyHFTracer
 from pippy import split_on_size_threshold, split_into_equal_size
 from transformers import  AutoModelForSeq2SeqLM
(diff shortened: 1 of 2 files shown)


examples/inference/HF_inference.py
```diff
@@ -95,17 +96,17 @@ def run_master(pp_ranks, args):
 
     # preparing inputs based on the model choice
     if 't5' in args.model_name:
-        inp = torch.empty(bs, seq_length, dtype=torch.long, device=device).random_(model.config.vocab_size)
+        inp = torch.empty(bs, seq_length, dtype=torch.long).random_(model.config.vocab_size)
```
kwen2501, line 99: You seem to have mentioned that this works but I never understand why :)
Have you tested it under both cuda and cpu modes?

examples/inference/HF_inference.py
```diff
@@ -114,30 +115,17 @@ def run_master(pp_ranks, args):
 
     print('Instantiating model Pipeline')
     model_init_start = time.time()
-    model_pipe = Pipe.from_tracing(model, MULTI_USE_PARAM_CONFIG, tracer=PiPPyHFTracer(), concrete_args=concrete_args,
-                                output_loss_value_spec=None, split_policy=split_policy
-                                )
```
kwen2501, line 119: Preparation of `MULTI_USE_PARAM_CONFIG` can be deleted too :)

examples/inference/HF_inference.py
```diff
@@ -6,6 +6,7 @@
 import time
 
 import torch
+import pippy
 import pippy.fx
 from pippy import run_pippy
 from pippy.IR import MultiUseParameterConfig, Pipe
```
kwen2501, line 12: This import is not needed anymore, I think.

kwen2501: Neither do we need
`from pippy.PipelineDriver import PipelineDriverFillDrain ...`

examples/inference/HF_inference.py
```diff
@@ -114,30 +115,17 @@ def run_master(pp_ranks, args):
...
-                                                                )
+  
+    pipe_driver = pippy.compile(
+        model,
+        num_ranks=args.world_size,
+        num_chunks=chunks,
+        schedule=args.schedule,
+        split_policy=split_policy,
+        tracer=PiPPyHFTracer(),
+        checkpoint=bool(args.checkpoint),
```
kwen2501, line 126: We can remove the `checkpoint` option. (also from parser)

kwen2501 (approved):
Thanks for the great work! Looks much cleaner now!

PR #760, opened 2023-03-02
Introduce pippy.all_compile

kwen2501:
The GSPMD equivalent of `pippy.compile`.

`pippy.all_compile` enables all ranks to call into it and would perform stage-local module initialization.

The all_compile output is a little bit different than compile -- it returns not just the driver, but also the stage module, i.e.
```
pipe_driver, stage_mod = pippy.all_compile(
    mod,
    ...
)
```
For non-master rank, `pipe_driver` is return as `None`.

So for the data run, user would need an `if`:
```
if pipe_driver:
    pipe_driver(**input_dict)
```

### DIFF
--- a/.github/workflows/pippy_gpu_tests.sh
+++ b/.github/workflows/pippy_gpu_tests.sh
@@ -36,7 +36,7 @@ python3 test/local_test_forward.py --replicate ${REPLICATE} -s ${SCHEDULE}
 python3 test/local_test_forward_backward.py --replicate ${REPLICATE} -s ${SCHEDULE}
 python3 test/local_test_compile.py -s ${SCHEDULE}
 python3 examples/hf/gpt2/pippy_gpt2.py --replicate ${REPLICATE} -s ${SCHEDULE}
-python3 examples/gspmd/pippy_gspmd.py --replicate ${REPLICATE} -s ${SCHEDULE}
+python3 examples/gspmd/pippy_gspmd.py -s ${SCHEDULE}
 
 # Run flaky integration tests
 python3 test/local_test_ddp.py --replicate ${REPLICATE} -s ${SCHEDULE}
--- a/examples/gspmd/pippy_gspmd.py
+++ b/examples/gspmd/pippy_gspmd.py
@@ -6,25 +6,13 @@ import unittest
 import torch
 import torch.autograd.profiler_legacy
 
+import pippy
 import pippy.fx
 from pippy import run_pippy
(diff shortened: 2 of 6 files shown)


HamidShojanazeri:
Thanks very much @kwen2501 for the prompt fix! one quick note about the API naming, I wonder this would be only dedicated to inference or general.  In case its inference only it would help if use names such as pippy.inference() or pippy.inference.init() to make this easier for user.

HamidShojanazeri:
It would be also great to know if input type to the pipedriver has changed. Previously, we needed to pass a dict as input to the pipedriver in HF_inference.py but this seems run into this [error](https://gist.github.com/HamidShojanazeri/7d424dbad09b04d841a6603a08e60f51) now. 

Also, on the GPU mem allocations, it  seems to be different from expected. Not sure if I am missing something in my [changes](https://gist.github.com/HamidShojanazeri/4376e5d353ec09bf5cd1aed7a1417b28#file-gistfile1-txt-L118).

```
memory_reserved after model intializaed with pipelines on each rank
memory_reserved after model intializaed with pipelines on each rank
memory_reserved after model intializaed with pipelines on each rank
===================================================================
===================================================================
===================================================================
 0.002 GB
 0.002 GB
memory_allocated after model intializaed with pipelines on each rank
 0.002 GB
memory_allocated after model intializaed with pipelines on each rank
===================================================================
memory_allocated after model intializaed with pipelines on each rank
===================================================================
 0.0 GB
===================================================================
 0.0 GB
 ```

**Comment by kwen2501** [comment]
> Thanks very much @kwen2501 for the prompt fix! one quick note about the API naming, I wonder this would be only dedicated to inference or general. In case its inference only it would help if use names such as pippy.inference() or pippy.inference.init() to make this easier for user.

It works for both training and inference modes.
Previously PiPPy uses `output_loss_spec` to determine whether there is a loss and hence whether backward is needed. 
I guess that's a bit implicit. We can add a flag like `train_mode=True|False` for users to more explicitly specify the purpose.

**Comment by kwen2501** [comment]
> but this seems run into this [error](https://gist.github.com/HamidShojanazeri/7d424dbad09b04d841a6603a08e60f51) now

The `all_compile` output is a little bit different than `compile` -- it returns not just the driver, but also the stage module.
```
pipe_driver, stage_mod = pippy.all_compile(
    mod,
    ...
)
```
For non-master rank, `pipe_driver` is return as `None`.

So for the data run, you would need an `if`:
```
if pipe_driver:
    pipe_driver(**input_dict)
```
Let me know what you think about this API signature.

PR #762, opened 2023-03-04
Simplify tp+pp example by using pippy.all_compile

kwen2501:
`pippy.all_compile` would output the stage module, hence easier for TP.

### DIFF
--- a/examples/tp+pp/pippy_tp.py
+++ b/examples/tp+pp/pippy_tp.py
@@ -5,18 +5,10 @@ import unittest
 
 import torch
 
+import pippy
 import pippy.fx
 from pippy import run_pippy
-from pippy.IR import (
-    Pipe,
-    pipe_split,
-)
-from pippy.PipelineDriver import (
-    PipelineDriverFillDrain,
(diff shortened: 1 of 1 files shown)


kwen2501:
Cc @fduwjj @wanchaol

examples/tp+pp/pippy_tp.py
```diff
@@ -180,8 +158,7 @@ def main(args=None):
         "-s",
         "--schedule",
         type=str,
-        default=list(schedules.keys())[0],
-        choices=schedules.keys(),
+        default="FillDrain",
```
fduwjj, line 161: N00b question what does this mean?

fduwjj (approved):
LGTM

examples/tp+pp/pippy_tp.py
kwen2501: Fill-drain is a pipeline scheduling policy. Other options include 1F1B, interleaved 1F1B. See for example [this paper](https://arxiv.org/pdf/2104.04473.pdf), section 2.2.

PR #774, opened 2023-04-04
Inference example update

HamidShojanazeri:


kwen2501 (approved):
LGTM. Thanks @HamidShojanazeri !

kwen2501:
Also including suggestion from https://github.com/zsc/tau/pull/1 to use more general class `AutoModelForCausalLM` instead of `OPTForCausalLM` to support more models.

kwen2501:
Thanks @zsc for your suggestion!

PR #779, opened 2023-04-18
Load lm_head from decoder.embed_tokens.weight

kwen2501:
Further fixes issue in https://github.com/pytorch/PiPPy/pull/777#pullrequestreview-1389045572

For OPT, the `lm_head` weight is shared with the embed tokens weight.

See https://github.com/huggingface/transformers/blob/main/src/transformers/models/opt/modeling_opt.py#L812-L820

### DIFF
--- a/pippy/IR.py
+++ b/pippy/IR.py
@@ -1062,7 +1062,19 @@ class Pipe(torch.nn.Module):
                     device=device,
                     dtype=dtype,
                 )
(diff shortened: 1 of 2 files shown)


kwen2501 (commented):
Landing this PR as it looks similar to the fix proposed by @jiqing-feng in https://github.com/pytorch/PiPPy/pull/780

PR #780, opened 2023-04-18
fix loading opt models

jiqing-feng:
Fixed loading opt models. The prefix of module names are different across models, so we need to consider different models in the loading module function. And I also add GPT to the support models
@kwen2501 @HamidShojanazeri Could you please review it? Thanks!

### DIFF
--- a/examples/inference/hf_generate.py
+++ b/examples/inference/hf_generate.py
@@ -108,7 +108,7 @@ if __name__ == "__main__":
 
     assert args.world_size % args.pp_group_size == 0
 
-    supported_model_categories = ["opt", "gpt2", "bloom", "EleutherAI/gpt", "codegen"]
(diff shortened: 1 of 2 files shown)


kwen2501:
Hi @jiqing-feng , thanks for the quick fix!!
Out of coincidence, I also created a PR yesterday: https://github.com/pytorch/PiPPy/pull/779
So glad to see that my fix is similar to yours.
Since the other PR has a check in addition (which can give us better knowledge when things fail), do you mind if we land the change there?
Thank you!

jiqing-feng:
> Hi @jiqing-feng , thanks for the quick fix!! Out of coincidence, I also created a PR yesterday: #779 So glad to see that my fix is similar to yours. Since the other PR has a check in addition (which can give us better knowledge when things fail), do you mind if we land the change there? Thank you!

Sure, it is great to see it has been fixed

kwen2501:
Thanks @jiqing-feng . Closing.

PR #794, opened 2023-05-04
Pure c10d example

kwen2501:
## Description

Small example that only uses c10d and no rpc.

## Test
```
$ torchrun --nproc_per_node 4 local_test_c10d.py
# cpu:
equivalence test passed 95468232.0 ref 95468232.0
# cuda:
equivalence test passed 95261472.0 ref 95261472.0
```


fduwjj:
Can you educate me on what does pure c10d refer to? Do you mean SPMD style or?

PR #814, opened 2023-06-20
Save module

eddogola:
## Description

Replaced the initial mechanism of writing to file from a direct file.write to a more atomic write, using tempfile, to prevent data corruption if the data to write is large.

Fixes #(issue)

## Type of change

Working on a test on another PR

## Checklist:


kwen2501:
Can you add some comments to the key operations in this PR please? Thanks!

fegin:
What is the PR for? Can you add some description? Als lint failure is real.

pippy/hf/_SaveModule.py
```diff
@@ -1,11 +1,59 @@
...
+        file_contents (str): contents to write to file
+        target_file_path (str): path to write to
+        mode (str, optional): mode to write file with. Defaults to "w". Only "w" and "a" are supported.
+    """
+    temp_file = tempfile.NamedTemporaryFile(
+        delete=False, dir=os.path.dirname(target_file_path)
+    )
+    try:
+        if os.path.exists(target_file_path):
+            _copy_with_metadata(target_file_path, temp_file.name)
```
kwen2501, line 43: To add comment: why do we need a copy first?
nit: what would be the impact on performance?

pippy/hf/_SaveModule.py
```diff
@@ -1,11 +1,59 @@
...
+        with open(temp_file.name, mode) as f:
+            f.write(file_contents)
+            f.flush()
+            os.fsync(f.fileno())
```
kwen2501, line 47: To add comment: what does `os.fsync` do?

pippy/hf/_SaveModule.py
```diff
@@ -1,11 +1,59 @@
...
+
+        os.replace(temp_file.name, target_file_path)
+    finally:
+        if os.path.exists(temp_file.name):
+            try:
+                os.unlink(temp_file.name)
```
kwen2501, line 53: To add comment: what does `os.unlink` do?

pippy/hf/_SaveModule.py
```diff
@@ -1,11 +1,59 @@
...
+            except:
+                pass
```
kwen2501, line 55: To add comment: why do we consider we can `pass` here?

pippy/hf/_SaveModule.py
```diff
@@ -41,11 +93,12 @@ def _save_index(
     if not os.path.exists(checkpoint_dir):
         os.mkdir(checkpoint_dir)
 
-    with open(filepath, "w") as f:
-        f.write(json_str)
+    _atomic_write(json_str, filepath)
```
kwen2501, line 96: To add comment: why do we think atomic write is important here?

pippy/hf/_SaveModule.py
```diff
@@ -1,11 +1,59 @@
...
+    """
+    Copy file with all its permisions and metadata.
+
+    Args:
+        source (str): path to source file
+        target (str): path to copy to
+    """
+    shutil.copy2(source, target)
+    st = os.stat(source)
+    os.chown(target, st.st_uid, st.st_gid)
```
kwen2501, line 26: To add comment: would `chown` normally work without certain privilege?

pippy/hf/_SaveModule.py
wz337: I think in our case, the target_file_path should never exist, as we will be writing the index file during saving. It means that we only need to support both file creation. What do you think?

eddogola: it was just a sanity check in case the user was overwrite a ckpt, but I agree, we can do away with that.

wz337: Makes sense. Sounds good. Then, let's keep it for now. We can revisit if necessary.

pippy/hf/_SaveModule.py
```diff
@@ -41,11 +51,15 @@ def _save_index(
...
         os.mkdir(checkpoint_dir)
 
-    with open(filepath, "w") as f:
-        f.write(json_str)
+    # write index file atomically to avoid partial/corrupted writes
+    _atomic_write(json_str, filepath)
+
+    logger.info(f"Saved index file to {filepath}")
+
+    logger.info(f"Saved index file to {filepath}")
```
wz337, line 59: nit: are these repeated? Let's remove one.

pippy/hf/_SaveModule.py
wz337: Maybe accidentally delete the entire _atomic_write?

pippy/hf/_SaveModule.py
```diff
@@ -0,0 +1,75 @@
...
+    weight_map = {}
+    for idx, (_, submod) in enumerate(pipe.split_gm.named_children()):
+        # chain both params and buffers generators together
+        params_buffers = chain(
+            submod.named_parameters(), submod.named_buffers()
+        )
+        for param_name, _ in params_buffers:
+            old_name = submod.remap_qualname(param_name)
+
+            binary_filename = _get_binary_filename(idx)
```
kwen2501, line 40: nit: it seems this line can be out of the second for loop.

pippy/hf/_SaveModule.py
```diff
@@ -0,0 +1,75 @@
...
+import pippy
+
+from itertools import chain
+import logging
+import json
+import os
+
+CKPT_INDEX_JSON_FILENAME = "pytorch_model.bin.index.json"
+
+logger = logging.getLogger(__name__)
```
kwen2501, line 11: nit: In PiPPy we mainly use `logging.info`. Maybe stay the same for now?

pippy/hf/_SaveModule.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def _save_index(
+    pipe: pippy.fx.GraphModule,
```
kwen2501, line 15: `pipe` should be of the `Pipe` class, not a `pippy.fx.GraphModule`

wz337 (approved):
LGTM

pippy/hf/_SaveModule.py
```diff
@@ -19,33 +19,35 @@ def _atomic_write(file_contents: str, target_file_path: str, mode="w") -> None:
         mode (str, optional): mode to write file with. Defaults to "w". Only "w" and "a" are supported.
     """
     try:
-        with open(temp_file.name, mode) as f:
+        with open(target_file_path, mode) as f:
```
wz337, line 22: Let's keep the temp_file here, because we want to make sure the target_file_path only exists when writing is finished. Otherwise, we could end up with a corrupted/half written file.

PR #818, opened 2023-06-23
add test to check index file output

eddogola:
## Description

Add single test to check if the contests of the ckpt metadata(index) file are accurate.

- [x] Test LocalIndexMetadataTest

## Checklist:

- [x] Have you added tests that prove your fix is effective or that this feature works?

### DIFF
--- a/test/local_test_ckpt_index_file.py
+++ b/test/local_test_ckpt_index_file.py
@@ -1,27 +1,42 @@
 from typing import List
 import unittest
 import argparse
+import shutil
 import json
 import os
 
 from pippy.IR import pipe_split, TrivialLossWrapper
+from pippy.hf._SaveModule import _save_index
 from pippy.compile import compile_stage
 import torch.distributed as dist
 import torch.optim as optim
 import torch
 
-d_hid = 512
-chunk_size = 256
+DEFAULT_FILENAME = "pytorch_model.bin.index.json"
+CKPT_DIR = "test_ckpts"
+WEIGHT_MAP = set(
+    [
+        "module.mm_param0",
+        "module.mm_param1",
+        "module.mm_param2",
+        "module.lin0.weight",
+        "module.lin0.bias",
+        "module.lin1.weight",
+        "module.lin1.bias",
+    ]
+)
+D_HID = 512
+CHUNK_SIZE = 256
 
 
 class ExampleCode(torch.nn.Module):
     def __init__(self):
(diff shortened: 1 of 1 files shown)


eddogola:
@wz337 @kwen2501

test/local_test_ckpt_index_file.py
```diff
@@ -0,0 +1,170 @@
...
+        # check file written on disk to given location
+        assert os.path.exists(filepath)
+
+        # check total_size correct
+        size_calc = sum(param.numel() for param in ec.parameters()) * 4
+        assert size_calc == data["metadata"]["total_size"]
+
+        # check all params present
+        assert len(data["weight_map"]) == 7
+        assert "module.mm_param0" in data["weight_map"]
```
wz337, line 106: As this would be a constant value, we can create a set with these keys as WEIGHT_MAP = set("module.mm_param0", "module.mm_param1"...) up on line 43 and loop through the set to see if they are all in data['weight_map'].

test/local_test_ckpt_index_file.py
```diff
@@ -0,0 +1,170 @@
...
+        optimizer.step()
+
+    dist.barrier()
+    print(f"Rank {args.rank} completes")
+
+    # Last rank checks index file
+    if args.rank == args.world_size - 1:
+        from pippy.hf._SaveModule import _save_index
+
+        CKPT_DIR = "test_ckpts"
```
wz337, line 88: We can put these constant up top outsides of the test case. Just separate them from the test logics so it is a bit clearer and easier to change if needed. 

In general, the recommended practice is to define constants at the top right after any import statements. This way, people reading your code will immediately know the constants' purpose and expected treatment.

wz337 (commented):
Overall LGTM! 
@kwen2501 if you also take a look at reviewing.

test/local_test_ckpt_index_file.py
```diff
@@ -0,0 +1,170 @@
...
+            stage()
+
+        # Take an optimization step
+        optimizer.step()
+
+    dist.barrier()
+    print(f"Rank {args.rank} completes")
+
+    # Last rank checks index file
+    if args.rank == args.world_size - 1:
```
wz337, line 85: Could we add a comment that we only need one rank to check the index file? 
Likely, during the actual saving, we will be using rank0 to generate the index file. Let's use rank 0 to check this instead.

PR #823, opened 2023-06-28
combine index-file-saving and params-saving into one API

eddogola:
## Description

bring together `_save_index` and `_save_params` into one function `save_checkpoint`. The entire module is used to write the index file, only in rank 0, while submodule parameters are saved to a file in each rank. The associated ckpt test is updated to reflect this change.

## Type of change

- [x] Test LocalIndexMetadataTest

## Checklist:

- [x] Have you added tests that prove your fix is effective or that this feature works?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+docs/build
+__pycache__
+build
+pippy.egg-info
+pippy/version.py
+dist
+.idea/
+.pyre/
+**/*.json
+**/*.out
+**/.DS_STORE
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-docs/build
 __pycache__
 build
 pippy.egg-info
--- a/test/local_test_ckpt_index_file.py
+++ b/test/local_test_ckpt_index_file.py
@@ -6,10 +6,7 @@ import shutil
 import json
 import os
 
-from pippy.hf._SaveModule import (
-    _save_index,
-    _save_checkpoint,
-)
+from pippy.hf._SaveModule import save_checkpoint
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -0,0 +1,11 @@
+docs/build
```
wz337, line 1: Let's remove this from the commit.

test/local_test_ckpt_index_file.py
```diff
@@ -6,10 +6,7 @@
 import json
 import os
 
-from pippy.hf._SaveModule import (
-    _save_index,
-    _save_checkpoint,
-)
+from pippy.hf._SaveModule import save_checkpoint
```
wz337, line 9: Also, I think we can update this test name to reflect what we are testing. 
local_test_save_load_checkpoint.py

test/local_test_ckpt_index_file.py
```diff
@@ -94,10 +91,13 @@ def run_worker(args: List[str | int]) -> None:
     else:
         stage()
 
+    # Take an optimization step
+    optimizer.step()
+    ref = deepcopy(stage.submod.state_dict())
+    save_checkpoint(stage, stage.submod, CKPT_DIR)
```
fegin, line 97: What's the reason to move the three lines here? Does this mean the original test case have an issue?

eddogola: save_checkpoint writes the index as well as saving weights. As they were decoupled before, I was writing the index in the last set of lines and the checkpoints further up. Now, I had to bring save_checkpoint up to do the whole thing.

PR #824, opened 2023-06-28
[BE] Apply ufmt to all Python files except for pippy/fx/*.py and change the check.sh to use ufmt

fegin:
Apply ufmt to all Python files except for pippy/fx/*.py and change the check.sh to use ufmt. 

This PR also fixes some failing tests.


kwen2501 (commented):
Approved.
Thanks much for aligning lint formats!

PR #826, opened 2023-06-29
save optim state dict

eddogola:
## Description

Save optimizer's state into a diff binary file such that for each rank, there are two state bin files: submod params and optim state.

Fixes #(issue)

## Type of change

Will update checkpoint test to reflect change in a different PR.

## Checklist:

### DIFF
--- a/pippy/hf/_SaveModule.py
+++ b/pippy/hf/_SaveModule.py
@@ -103,7 +103,7 @@ def _save_index(
     logging.info(f"Saved index file to {filepath}")
 
 
-def _get_binary_filename(cur_idx: int) -> str:  # type: ignore[valid-type]
+def _get_binary_filename(cur_idx: int, is_optim: bool = False) -> str:  # type: ignore[valid-type]
     """
     Gets filename for pytorch checkpoint binary based on current index and world size.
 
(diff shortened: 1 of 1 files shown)


eddogola:
@wz337 @fegin @kwen2501

eddogola:
I'm going to work on the test in a different PR. I am thinking of writing a different function to load the optimizer since I'm afraid load_checkpoint won't cover that.

pippy/hf/_SaveModule.py
```diff
@@ -141,7 +143,33 @@ def _save_params(submod: torch.nn.Module, checkpoint_dir: str) -> None:
...
+    Args:
+        optimizer(`torch.optim.Optimizer`): pytorch optimizer
+        checkpoint_dir(`str`): where to keep the checkpoint binaries
+    """
+    if not os.path.exists(checkpoint_dir):
+        os.mkdir(checkpoint_dir)
+    filepath = os.path.join(
+        checkpoint_dir, _get_binary_filename(dist.get_rank(), is_optim=True)
+    )
+    torch.save(
```
wz337, line 161: You can directly just torch.save(optimizer.state_dict()).
The optimizer structure is different from model_state_dict if you print it out.

wz337 (approved):

wz337:
We will need to rethink the flow a little bit more regarding what the API looks like, but we can land it as intermediate commit.

pippy/hf/_SaveModule.py
fegin: What's the motivation of saving parameters individually? If there is a valid reason to do so, should we do the same thing for the optimizer state_dict, saving each optimizer state in a separate file? We should consider how users want the states to be saved and it is best to be consistent for model and optimizer.

fegin:
Add the comment here to avoid being resolved:

What's the motivation of saving parameters individually? If there is a valid reason to do so, should we do the same thing for the optimizer state_dict, saving each optimizer state in a separate file? We should consider how users want the states to be saved and it is best to be consistent for model and optimizer. 

We need to think about how the final API look lik.

PR #827, opened 2023-06-29
test optimizer using torch.load

eddogola:
## Description

Since we're also saving the optimizer's state dict (in a different binary file for each rank), we can't use load_checkpoint to check that it works. Here, I use torch.load to check that a certain reference optimizer state dict is similar/close to the saved optimizer's state dict, after a `step()` run.

## Type of change

## Checklist:

### DIFF
--- a/test/local_test_checkpoint.py
+++ b/test/local_test_checkpoint.py
@@ -12,7 +12,7 @@ import torch.distributed as dist
 import torch.optim as optim
 from pippy.compile import compile_stage
 
-from pippy.hf._SaveModule import save_checkpoint, _get_binary_filename
+from pippy.hf._SaveModule import _get_binary_filename, save_checkpoint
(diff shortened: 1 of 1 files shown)


eddogola:
@wz337 @kwen2501 @fegin

test/local_test_checkpoint.py
```diff
@@ -96,8 +96,9 @@ def run_worker(args: List[str | int]) -> None:
 
     # Take an optimization step
     optimizer.step()
-    ref = deepcopy(stage.submod.state_dict())
-    save_checkpoint(stage, CKPT_DIR)
+    submod_ref = deepcopy(stage.submod.state_dict())
```
wz337, line 99: Let's call it ref_state_dict and ref_opt_state_dict.

wz337 (approved):

PR #828, opened 2023-06-30
Enable pipeline + DDP (c10d version)

kwen2501:
This is a mirror of #108, implemented in the c10d context.

User can use DDP to wrap the pipeline stage submodule compiled by PiPPy, achieving composability compared to the previous `init_data_parallel` API offered by PiPPy.

For each pipeline stage, we omit DDP synchronization (i.e. gradient all-reduce) on all except the last backward micro-batch.

A `local_test_c10d_ddp.py` has been added too.

**Update 07/14:**
Added gradient equivalence test.
Test command:
```
torchrun --nproc-per-node 8 local_test_c10d_ddp.py
```
Test output:
```
PP group size = 4, DP group size = 2
...
Output equivalence test passed
Output equivalence test passed
Checking gradient of lin1.weight
Checking gradient of lin1.bias
Gradient equivalence test passed
Gradient equivalence test passed
Checking gradient of mm_param2
Gradient equivalence test passed
Gradient equivalence test passed
Checking gradient of mm_param1
Checking gradient of lin0.weight
Checking gradient of mm_param0
Checking gradient of lin0.bias
Gradient equivalence test passed
Gradient equivalence test passed
Gradient equivalence test passed
Gradient equivalence test passed
```

### DIFF
--- a/pippy/PipelineStage.py
+++ b/pippy/PipelineStage.py
@@ -486,6 +486,8 @@ class PipelineStage(torch.nn.Module):
         return grad_send_reqs
 
     def forward(self, *args, **kwargs):
+        # If submod is wrapped with DDP, we use the `no_sync` context manager to
+        # avoid gradient all-reduce per microbatch
         def forward_maybe_with_ddp(*args, **kwargs):
             if isinstance(self.submod, DistributedDataParallel):
                 with self.submod.no_sync():  # type: ignore[operator]
@@ -510,6 +512,7 @@ class PipelineStage(torch.nn.Module):
                     with self.submod.no_sync():  # type: ignore[operator]
                         grads_input, _ = stage_backward(**bwd_kwargs)
             else:
+                # Non-DDP submodule, regular backward
                 grads_input, _ = stage_backward(**bwd_kwargs)
             return grads_input
 
--- a/pippy/PipelineStage.py
+++ b/pippy/PipelineStage.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import torch
 import torch.distributed as dist
+from torch.nn.parallel import DistributedDataParallel
 
 import pippy
 import pippy.fx
@@ -485,6 +486,14 @@ class PipelineStage(torch.nn.Module):
         return grad_send_reqs
 
     def forward(self, *args, **kwargs):
+        def forward_maybe_with_ddp(*args, **kwargs):
+            if isinstance(self.submod, DistributedDataParallel):
+                with self.submod.no_sync():  # type: ignore[operator]
+                    out_val = self.submod(*args, **kwargs)
+            else:
+                out_val = self.submod(*args, **kwargs)
+            return out_val
+
         # map microbatch ID to list of forward tensor args
         fwd_cache: Dict[int, Tuple[Any, List[torch.Tensor]]] = {}
 
@@ -513,7 +522,9 @@ class PipelineStage(torch.nn.Module):
 
             # Compute forward
             try:
-                output = self.submod(*composite_args, **composite_kwargs)
+                output = forward_maybe_with_ddp(
+                    *composite_args, **composite_kwargs
+                )
 
             except Exception as e:
                 exc_msg = f"""
@@ -565,6 +576,19 @@ class PipelineStage(torch.nn.Module):
                 bwd_kwargs["output_grads"] = (
                     grads if len(grads) > 0 else (None,)
                 )
+
+                if (
+                    isinstance(self.submod, DistributedDataParallel)
+                    and bwd_chunk == self.chunks - 1
+                ):
+                    # HACK: reaching into DDP implementation details here. Is there a better way?
+                    self.submod.reducer.prepare_for_backward(  # type: ignore[union-attr, operator]
+                        list(
+                            torch.nn.parallel.distributed._find_tensors(  # type: ignore[attr-defined]
+                                bwd_kwargs["stage_output"]
+                            )
+                        )
+                    )
 
                 # `stage_backward` node does not have `args`, only `kwargs`
                 grads_input, _ = stage_backward(**bwd_kwargs)
--- a/pippy/PipelineStage.py
+++ b/pippy/PipelineStage.py
@@ -494,6 +494,25 @@ class PipelineStage(torch.nn.Module):
                 out_val = self.submod(*args, **kwargs)
             return out_val
 
+        def backward_maybe_with_ddp(bwd_kwargs: Dict, is_last_chunk: bool):
+            if isinstance(self.submod, DistributedDataParallel):
+                if is_last_chunk:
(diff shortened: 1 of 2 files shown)


pippy/PipelineStage.py
```diff
@@ -485,6 +486,36 @@ def _send_grads(
...
+        # avoid gradient all-reduce per microbatch
+        def forward_maybe_with_ddp(*args, **kwargs):
+            if isinstance(self.submod, DistributedDataParallel):
+                with self.submod.no_sync():  # type: ignore[operator]
+                    out_val = self.submod(*args, **kwargs)
+            else:
+                out_val = self.submod(*args, **kwargs)
+            return out_val
+
+        def backward_maybe_with_ddp(bwd_kwargs: Dict, is_last_chunk: bool):
```
fegin, line 499: Probably change the name with `backward_maybe_with_nosync`. We can use the same logic for FSDP.

pippy/PipelineStage.py
```diff
@@ -485,6 +486,36 @@ def _send_grads(
         return grad_send_reqs
 
     def forward(self, *args, **kwargs):
+        # If submod is wrapped with DDP, we use the `no_sync` context manager to
+        # avoid gradient all-reduce per microbatch
+        def forward_maybe_with_ddp(*args, **kwargs):
```
fegin, line 491: Probably change the name with `forward_maybe_with_nosync`. We can use the same logic for FSDP.

pippy/PipelineStage.py
```diff
@@ -485,6 +486,36 @@ def _send_grads(
...
+                    out_val = self.submod(*args, **kwargs)
+            else:
+                out_val = self.submod(*args, **kwargs)
+            return out_val
+
+        def backward_maybe_with_ddp(bwd_kwargs: Dict, is_last_chunk: bool):
+            if isinstance(self.submod, DistributedDataParallel):
+                if is_last_chunk:
+                    # HACK: reaching into DDP implementation details here. Is there a better way?
+                    self.submod.reducer.prepare_for_backward(  # type: ignore[union-attr, operator]
```
fegin, line 503: It seems to me that unless we change DDP's code, it is unlikely we can do without an intrusive call here.

fegin:
This supports the cases, DDP or DDP + TP. However, if users cut the stage into two different parts, the first part applies DDP while the rest apply DDP + TP, it seems to me that the PR could not support the usage. Let me know if I understand correctly.

kwen2501:
Do you mind describing more about why @fegin ? (This is not a counter argument). 
I want to understand more so that we know what to do for follow-up.
Thanks!

pippy/PipelineStage.py
kwen2501: @mrshenli I wonder if you have any suggestion?

fegin:
> Do you mind describing more about why @fegin ? (This is not a counter argument). I want to understand more so that we know what to do for follow-up. Thanks!

It is most likely happen when using composable APIs. But the composable API currently does not support `no_sync`. We can postpone the discuss to later.

pippy/PipelineStage.py
mrshenli: Would I be correct the problem is that DDP expects interleaving fwd and bwd pass, but here we need multiple fwd (i.e., # of micro-batches) and then multiple bwd?

I agree with @fegin that we might need to change DDP source code to create a dedicated API for this. 

One concern is that the contract of `no_sync()` is to call one pair of fwd+bwd after exiting the context. This PR is doing prepare_for_backward + bwd. `no_sync` also has impact on pre-forward logic like this one:

https://github.com/pytorch/pytorch/blob/c00dd43e43e9bdd041e44e5293c3fb8cd9e1bbf0/torch/nn/parallel/distributed.py#L1383-L1386

fduwjj: @fegin I am also wondering why the code here can work.

Also, here pippy accumulate grad while DDP do the average. Is it possible that we can do grad average across all micro-batches? We got this feedback from Mosaic AI and they said TP+FSDP does not have same loss curve. In TP, we do grad accumulation as well.

test/local_test_c10d_ddp.py
```diff
@@ -0,0 +1,188 @@
...
+            range(dp_rank, args.world_size, args.dp_group_size)
+        )
+        pp_group = dist.new_group(ranks=pp_group_ranks)
+        if dp_rank == my_dp_rank:
+            my_pp_group = pp_group
+    print(f"Rank {args.rank} done getting pp group")
+    return my_pp_group, my_pp_rank
+
+
+def run_worker(args):
```
fduwjj, line 77: Is it possible that we can compare the runs of two models, one is only using DDP and another one is DDP+PP?

pippy/PipelineStage.py
kwen2501: > here we need multiple fwd (i.e., # of micro-batches) and then multiple bwd?

Yes, @mrshenli, that's the use case here. 

Summarizing the comments so far, we have two questions:

Q1: is it okay to accumulate gradients across micro-batches? Or do we need gradient averaging?

Q2: if `no_sync()` is not ideal, do you know if there is a better way to compose with DDP? Thanks!

fegin: @mrshenli  It seems that `prepare_forward()` is for statistics and log only. Although it is not ideal to lose some functions, the accuracy should not be affected? 

@kwen2501 I feel `no_sync()` is the right way to go since `no_sync()` is created for people manually doing micro-batch for large activation. Can we verify the gradient with pure DDP to further understand the case?

rohan-varma: @kwen2501 @fegin Could you folks help me further understand the constraints here and why we need to use prepare_for_backward internal calls? If we need to support multiple forward + multiple backward under no_sync, shall we try to make that work in today's DDP? 

My concern with prepare_for_backward, call to _find_tensors, etc is that there may be edge cases around static_graph, find_unused_parameters, etc that we miss and we should try to resolve all the issues using public supported DDP APIs.

fegin: Could we land this version first if the accuracy is correct? This way we can unblock the testing TP + DDP. We can discuss how to do a proper change in DDP.

rohan-varma: sure!

fduwjj:
@fegin for DDP + TP, we are working to make it work in https://github.com/pytorch/pytorch/pull/103604/files.

pippy/PipelineStage.py
kwen2501: Hi @fegin @rohan-varma I added gradient equivalence test in `local_test_c10d_ddp.py`. The test is passing. 
Would you mind taking a second look? Thanks!

test/local_test_c10d_ddp.py
kwen2501: Added

fegin (approved):

fegin:
Let's land the PR to unblock the internal testing.

PR #830, opened 2023-07-04
selective 2d api/example added for fine-grained tp/pp demo

moonbucks:
## Description

added 2d parallelism (tp+pp) API/example for fine-grained tp/pp 

## Checklist:

### DIFF
--- a/examples/selective2d/2d_train.py
+++ b/examples/selective2d/2d_train.py
@@ -1,0 +1,349 @@
+"""
+This training script updates NanoGPT to run with either TP, PP, or TP+PP (2D).
+Usage:
+gpurun4 torchrun --nproc-per-node 4 2d_train.py 
+"""
+
+import argparse
+import math
+import os
+import time
+
+import numpy as np
+import torch
+import torch.distributed as dist
+import torch.nn as nn
+
+from model import GPT, GPTConfig
+
+from torch.distributed._tensor import DeviceMesh
+from torch.distributed.tensor.parallel import (
+    ColwiseParallel,
+    PairwiseParallel,
+    RowwiseParallel,
+    parallelize_module,
+)
+
+from torch.nn import functional as F
+
+def get_args():
+  # default config values designed to train a gpt2 (124M) on OpenWebText
+
+  def str_to_bool(v):
+    if isinstance(v, bool):
+        return v
+    if v.lower() in ('true', 't', '1'): 
+        return True
+    elif v.lower() in ('false', 'f', '0'):
+        return False
+    else:
+        raise ArgumentTypeError('Boolean expected.')
+
+  # I/O
+  parser = argparse.ArgumentParser()
+  parser.add_argument('--out_dir', type=str, default='out')
+  parser.add_argument('--eval_interval', type=int, default=2000)
+  parser.add_argument('--log_interval', type=int, default=2)
+  parser.add_argument('--eval_iters', type=int, default=200)
+  parser.add_argument('--eval_only', type=str_to_bool, default=False) # if True, script exits right after the first eval
+  parser.add_argument('--always_save_checkpoint', type=str_to_bool, default=True)  # if True, always save a checkpoint after each eval
+  parser.add_argument('--init_from', type=str, default="scratch")  # 'scratch', 'resume', 'gpt2*'
+  parser.add_argument('--train_iters', type=int, default=200000)
+  parser.add_argument('--seed', type=int, default=1337)
+
+  # data
+  parser.add_argument('--dataset', type=str, default="shakespeare_char") # "openwebtext"
+  parser.add_argument('--gradient_accumulation_steps', type=int, default=1) # used to simulate larger batch sizes
+  parser.add_argument('--batch_size', type=int, default=12)  # if gradient_accumulation_steps > 1, this is the micro-batch size
+  parser.add_argument('--block_size', type=int, default=1024)
+  parser.add_argument('--vocab_size', type=int, default=50304)
+
+  # model
+  parser.add_argument('--n_layer', type=int, default=12)
+  parser.add_argument('--n_head', type=int, default=12)
+  parser.add_argument('--n_embd', type=int, default=768)
+  parser.add_argument('--dropout', type=float, default=0.0)  # for pretraining 0 is good, for finetuning try 0.1+
+  parser.add_argument('--bias', type=str_to_bool, default=False) 
+
+  # adamw optimizer
+  parser.add_argument('--learning_rate', type=float, default=4e-4)  # max learning rate
+  parser.add_argument('--max_iters', type=int, default= 600000)  # total number of training iterations
+  parser.add_argument('--weight_decay', type=float, default=1e-2)
+  parser.add_argument('--beta1', type=float, default=0.9)
+  parser.add_argument('--beta2', type=float, default=0.95)
+  parser.add_argument('--grad_clip', type=float, default=1.0) # clip gradients at this value, or disable if == 0.0
+  parser.add_argument('--decay_lr', type=str_to_bool, default=True) # whether to decay the learning rate
+  parser.add_argument('--warmup_iters', type=int, default=2000)  
+  parser.add_argument('--lr_decay_iters', type=int, default=600000) 
+  parser.add_argument('--min_lr', type=float, default= 6e-5)  # minimum learning rate 
+  
+  # distributed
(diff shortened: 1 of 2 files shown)


examples/selective2d/2d_train.py
```diff
@@ -0,0 +1,354 @@
...
+  
+  print(f'[Rank{_rank}] {stage.submod.print_readable()}')
+  return model, stage 
+
+def pp_and_tp(model, mesh, args):
+  from pippy.compile import compile_stage
+  from pippy.microbatch import TensorChunkSpec, sum_reducer
+
+  pp_dim, tp_dim = 0, 1
+  pp_rank, tp_rank = args.local_rank // args.tp_size, args.local_rank % args.tp_size
```
fduwjj, line 154: What if we are doing this on multiple host?

examples/selective2d/model.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def forward(self, input):
+        return F.layer_norm(input, self.weight.shape, self.weight, self.bias, 1e-5)
+
+
+class CausalSelfAttention(nn.Module):
+
+    def __init__(self, mesh, config):
+        super().__init__()
+        assert config.n_embd % config.n_head == 0
```
fduwjj, line 45: Let's also add an assert for self.n_head % tp_size == 0?

examples/selective2d/2d_train.py
```diff
@@ -0,0 +1,354 @@
...
+  for i in range(args.n_layer):
+    name = f'transformer.h.{i}'
+    att = tp_attention(model, f'{name}.attn', mesh, tp_dim)
+    mlp = tp_mlp(model, f'{name}', mesh, tp_dim)
+
+  X, Y = get_rand(args)
+
+  # PP
+  cut_fn(model, args, args.pp_size)
+  stage = compile_stage(model, pp_rank, args.world_size, args.pp_size, args.device, pp_groups,
```
fduwjj, line 210: For my learning purpose, does it work if we first do TP and then call `compile_stage`? So DTensor can already be traced by torch.fx?

fduwjj (commented):
Thanks for making all this happening!!

examples/selective2d/2d_train.py
moonbucks: for now we apply TP first and then PP -->  so Dtensor is traced by torch.fx
if we do PP first and TP, the result does not change but PP changes layer name after tracing 
(e.g., transformer.block.i.attn --> transformer_block_i_attn) so we should change the name that we pass to TP.

examples/selective2d/2d_train.py
moonbucks: in multiple host case, we should use args.rank instead of args.local_rank. i'll change the line to handle multiple hosts in the next commit

examples/selective2d/model.py
moonbucks: okay!

examples/selective2d/2d_train.py
fduwjj: I see. Good to know that DTensor is traceable by torch.fx. I have a n00b question here, what's the difference between first applying TP vs first applying PP?

moonbucks: performance-wise, there is no difference. i think only difference is api. pippy's compile stage breaks higher class (e.g., block/transformer/MLP/Attention) into low-level layers (linear) so we just need to be careful of layer name.

examples/selective2d/2d_train.py
```diff
@@ -0,0 +1,354 @@
...
+                                     'mlp': PairwiseParallel()},
+                                     tp_mesh_dim=tp_dim) 
+
+  return model
+
+def pp(model, pp_device_mesh, args):
+  from pippy.IR import annotate_split_points, PipeSplitWrapper 
+  from pippy import split_into_equal_size
+  from pippy.compile import compile_stage
+  from pippy.microbatch import TensorChunkSpec, sum_reducer
```
fegin, line 136: Import should always live in the top of the file.

examples/selective2d/2d_train.py
```diff
@@ -0,0 +1,354 @@
...
+  X, Y = get_rand(args)
+
+  # PP
+  stage = compile_stage(model, pp_rank, args.world_size, args.pp_size, args.device, pp_groups, 
+            example_inputs=[X, Y],
+            )
+  
+  return model, stage
+
+def even_cut(model, args, pp_size, cut={}):
```
fegin, line 169: Should add a one line docstring to describe the function.

examples/selective2d/2d_train.py
```diff
@@ -0,0 +1,354 @@
...
+  from pippy.IR import annotate_split_points, PipeSplitWrapper 
+  cutpoint = args.n_layer // pp_size
+  for i in range(args.n_layer):
+    name = f'transformer.h.{i}'
+    if i > 0 and i % cutpoint == 0:
+      cut[name] = PipeSplitWrapper.SplitPoint.BEGINNING # or END
+
+  annotate_split_points(model, cut)
+
+def after_ar_cut(model, args, pp_size, cut={}):
```
fegin, line 179: Should add a one line docstring to describe the function.

examples/selective2d/2d_train.py
```diff
@@ -0,0 +1,354 @@
...
+  from pippy.IR import annotate_split_points, PipeSplitWrapper 
+  cutpoint = args.n_layer // pp_size
+  for i in range(args.n_layer):
+    name = f'transformer.h.{i}'
+    if i != args.n_layer - 1 and i % cutpoint == cutpoint - 1:
+      cut[f'{name}.mlp.dropout'] = PipeSplitWrapper.SplitPoint.BEGINNING
+
+  annotate_split_points(model, cut)
+
+def pp_and_tp_fg(model, mesh, args, tp_attn_layers=None, tp_mlp_layers=None, cut_fn=even_cut):
```
fegin, line 189: The naming is pretty confusing. What is fg?

fegin: Should add a one line docstring to describe the function.

fegin (commented):
Can you format the file by using `ufmt`?

examples/selective2d/2d_train.py
```diff
@@ -0,0 +1,354 @@
...
+                                     'attn.v': ColwiseParallel(),
+                                     'attn.c_proj': RowwiseParallel(),
+                                     'mlp': PairwiseParallel()},
+                                     tp_mesh_dim=tp_dim) 
+
+  return model
+
+def pp(model, pp_device_mesh, args):
+  from pippy.IR import annotate_split_points, PipeSplitWrapper 
+  from pippy import split_into_equal_size
```
fegin, line 134: Any reason why this is imported but not used?

examples/selective2d/2d_train.py
```diff
@@ -0,0 +1,354 @@
...
+
+  # PP
+  stage = compile_stage(model, pp_rank, args.world_size, args.pp_size, args.device, pp_groups, 
+            example_inputs=[X, Y],
+            )
+  
+  return model, stage
+
+def even_cut(model, args, pp_size, cut={}):
+  from pippy.IR import annotate_split_points, PipeSplitWrapper
```
fegin, line 170: Move the import to the top of the file.

examples/selective2d/2d_train.py
fegin, line 169: Why `cut` is passed as an argument? And using `{}` as the default value is never good.

examples/selective2d/2d_train.py
```diff
@@ -0,0 +1,354 @@
...
+      optimizer.step()
+      t1 = time.perf_counter()
+      dt = t1 - t0
+      local_iter_num += 1
+      iter_time += dt
+
+  return local_iter_num, iter_time
+  
+def tp_train():
+  local_iter_num = 0
```
fegin, line 267: The inconsistent indentation is never good.

examples/selective2d/2d_train.py
moonbucks: My bad. I used in my draft and moved on to my own cut. Will remove.

examples/selective2d/2d_train.py
moonbucks: I was thinking mixing two algorithms but at this point we may not need it. removed :)

examples/selective2d/2d_train.py
moonbucks: resolved

fegin (approved):

PR #831, opened 2023-07-06
mnist example update to new compile stage API

eddogola:
## Description

Update the mnist example to use the new compile_stage api

## Checklist:


eddogola:
https://www.internalfb.com/phabricator/paste/view/P782391812

eddogola:
See new error here: https://www.internalfb.com/phabricator/paste/view/P784276221

eddogola:
I'm currently saving the ckpts for each epoch in its own directory s.t. if were saving for every 500 epochs, there is a dir under "checkpoints"(or whatever other string that may be set as the checkpoint_dir) called 500, another 1000, and so on. What do you think of this design choice?

wz337:
Could you double check the files changed to make sure that the change is reflected correctly?

wz337 (approved):
LGTM!

PR #835, opened 2023-07-10
Fix generation of loss spec from output spec

kwen2501:
## Description

In PiPPy we automatically generate loss spec from output chunk spec.
Terminology:
"loss spec": a one-hot map indicating which output value is a loss. It consists of True, False values.
"output chunk spec": a data structure corresponding to output format describing which (chunked) value should be merged and which should be reduced (such as loss).

## Issue

In previous code, we only considered the case where the output chunk spec is a dictionary.
But in cases such as `return logits, loss`, the output chunk spec is a tuple, i.e. `(TensorChunkSpec(0), sum_reducer)`.

## Fix

Use `fx.node.map_aggregate` to generalize the auto generation.
It takes a lambda function that outputs True/False.

## Testing

torchrun --nproc-per-node 4 local_test_chunkspec.py

### DIFF
--- a/pippy/compile.py
+++ b/pippy/compile.py
@@ -81,10 +81,10 @@ def _compile(
 
     # Figure out which output is loss from output_chunk_spec
     output_loss_value_spec: Any = None
-    if isinstance(output_chunk_spec, dict):
-        output_loss_value_spec = {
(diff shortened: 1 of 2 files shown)


yjhao:
Thanks for fixing this!

yjhao (approved):

PR #847, opened 2023-07-30
profiler for example selective-tp

moonbucks:
## Type of change

profiler added

## Checklist:

### DIFF
--- a/examples/selective2d/2d_train.py
+++ b/examples/selective2d/2d_train.py
@@ -25,7 +25,7 @@ from torch.distributed.tensor.parallel import (
     RowwiseParallel,
 )
 
-from torch.profiler import profile, ProfilerActivity, record_function
+from torch.profiler import profile, ProfilerActivity
 
 
 def get_args():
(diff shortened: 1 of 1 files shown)


examples/selective2d/2d_train.py
```diff
@@ -328,21 +331,30 @@ def pp_tp_train(stage, mesh, args):
...
+        ),
+    ) as prof:
+        while local_iter_num < train_iters:
+            optimizer.zero_grad()
+            t0 = time.perf_counter()
+            X, Y = get_rand(args)
+            if pp_rank == 0:
+                out = stage(X)
+            elif pp_rank == args.pp_size - 1:
+                out = stage(Y)
```
fduwjj, line 347: does this mean we have two pipeline here?

fduwjj (approved):
LGTM

fduwjj:
Let's also make sure Lint is happy before merge, thanks!

examples/selective2d/2d_train.py
moonbucks: we have args.pp_size stages but we only pass 'input' to the first stage and 'label' to the last stage (to calculate loss and backpropagation). intermediate values are automatically passed by pippy

PR #849, opened 2023-08-01
Separate compute streams in forward path

moonbucks:
## Description

Separate streams in forward path and allows concurrent execution

## Type of change

- [v] Bug fix (non-breaking change which fixes an issue)
- [v] New feature (non-breaking change which adds functionality)

Please describe the Unit or Integration tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [v] Test LocalTestForwardTest
equivalence test passed 196872352.0 ref 196872352.0
profiling run completed 196872352.0 ref 196872352.0

- [v] Test LocalTestTwostreamTest
passed

- [v] Test LocalTestForwardBackwardTest
equivalence test passed 2.833994441818112e+16 ref 2.833994227069747e+16
Gradient equivalence test passed
correctness checks with original module passed

## Checklist:

- [v] Have you added tests that prove your fix is effective or that this feature works?

### DIFF
--- a/pippy/PipelineStage.py
+++ b/pippy/PipelineStage.py
@@ -70,6 +70,9 @@ class PipelineStage(torch.nn.Module):
         self.kwargs_chunk_spec = kwargs_chunk_spec
         self.output_chunk_spec = output_chunk_spec
 
+        self.even = torch.cuda.Stream()
+        self.odd = torch.cuda.Stream()
+
         # Find my submodule
         self.split_gm = self.pipe.split_gm
         named_children = list(self.split_gm.named_children())
@@ -606,16 +609,23 @@ class PipelineStage(torch.nn.Module):
         # Activation send requests of all chunk
         all_send_reqs: List[dist.Work] = []
 
-        output_chunks = []
+        output_chunks = [None] * self.chunks
 
         # Forward pass of all chunks
         for chunk in range(self.chunks):
-            output, send_reqs = self.forward_one_chunk(
-                chunk, args_split, kwargs_split, fwd_cache
-            )
-            all_send_reqs += send_reqs
-            # Prepare for final output merge or reduction
-            output_chunks.append(output)
+            if chunk % 2 == 0:
+              s = self.even
+            else:
+              s = self.odd
+
+            with torch.cuda.stream(s):
+              output, send_reqs = self.forward_one_chunk(
+                  chunk, args_split, kwargs_split, fwd_cache
+              )
+              all_send_reqs += send_reqs
+              # Prepare for final output merge or reduction
+              output_chunks[chunk] = output
+
 
         # Wait for all sends to finish
         # TODO: okay to delay the sync till completion of all chunks?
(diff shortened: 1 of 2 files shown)


pippy/PipelineStage.py
```diff
@@ -70,6 +70,9 @@ def __init__(
         self.kwargs_chunk_spec = kwargs_chunk_spec
         self.output_chunk_spec = output_chunk_spec
 
+        self.even = torch.cuda.Stream()
+        self.odd = torch.cuda.Stream()
```
fduwjj, line 74: instead of hard-code even and odd, would it be possible that we let user to set number of streams to use and we have a list of streams?

test/local_test_twostream.py
```diff
@@ -0,0 +1,186 @@
...
+        x = self.mlp1(x)
+        pipe_split()
+        x = self.mlp2(x)
+        pipe_split()
+        x = self.mlp3(x)
+        loss = self.mse_loss(x, target)
+        return {"logits": x, "loss": loss}
+
+
+class ExampleCodeRef(torch.nn.Module):
```
fduwjj, line 53: Looks like this class is not even used?

moonbucks: oh.. I added it for initial test but forgot to remove it. fixed in the followed commit

pippy/PipelineStage.py
moonbucks: we can make it to have a stream for each chunk. please see the following commit

test/local_test_twostream.py
```diff
@@ -0,0 +1,164 @@
...
+
+    example_input = torch.randn(
+        args.batch_size, args.hidden_size, device=args.device
+    )
+    example_target = torch.randn(
+        args.batch_size, args.hidden_size, device=args.device
+    )
+
+    model.to(args.device)
+    ref_out = model(example_input, example_target)
```
fduwjj, line 121: not a blocker, but we have not tested bwd yet right?

pippy/PipelineStage.py
```diff
@@ -70,6 +70,10 @@ def __init__(
         self.kwargs_chunk_spec = kwargs_chunk_spec
         self.output_chunk_spec = output_chunk_spec
 
+        self.streams = []
+        for i in range(chunks):
+            self.streams.append(torch.cuda.Stream())
```
fduwjj, line 75: this looks good to me, but originally I mean we can let users set how many streams they want to use.  We can let them set 2/3 or maybe like what you are doing here # of stages. Basically this number falls in the range of [2, # of stages]. But we can address this in a separate PR.

fduwjj (approved):
unblock and please make sure CI passed.

test/local_test_twostream.py
moonbucks: pippy originally has backward test and this modification passed the backward test. This feature didn't modify the behavior ( stream or synchronization) of backward path so I didn't add the test.

pippy/PipelineStage.py
moonbucks: resolved in a subsequent PR: https://github.com/pytorch/PiPPy/pull/852

PR #858, opened 2023-08-05
inner cache implemented

moonbucks:
## Description

instead of direct passing, use internal cache

## Type of change

- [v] New feature (non-breaking change which adds functionality)

## Checklist:

- [v] Have you added tests that prove your fix is effective or that this feature works?


pippy/PipelineStage.py
```diff
@@ -85,9 +85,12 @@ def __init__(
             self.global_depth = nstages
             self.inner_depth = 1
 
+        self.pipe_cache = [{} for i in range(chunks)] # [chunk][inner_rank] = value
+
         # Find my submodule
         self.split_gm = self.pipe.split_gm
         named_children = list(self.split_gm.named_children())
+
```
fduwjj, line 93: is this still needed?

moonbucks: @fduwjj  comment is left on blank line. if it indicate L91-L92 or L95, we will leave it for safety reason until we modify all functions to use 'list of submods' instead of one submod.

moonbucks:
code rebased

pippy/PipelineStage.py
fduwjj: I am indeed asking about L95.

pippy/PipelineStage.py
```diff
@@ -709,16 +814,28 @@ def forward(self, *args, **kwargs):
...
+            # Forward pass of all chunks
+            for chunk in range(self.chunks):
+                s = self.streams[chunk % self.nstreams]
+                with torch.cuda.stream(s):
+                    output, send_reqs = self.forward_one_chunk(
+                        chunk, args_split, kwargs_split, fwd_cache
+                    )
+                    all_send_reqs += send_reqs
+                    # Prepare for final output merge or reduction
+                    output_chunks[chunk] = output
```
fduwjj, line 838: The code has lots of duplication here, can you kindly consolidate a little bit?

pippy/PipelineStage.py
```diff
@@ -653,6 +671,93 @@ def forward_one_chunk(
...
         return output, send_reqs
 
+    def forward_one_chunk_ipipe(
+        self,
+        chunk: int,
+        args_split,
+        kwargs_split,
+        fwd_cache: Dict[int, Any],
+    ):
+        if self.rank == self.nstages - 1:
```
fduwjj, line 681: If I understand correctly, `self.nstages` is the global stages?

moonbucks: ```self.nstages``` has been used in original codebase so i added a new variable ```self.global_depth``` to store global stages. so ```self.nstages``` is same as pp_group_size. (# of ranks)

pippy/PipelineStage.py
```diff
@@ -653,6 +671,93 @@ def forward_one_chunk(
...
+
+        composite_args, composite_kwargs = self._recv_and_fill_inputs(
+            chunk,
+            args_split,
+            kwargs_split,
+        )
+
+        try:
+            if self.rank == self.nstages - 1:  # last stage
+                output = self.forward_maybe_with_nosync(
```
fduwjj, line 696: Do you need a inner_depth == 0 check here?

moonbucks: no we don't. inner_depth == 1 is default (no inner pipelining) and when inner_depth == 1, this function is not called. Instead, it calls forward_one_chunk function.

pippy/PipelineStage.py
moonbucks: cleaned in the subsequent commit

fduwjj (approved):
LGTM, maybe we can consolidate the logic and extrac common logic to make the code look cleaner?

PR #860, opened 2023-08-07
Interleaved 1F1B

kwen2501:
Add interleaved 1F1B. 

To do that:
- We modularize code into `forward_one_chunk` and `backward_one_chunk`


kwen2501:
Implemented. Closing.

PR #875, opened 2023-11-29
Migrate GPT-2 to new tracer

kwen2501:
## Description
Migrated GPT-2 example to work with new tracer based pippy.

examples/hf/hf_utils.py contains utility to generate inputs for HuggingFace models.

Model architecture:
```
GPT2ForSequenceClassification(
  (transformer): GPT2Model(
    (wte): Embedding(50257, 768)
    (wpe): Embedding(1024, 768)
    (drop): Dropout(p=0.1, inplace=False)
    (h): ModuleList(
      (0-11): 12 x GPT2Block(
        (ln_1): LayerNorm((768,), eps=1e-05, elementwise_affine=True)
        (attn): GPT2Attention(
          (c_attn): Conv1D()
          (c_proj): Conv1D()
          (attn_dropout): Dropout(p=0.1, inplace=False)
          (resid_dropout): Dropout(p=0.1, inplace=False)
        )
        (ln_2): LayerNorm((768,), eps=1e-05, elementwise_affine=True)
        (mlp): GPT2MLP(
          (c_fc): Conv1D()
          (c_proj): Conv1D()
          (act): NewGELUActivation()
          (dropout): Dropout(p=0.1, inplace=False)
        )
      )
    )
    (ln_f): LayerNorm((768,), eps=1e-05, elementwise_affine=True)
  )
  (score): Linear(in_features=768, out_features=2, bias=False)
)
```

## Run
```
$ torchrun --nproc-per-node 4 pippy_gpt2.py
```

## Output
https://gist.github.com/kwen2501/b9ed6158d8d0dc90b16824aa6abd8d72


examples/hf/pippy_gpt2.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def add_split_points(gpt2, nranks):
+    decoders_per_rank = (gpt2.config.n_layer + nranks - 1) // nranks
+    print(f"decoders_per_rank = {decoders_per_rank}")
+    nstages = 1
+    for i in range(1, gpt2.config.n_layer // decoders_per_rank):
+        annotate_split_points(
+            gpt2,
+            {f'transformer.h.{i * decoders_per_rank}': PipeSplitWrapper.SplitPoint.BEGINNING},
```
xw285cornell, line 32: Does this mean the split point is right after this submodule (specified by FQN)?

kwen2501: `BEGINNING` means right before; `END` would mean right after.

kwen2501: Annotation occurs before tracing. So the module structure is still there at that time.

H-Huang: How come there is no corresponding `END`?

PR #876, opened 2023-11-29
Add pipeline schedules script

H-Huang:
Adding `PipelineStage` and pipeline schedules to source control under PiPPy for the OSS use. File can be run as-is with `torchrun PipelineScheule.py` or with torchrun

Currently it is a standalone file but will be split out into different components


kwen2501:
Quick comment regarding breaking up components:
the main function can go into the test dir, the pipeline components can go into the pippy dir.

kwen2501:
Do you mind doing the break-up in this PR if it isn't too much work?

pippy/PipelineSchedule.py
```diff
@@ -0,0 +1,676 @@
...
+            pipeline = PipelineScheduleLoopedDFS(
+                stage_model_looped,
+                n_microbatch=n_microbatches,
+                pp_id=rank,
+                n_pp=n_pp,
+            )
+
+        logger.info(f"====== Rank {rank} profile ======")
+
+        # with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA]) as prof:
```
lessw2020, line 587: future item - maybe we can switch this to a context manager that is null context when not profiling, to avoid having profiler hanging around when not in use.

kwen2501:
For code quality checks:
You can first run `./format.sh`
Then run `./check.sh` to address non-format issue.

H-Huang:
> Do you mind doing the break-up in this PR if it isn't too much work?

Thanks @kwen2501, will do!

pippy/PipelineSchedule.py
H-Huang: Makes sense, will update that.

Also hope splitting up the file is okay for the IBM folks to use, if not please direct them to this version of the single file

https://github.com/pytorch/PiPPy/blob/6c8163ae936792b8ce3698a71934e3ec0daaa9e4/pippy/PipelineSchedule.py

pippy/PipelineSchedule.py
```diff
@@ -0,0 +1,460 @@
...
+    def forward(self, input_data, is_first_mb, is_last_mb):
+        logger.info(
+            f"[{self.rank} FORWARD {self.stage_id}] is_first_mb {is_first_mb} is_last_mb {is_last_mb}"
+        )
+        if self.is_first_stage:
+            self.fwd_input = input_data
+
+        # this is needed when we access the gradients for this in backward()
+        self.fwd_input.requires_grad = True
+        self.fwd_input.retain_grad()
```
kwen2501, line 102: Does this depend on the stage? i.e. would the first stage's input require grad?

pippy/PipelineSchedule.py
```diff
@@ -0,0 +1,460 @@
...
+    def get_bwd_send_ops(self) -> List[dist.P2POp]:
+        if self.is_first_stage:
+            return []
+        assert self.fwd_input.grad is not None, "grad must be valid"
+        return [dist.P2POp(dist.isend, self.fwd_input.grad, self.prev_stage)]
+
+    def get_bwd_recv_ops(self) -> Optional[dist.P2POp]:
+        if self.is_last_stage:
+            return []
+        return [dist.P2POp(dist.irecv, self.fwd_output_grads, self.next_stage)]
```
kwen2501, line 127: Mismatch between return typing and actual return?

pippy/PipelineSchedule.py
```diff
@@ -0,0 +1,460 @@
...
+
+    def compute_loss(self):
+        if self.fwd_output is None:
+            raise RuntimeError("forward() must be called before compute_loss()")
+        # TODO: use a real loss function passed in
+        return self.fwd_output.mean()
+
+
+class PipelineScheduleGPipe:
+    def __init__(self, stage):
```
kwen2501, line 170: typing stage?

pippy/PipelineSchedule.py
```diff
@@ -0,0 +1,460 @@
...
+        if self.is_first_stage:
+            return []
+        return [dist.P2POp(dist.irecv, self.fwd_input, self.prev_stage)]
+
+    def get_fwd_send_ops(self) -> List[dist.P2POp]:
+        if self.is_last_stage:
+            return []
+        return [dist.P2POp(dist.isend, self.fwd_output, self.next_stage)]
+
+    def forward(self, input_data, is_first_mb, is_last_mb):
```
kwen2501, line 93: `is_first_mb` & `is_last_mb` used only for logging?

pippy/PipelineSchedule.py
```diff
@@ -0,0 +1,460 @@
...
+class PipelineScheduleLoopedDFS:
+    def __init__(self, stages, n_microbatch, pp_id, n_pp):
+        assert (
+            n_microbatch % n_pp == 0
+        ), f"Looped DFS schedule requires microbatch_size ({n_microbatch}) to be a multiple of n_pp ({n_pp})"
+
+        self.stages = stages
+        self.n_microbatch = n_microbatch
+
+        self.n_stages = len(stages)
```
kwen2501, line 258: Does `n_stages` mean `n_local_stages`?

pippy/PipelineSchedule.py
```diff
@@ -0,0 +1,460 @@
...
+
+        self.stage_id_to_global_stage_id = [
+            (i * n_pp) + pp_id for i in range(self.n_stages)
+        ]
+
+        # pp_id is the same as local rank within the PP dimension
+        self.pp_id = pp_id
+
+        # number of sequences (chunks) to divide microbatches into == microbatch_size / (microbatch_size / n_pp)
+        self.seq_size = n_pp
```
kwen2501, line 271: I am a bit confused by this comment. What does seq_size refer to?

test/test_pipeline_schedule.py
```diff
@@ -0,0 +1,238 @@
...
+def setup(local_rank, world_size):
+    # If this is a child process (i.e., its PID is not the same as the PID of the process that started this script)
+    if os.getppid() != os.getpid():
+        set_up_logging(local_rank)
+
+    # initialize the process group
+    logger.info(f"init for rank {local_rank}")
+    dist.init_process_group("nccl")
+    if torch.distributed.is_initialized():
+        torch.cuda.set_device(local_rank)
```
kwen2501, line 78: Support CPU as well for our CI purpose?

test/test_pipeline_schedule.py
```diff
@@ -0,0 +1,238 @@
...
+    microbatch_size = 8
+    global_batch_size = 64
+    assert global_batch_size % microbatch_size == 0
+    n_microbatches = int(global_batch_size / microbatch_size)
+    n_pp = world_size
+
+    x = torch.randn([microbatch_size, input_dim]).to("meta")
+
+    stage_model = PipelineStage(
+        module_list[rank], rank, world_size, rank, world_size, x
```
kwen2501, line 113: QQ: why are `rank, world_size` provided twice?

kwen2501 (approved):
LGTM! Thanks for the implementations!

test/test_pipeline_schedule.py
```diff
@@ -0,0 +1,238 @@
...
+                n_microbatch=n_microbatches,
+                pp_id=rank,
+                n_pp=n_pp,
+            )
+
+        logger.info(f"====== Rank {rank} profile ======")
+
+        # with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA]) as prof:
+        #    with record_function(schedule):
+        pipeline.step(microbatches)
```
kwen2501, line 151: Curious: in reality/more formalized code, would every rank step on `microbatches` or just the first rank? Just to understand the API better.

pippy/PipelineSchedule.py
H-Huang: yep, only used for logging. Before it was used when communication was done within forward, but now communication (batch_isend_irecv) is outside in the schedule implementation so its not longer checked.

test/test_pipeline_schedule.py
H-Huang: I will clean this up, we expect each rank to call step but the microbatches argument is optional except for the first stage/rank which must pass in the microbatches

test/test_pipeline_schedule.py
H-Huang: The respective arguments are:

        rank -> stage_id: int,
        world_size -> num_stages: int,
        rank -> rank: int,
        world_size -> world_size: int,
        
For the nonlooped schedules the stage_id and num_stages are always equal to rank and world_size

pippy/PipelineSchedule.py
H-Huang: the comment looks old, sequence size should be world_size

PR #881, opened 2023-12-01
Migrate T5 example

kwen2501:
## Description

This example requires PyTorch PR https://github.com/pytorch/pytorch/pull/114982 to work, because stage 0 and stage 2 seem to be transmitting non-contiguous tensors.

## Test
https://gist.github.com/kwen2501/33fa5723496992691f8b1cc7daaadd89


examples/hf/pippy_t5.py
```diff
@@ -0,0 +1,129 @@
...
+
+from hf_utils import generate_inputs_for_model, get_number_of_params
+
+
+def add_split_points(t5, nranks):
+    # Number of encoder layers: t5.config.num_layers
+    # Number of decoder layers: t5.config.num_decoder_layers
+    # 6 encoder layers, 6 decoder layers, 12 layers in total
+    total_layers = t5.config.num_layers + t5.config.num_decoder_layers
+    layers_per_rank = (total_layers + nranks - 1) // nranks
```
H-Huang, line 26: what is this equation trying to express, is it the same thing as `math.ceil(total_layers / nranks)`?

H-Huang (approved):
Just curious, in the outpu:

```
Pipeline stage 0 25M params
Pipeline stage 1 9M params
Pipeline stage 2 29M params
Pipeline stage 3 29M params
```

how come stage 1 has so much fewer params?

examples/hf/pippy_t5.py
kwen2501: yes, ceil.

kwen2501:
> how come stage 1 has so much fewer params?
bc I did not pay attention to optimizing the split positions :)
A smarter split algorithm can help with that!

PR #899, opened 2023-12-05
Add torch profiling based on context manager for pipeline scheduler

lessw2020:
## Description
This PR adds a context based profiling manager (maybe_run_profiler) to the pipeline scheduler training loop. 
The idea of using a context manager is to ensure no profiler overhead if not profiling, while at the same time not requiring two different training loops (where one has a profiler, the other does not, resulting in duplicate code). 

Profiling is controlled via the --profiler True flag on launch.  
Traces are saved to the --trace_dir which now defaults to ./pipeline_traces
Updates to how the profiler is run, can be done in the context manager code block (i.e. adding warmup, etc). 

profiler_custom_handler

Side update is the multinode.slurm file is updated to launch the new name, test_pipeline_schedules.py. 
Fixes #(issue)

## Type of change

- [ X] New feature (non-breaking change which adds functionality)

Tested with and without running profiler on single node.  
Tested without profiler on multinode.

### DIFF
--- a/test/test_pipeline_schedule.py
+++ b/test/test_pipeline_schedule.py
@@ -79,7 +79,7 @@ def setup(local_rank, world_size):
 
 def main(**kwargs):
     torch.manual_seed(42)
-    print(f"MY KWARGS ARE {kwargs}")
+    
     rank = kwargs["rank"]
     local_rank = kwargs["local_rank"]
     world_size = kwargs["world_size"]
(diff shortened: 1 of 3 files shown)


test/test_pipeline_schedule.py
```diff
@@ -34,9 +35,42 @@
...
+            f"{trace_dir}/{schedule}_rank{rank}_trace.json"
+        )
+
+    if use_profiler:
+        with torch.profiler.profile(
+            activities=[
+                torch.profiler.ProfilerActivity.CPU,
+                torch.profiler.ProfilerActivity.CUDA,
+            ],
+            # schedule=torch.profiler.schedule(wait=1, warmup=2, active=3, repeat=1),
```
H-Huang, line 63: nice i didnt know about schedule `arg` in profile. Looks like this could be useful in the future if we wanted to profile for a particular step or even microbatch

H-Huang (approved):
Thanks for adding this!

test/test_pipeline_schedule.py
lessw2020: yes definitely should be useful as we expand out the profiling coverage.

PR #905, opened 2023-12-06
Mute an assert to coup with pytorch upstream change

kwen2501:
After certain torch 2.2.0.dev version, submod_0, submod_1, submod_2 ... are named as submod_0, submod_2, submod_4 ...

Muting this assert to let program pass.


lessw2020 (approved):
awesome, thanks for the fast fix. 
lgtm!

PR #914, opened 2023-12-12
Upgrade pipeline scheduler test to use device_mesh

lessw2020:
## Description
Moves the pipeline scheduler testing to using device_mesh, instead of using init_process_group and torch.cuda.set_device.

## Type of change

Verified on single node and multi-node.


lessw2020:
the lint issues are trivial but conflicts in style between ruff and the formatter here.

H-Huang (approved):
Woot, thanks!

PR #917, opened 2023-12-15
Add LLaMA example

kwen2501:
## Description

Apply PiPPy to LLaMA 2 with equal layer cutting.

We are getting the llama model from HuggingFace Model Hub:
https://huggingface.co/meta-llama/Llama-2-7b-chat-hf

You may need to request access to this model before running the example.

The prompt generation and decoding is are learned from @SunMarc

### DIFF
--- a/examples/llama/pippy_llama.py
+++ b/examples/llama/pippy_llama.py
@@ -1,0 +1,113 @@
+# Minimum effort to run this example:
+# $ torchrun --nproc-per-node 2 pippy_llama.py
+
+import argparse
+import os
+import time
+
+import torch
+import torch.distributed as dist
+
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from pippy.IR import Pipe, PipeSplitWrapper, annotate_split_points
+from pippy.PipelineStage import PipelineStage
+
+
+def add_split_points(llama, nranks):
+    layers_per_rank = (llama.config.num_hidden_layers + nranks - 1) // nranks
+    print(f"layers_per_rank = {layers_per_rank}")
+    nstages = 1
+    for i in range(1, nranks):
+        annotate_split_points(
+            llama,
+            {f'model.layers.{i * layers_per_rank}': PipeSplitWrapper.SplitPoint.BEGINNING},
+        )
+        nstages += 1
+    assert nstages == nranks, f"nstages = {nstages} nranks = {nranks}"
+
+
+def get_number_of_params(model):
+    return sum(p.numel() for p in model.parameters() if p.requires_grad)
+
+
+def run(args):
+    # Create a blank model
+    llama = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-2-7b-chat-hf", low_cpu_mem_usage=True)
+    tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-2-7b-chat-hf")
+
+    prompts = (
+        "How do you", "I like to", "Can I help", "You have to",
+        "The weather is", "I have a", "What is your", "You are a",
+    )  # bs = 8
+    tokenizer.pad_token = tokenizer.eos_token
+    inputs = tokenizer(prompts, return_tensors="pt", padding=True).to(args.device)
+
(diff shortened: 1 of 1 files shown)


examples/llama/pippy_llama.py
```diff
@@ -0,0 +1,109 @@
+# Minimum effort to run this example:
+# $ torchrun --nproc-per-node 2 pippy_llama.py
```
HamidShojanazeri, line 2:
```suggestion
# $ pip install transformers
# $ pip install accelerate
# $ torchrun --nproc-per-node 2 pippy_llama.py
```

examples/llama/pippy_llama.py
```diff
@@ -0,0 +1,109 @@
...
+    # Run
+    output = None
+    if args.rank == 0:
+        stage(inputs['input_ids'])
+    elif args.rank == args.world_size - 1:
+        output = stage()
+    else:
+        stage()
+
+    if output is not None:
```
HamidShojanazeri, line 78: @kwen2501 I wonder if we are using HF generate function under the hood/ we have our own implementation/ if we need any special treatment to work with HF generate function?

examples/llama/pippy_llama.py
```diff
@@ -0,0 +1,109 @@
...
+
+import torch
+import torch.distributed as dist
+
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from pippy.IR import Pipe, PipeSplitWrapper, annotate_split_points
+from pippy.PipelineStage import PipelineStage
+
+
```
HamidShojanazeri, line 15: it would be helpful if we add a comment here on our splitting points policy, how we are cutting the model/ how assigning devices to stages.

examples/llama/pippy_llama.py
```diff
@@ -0,0 +1,109 @@
...
+    tokenizer.pad_token = tokenizer.eos_token
+    inputs = tokenizer(prompts, return_tensors="pt", padding=True).to(args.device)
+
+    # Move model to `device` and set to evaluation
+    llama.to(args.device)
+    llama.eval()
+    print(llama)
+
+    # Annotate split points
+    add_split_points(llama, args.world_size)
```
HamidShojanazeri, line 48: I wonder if our auto_split policy is still applicable here?

examples/llama/pippy_llama.py
```diff
@@ -0,0 +1,109 @@
...
+
+    prompts = (
+        "How do you", "I like to", "Can I help", "You have to",
+        "The weather is", "I have a", "What is your", "You are a",
+    )  # bs = 8
+    tokenizer.pad_token = tokenizer.eos_token
+    inputs = tokenizer(prompts, return_tensors="pt", padding=True).to(args.device)
+
+    # Move model to `device` and set to evaluation
+    llama.to(args.device)
```
HamidShojanazeri, line 43: are we still able to do deferred init? as this would be a blocker for larger models.

examples/llama/pippy_llama.py
kwen2501: accelerate is not needed for this example. It is needed when you want to use it to automatically figure out the split points.

examples/llama/pippy_llama.py
kwen2501: No, we are not using HF's generate function here. Just a forward pass of the model.

examples/llama/pippy_llama.py
kwen2501: It might be. I'd need to check after previous migrations.

examples/llama/pippy_llama.py
kwen2501: That's in progress. I am trying to make things back to a working state after our migrations.

examples/llama/pippy_llama.py
kwen2501: Added the pip command.

examples/llama/pippy_llama.py
kwen2501: Done!

HamidShojanazeri (approved):

PR #920, opened 2023-12-19
Add example for unrolling iterative blocks

kwen2501:
## Description

Demonstrate PiPPy's functionality in unrolling iterative blocks.

For details, please see [README](https://github.com/pytorch/PiPPy/tree/unroll_example/examples/unrolling).

Many thanks to @mortzur 's inspiration!

### DIFF
--- a/examples/unrolling/README.md
+++ b/examples/unrolling/README.md
@@ -1,0 +1,93 @@
+## What does this example do?
+
+This is a synthetic example used to demonstrate PiPPy's functionality in unrolling iterative blocks in a model.
+
+We create a model that runs an iteration block in a for loop:
+```python
+class IterationBlock(torch.nn.Module):
+    def __init__(self, d_hid):
+        super().__init__()
+        self.lin = torch.nn.Linear(d_hid, d_hid)
+
+    def forward(self, x):
+        x = self.lin(x)
+        x = torch.relu(x)
+        return x
+
+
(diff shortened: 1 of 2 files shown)


examples/unrolling/README.md
```diff
@@ -0,0 +1,93 @@
...
+def forward(self, l_x_):
+    l__self___iter_block_mod_lin = self.L__self___iter_block_mod_lin(l_x_);  l_x_ = None
+    relu = torch.relu(l__self___iter_block_mod_lin);  l__self___iter_block_mod_lin = None
+    return relu
+```
+
+## How can this functionality help?
+Increase throughput of your model.
+
+Imagine your for loop needs to iterate on the data for `n` times, and it takes time `t` to process 1 sample (yielding a throughput of `1/t`). If we were to unroll the for loop onto `n` devices, then we can push `n` microbatches into the pipeline, each microbatch containing 1 sample. Then at any timeslot, the pipeline is processing `n` samples, yielding a throughput of `n/t`.
```
H-Huang, line 93: does this only work if the batch size can be split into `n` microbatches?

H-Huang (approved):
Very cool! And good, easy to understand example

examples/unrolling/README.md
kwen2501: That's right.

PR #922, opened 2023-12-21
[WIP] PyTorch native 2D LLaMA

kwen2501:
Status: 2D working without TP self attention
```
# PP = 2, TP = 4
$ torchrun --nproc-per-node 8 pippy_llama.py
['make', 'think', 'you', 'be', 'getting', 'great', 'favorite', 'right']
['make', 'think', 'you', 'be', 'getting', 'great', 'favorite', 'right']
['make', 'think', 'you', 'be', 'getting', 'great', 'favorite', 'right']
['make', 'think', 'you', 'be', 'getting', 'great', 'favorite', 'right']
```
so far so good.

TP self attention hitting the following issue:
```
view = l__self___model_layers_0_self_attn_q_proj.view(4, 4, 32, 128);  l__self___model_layers_0_self_attn_q_proj = None
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: shape '[4, 4, 32, 128]' is invalid for input of size 16384
```
4 * 4 * 32 * 128 = 65536
65536 / 4 = 16384 (4 is my TP size)
so that explains it.
User code:
```
xq = xq.view(bsz, seqlen, self.n_local_heads, self.head_dim)
```

Cc: @fduwjj @wanchaol @HamidShojanazeri @wconstab 

Can you shed light here?
I can modify the view node to change its arg, for example, 32 -> 8. That's slightly better than asking user to modify model code. But, is there an even better way?


examples/llama/2d_llama.py
```diff
@@ -0,0 +1,78 @@
...
+        #f"{layer_name}_self_attn_q_proj": ColwiseParallel(),
+        #f"{layer_name}_self_attn_k_proj": ColwiseParallel(),
+        #f"{layer_name}_self_attn_v_proj": ColwiseParallel(),
+        #f"{layer_name}_self_attn_o_proj": RowwiseParallel(),
+        f"{layer_name}_mlp_gate_proj": ColwiseParallel(),
+        f"{layer_name}_mlp_up_proj": ColwiseParallel(),
+        f"{layer_name}_mlp_down_proj": RowwiseParallel(),
+    })
+tp_mesh = mesh_2d["tp"]
+parallelize_module(stage.submod, tp_mesh, plan)
```
fduwjj, line 65: What you need to do for a submod is that:

self.n_local_heads = self.n_local_heads // tp_degree

kwen2501: Update: tried, but I cannot make such modification to model.
- Case 1: modify `num_heads` before tracing
Tracer will error out:
```
shape '[4, 4, 8, 128]' is invalid for input of size 65536
```
i.e. tracer seems to be very smart to do some size check.
- Case 2: modify `num_heads` after tracing
This won't work because the old value of `num_heads` has been burned into the generated program during tracing.

kwen2501: What works:
Modify the `view` and `reshape` ops' arguments after tracing
e.g. Modified "view" node's arg from:
(l__self___model_layers_0_self_attn_q_proj, 4, 4, 32, 128) to
(l__self___model_layers_0_self_attn_q_proj, 4, 4, 8, 128)
See the added util `modify_view`

examples/llama/2d_llama.py
```diff
@@ -0,0 +1,105 @@
+# $ torchrun --nproc-per-node 8 2d_llama.py
+import os
+import torch
+from transformers import AutoModelForCausalLM, AutoTokenizer
+from pippy import Pipe, PipeSplitWrapper, annotate_split_points, PipelineStage
+from torch.distributed._tensor import init_device_mesh
+
+
+# Utility
+def modify_view(
```
wanchaol, line 10: hmmmm this is a very dangerous operation, we shouldn't compose PP + TP like this IMO. In particular this would modify any view operations in the traced graph and thereby making the graph be super unsafe to the user, also this is not scalable to another models especially non-llama model might have view operations in non-attention layers, this would either trigger the assertion failure or wrongly modify the view ops.

wanchaol (changes requested):
Please see comments inlined. I don't think we should have this `modify_view` operation to compose TP and PP, it's very dangerous to users. I think we should probably not trace into the TranformerBlock and leave it for TP, or make TP happen first.

examples/llama/2d_llama.py
```diff
@@ -0,0 +1,105 @@
...
+rank = int(os.environ["RANK"])
+world_size = int(os.environ["WORLD_SIZE"])
+device = torch.device(f"cuda:{rank % torch.cuda.device_count()}")
+
+pp_group_size = 2
+tp_group_size = 4
+mesh_2d = init_device_mesh("cuda", (pp_group_size, tp_group_size), mesh_dim_names=("pp", "tp"))
+pp_group = mesh_2d["pp"].get_group()
+
+llama.to(device).eval()
```
HamidShojanazeri, line 49: it would be great to start envisioning the deferred init for the next step.

examples/llama/2d_llama.py
kwen2501: Thanks much @wanchaol for the comments. I agree with your thoughts. `modify_view` is more of a workaround than a general solution. For a better one, I wonder if the following direction is worth consideration. 
Today, for TP to work well with attention, view ops of user code needs to be modified. This is true even for eager mode. I wonder if it would be possible for the output of ColwiseParalell to play well with unchanged view ops. This may require: the output of ColwiseParalell to stay in some kind of DTensor form (rather than a regular tensor), and for this DTensor form to have special rule with the view ops (i.e. recognizing that some dimension of the view ops needs to be adjusted based on how the source operation is distributed).
Any thoughts here?
Cc @fduwjj

examples/llama/2d_llama.py
kwen2501: Agree. We are working towards that direction. See for example this PR and the BERT example in it.
https://github.com/pytorch/PiPPy/pull/923

HamidShojanazeri (approved):
Functionality wise is working for me and LGTM.

examples/llama/2d_llama.py
```diff
@@ -0,0 +1,105 @@
...
+            node.target == "view" or node.target == "reshape"
+        ):
+            assert len(node.args) >= 4
+            node.update_arg(3, node.args[3] // tp)
+    gm.recompile()
+
+
+# Grab the model
+llama = AutoModelForCausalLM.from_pretrained(
+    "meta-llama/Llama-2-7b-chat-hf", low_cpu_mem_usage=True
```
HamidShojanazeri, line 30:
```suggestion
    "meta-llama/Llama-2-7b-chat-hf", low_cpu_mem_usage=True, torch_dtype=torch.float16
```

examples/llama/2d_llama.py
kwen2501: Removed modify_view() now

examples/llama/2d_llama.py
```diff
@@ -0,0 +1,95 @@
...
+import torch
+from transformers import AutoModelForCausalLM, AutoTokenizer
+from pippy import Pipe, PipeSplitWrapper, annotate_split_points, PipelineStage
+from torch.distributed._tensor import init_device_mesh, DTensor
+from torch.distributed.tensor.parallel import parallelize_module, ColwiseParallel, RowwiseParallel
+
+
+# We set this flag to true to allow operations on a mix of tensor and dtensor
+# arguments. The mix is a result of `use_local_output=False`
+DTensor._op_dispatcher._allow_implicit_replication = True
```
wanchaol, line 12: I don't think we should set this flag, it's a hack and only suppose to be used by FSDP...

pippy/PipelineStage.py
```diff
@@ -476,7 +476,11 @@ def _send_activations(
                 )
                 peer_rank = self.stage_index_to_group_rank[dst]
                 work = dist.isend(
-                    out,
+                    # HACK: we convert DTensor to regular tensor here for it to
+                    # work with send ops. DTensor may show up in PP + TP cases.
+                    out.to_local()
+                    if isinstance(out, torch.distributed._tensor.DTensor)
```
wanchaol, line 482: I think we should understand why isend would get a DTensor, if Pipeline split each transformerblock it should not get DTensor as inputs

wanchaol:
I still think the cleanest fix here is to make PP tracing + `unflattener` work, otherwise we should probably wait for DTensor supports scaled dot product attention op instead, the current thing that `use_local_outputs` works surprised me, I think the only reason is that the llama 7B does not use scaled_dot_product_attention

kwen2501:
Documenting my discussion with @wanchaol wrt to DTensor and `scaled_dot_product_attention`:

@kwen2501 :
Should we do to_local as soon as we did colwise, or should we do to_local when we hit some op like scaled dot product, or should we have scaled dot product support a local form of DTensor. Maybe 2 and 3 are the same thing, meaning, the dispatcher of DTensor performs a to_local before calling the actual scaled dot product.

@wanchaol :
The current way is that we do to_local as soon as we leave the linear layer computation, this is the easiest thing to do with module forward hooks, if we do to_local as soon as we hit op like scaled dot product attention, I feel this is technically like implementing the scaled dot product attention op already. i.e. when implementing a DTensor op, we just figure out the sharding and then call local tensor with the op

@kwen2501 :
Now, in this case, the view ops are between colwise and scaled dot product. So it seems that the delayed route would work better. But i do agree that, if without the view ops, the early route would be easier. This means, the delayed route is a user choice (likely non-default), and we patch that route with DTensor support of scale dot product.

@wanchaol :
Yeah I think we should support both routes via use_local_output=False/True.
the delayed route require us to implement scaled dot product attention I think but it shouldn’t be too hard to enable it.

examples/llama/2d_llama.py
wconstab: what does it do?

PR #924, opened 2023-12-28
update import for _export_to_torch_ir

lessw2020:
PyTorch main had a refactor for torch export related code, which breaks tracing in IR.py with latest nightlies.
This PR simply updates the import for the _export_to_torch_ir function, in order to enable access again for tracing.

Link to refactor PR for reference:
https://github.com/pytorch/pytorch/pull/114768/files

Tested on bert and bert_cpu_init as quick test.

### DIFF
--- a/pippy/IR.py
+++ b/pippy/IR.py
@@ -9,6 +9,7 @@ import torch
 import torch.fx as fx
 from packaging import version
 from torch.export import Constraint
+from torch.export._trace import _export_to_torch_ir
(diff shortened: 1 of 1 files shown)


lessw2020:
the test failure seems to be the inverse of the original problem - the PyTorch build running CI does not have the refactor.

lessw2020:
same PR with but updated for backwards support with this PR:  https://github.com/pytorch/PiPPy/pull/926
Thus closing here.

PR #926, opened 2024-01-02
[copy] update import for _export_to_torch_ir

kwen2501:
## Description

This is a copy of #924 with an addition to support both new and old import paths.


lessw2020 (approved):
lgtm!

PR #927, opened 2024-01-02
Remove cap on torch version

kwen2501:
## Description

No more cap needed after #926


lessw2020 (approved):
looks good - great job re: remembering to update this.

Issue #928, opened 2024-01-17
Re-support kwargs at run time


kwen2501:
People would like to pass kwargs to model as in non-pipeline runs, instead of converting it to args tuple first.
See: https://github.com/huggingface/accelerate/pull/2345
```
args = (
    example_inputs["input_ids"].to("cuda:0"),
    example_inputs["decoder_input_ids"].to("cuda:0")
)
with torch.no_grad():
    output = model(*args)
```

PiPPy used to support kwargs before tracer migration. We'd need to add it back.

PR #929, opened 2024-01-19
Re-support kwargs at run time

kwen2501:
## Description

Implements #928 

Users want the first pipeline stage to accept kwargs if the original program does.
This is controlled by the `_codegen` field of the graph as @angelayi suggests, so we make a copy from the traced program to submod0. 

Added kwargs in test_fwd.py.
Also changed a few HF examples to directly kwargs.

### DIFF
--- a/pippy/IR.py
+++ b/pippy/IR.py
@@ -3,6 +3,7 @@ import copy
 import logging
 import operator
 from enum import Enum
+from inspect import Parameter, signature, Signature
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import torch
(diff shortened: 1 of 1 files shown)


pippy/IR.py
```diff
@@ -1005,6 +1005,21 @@ def move_param_to_callee(
...
+        # so we make a copy here. Note: we only want the input spec and not the
+        # output spec, because the output spec is for the last stage. Maybe a
+        # TODO? Not sure yet.
+        submod0 = list(split.children())[0]
+        submod0.graph._codegen = copy.deepcopy(traced.graph._codegen)
+        # `_replace` is actually not "private" or internal. based on this doc:
+        # To prevent conflicts with field names, the method and attribute names
+        # start with an underscore
+        submod0.graph._codegen.pytree_info = (
+            submod0.graph._codegen.pytree_info._replace(out_spec=None)
```
angelayi, line 1019: why do you want to set it to be None? Then the output structure will not match the eager module -- but not sure if you care about this.

angelayi (approved):

pippy/IR.py
kwen2501: Because the original program's output format should be applied to the last stage, not stage 0 :)

PR #933, opened 2024-01-24
Fill value for non-tensor inputs during shape prop

### LINKED ISSUE #932: Shape prop error when kwargs have constants
Say the forward function is as follows:
```
    def forward(
        self, x,
        c: bool=True,
        y=torch.zeros(batch_size, d_hid),
    ):
```
When calling `from_tracing` with `example_kwargs={"c": True, "y": y}`, the following error occurs:
```
  File "pippy/IR.py", line 1172, in from_tracing
    PipeFakeTensorProp(pipe.split_gm).run()
  File "pippy/IR.py", line 1287, in run
    inp = tuple(
          ^^^^^^
  File "pippy/IR.py", line 1288, in <genexpr>
    node.meta["val"]
    ~~~~~~~~~^^^^^^^
KeyError: 'val'
```
The reason is that for non-tensor inputs, node.meta is empty.

kwen2501:
Fixes #932

### DIFF
--- a/pippy/IR.py
+++ b/pippy/IR.py
@@ -1004,34 +1004,6 @@ class Pipe(QualnameMapMixin, torch.nn.Module):
 
         split.delete_all_unused_submodules()
 
-        # Users want the first pipeline stage to accept kwargs if the original
-        # program does. This is controlled by the `_codegen` field of the graph,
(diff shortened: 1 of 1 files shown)


H-Huang (approved):
Nice!

PR #934, opened 2024-01-25
Add back support for stage.remap_qualname()

kwen2501:
## Description
`stage.remap_qualname(key)` maps a stage's parameter name (`key`) back to the original model's parameter name.

This now works:

```
    # Stage module's state dict
    sd = stage.submod.state_dict()
    remapped_keys = [stage.remap_qualname(k) for k in sd.keys()]

    # Original model's state dict
    old_keys = mod.state_dict().keys()

    # Confirm they match
    assert all(rk in old_keys for rk in remapped_keys)
```

### DIFF
--- a/test/test_fwd.py
+++ b/test/test_fwd.py
@@ -83,6 +83,16 @@ def run_worker(args):
             f"equivalence test passed {torch.sum(out)} ref {torch.sum(ref_out)}"
         )
 
+    # Test qualname mapping
+    sd = stage.submod.state_dict()
+    print(f"Rank {args.rank} state dict keys: {sd.keys()}")
+    remapped_keys = [stage.remap_qualname(k) for k in sd.keys()]
+    print(f"Rank {args.rank} remapped keys: {remapped_keys}")
+    # Confirm remapped keys are consistent with original model
+    old_keys = mod.state_dict().keys()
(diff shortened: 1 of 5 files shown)


test/test_fwd.py
```diff
@@ -83,6 +83,16 @@ def run_worker(args):
...
         )
 
+    # Test qualname mapping
+    sd = stage.submod.state_dict()
+    print(f"Rank {args.rank} state dict keys: {sd.keys()}")
+    remapped_keys = [stage.remap_qualname(k) for k in sd.keys()]
+    print(f"Rank {args.rank} remapped keys: {remapped_keys}")
+    # Confirm remapped keys are consistent with original model
+    old_keys = mod.state_dict().keys()
+    assert all(rk in old_keys for rk in remapped_keys)
```
H-Huang, line 93: nit: wouldn't this evaluate to be `True` if remapped_keys was a subset of old_keys.

Maybe a better check would be sorted(remapped_keys) == sorted(list(old_keys))

H-Huang (approved):

test/test_fwd.py
kwen2501: `remapped_keys` is always a subset of `old_keys`, because `remapped_keys` are only keys of a stage, while `old_keys` contains keys of the full original model.

PR #937, opened 2024-01-26
Add pypi instructions

kwen2501:


H-Huang (approved):
I've never published a python package so can't comment on the current setup, but I have seen [poetry](https://python-poetry.org/) become very popular for building and publishing, so maybe it can be something to consider in the future

PR #938, opened 2024-01-31
Update manual PipelineStages constructor

H-Huang:
- update manual PipelineStages constructor to require meta inputs and outputs
- remove device in constructor and requires users to call .to on PipelineStages to materialize it
- change all logging to debug
- fix the test pipeline schedule script to adhere to these changes


pippy/PipelineSchedule.py
```diff
@@ -131,11 +121,51 @@ def __init__(
...
             finished pipeline stage init, {self.stage_id=}, {self.is_first_stage=},
             {self.is_last_stage=}, {self.num_stages=},
-            {[fwd_input.shape for fwd_input in self.fwd_inputs]},
-            {[fwd_output_grad.shape for fwd_output_grad in self.fwd_output_grads]}
             """
         )
 
+    def to(self, *args, **kwargs):
+        """
+        Move the module to a new device or data type, including the buffers.
```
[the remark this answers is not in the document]
H-Huang: I think 1 is a great idea, let me try that out.
For 2, i agree that have 1 umbrella utility to apply all the parallelized models to device would be better. I also don't think that the `.to()` is necessary but for now it is a good stopgap in order to explicitly declare when to materialize PP which I found helps for code simplification.

pippy/PipelineSchedule.py
```diff
@@ -131,11 +121,51 @@ def __init__(
...
+        """
+        super().to(*args, **kwargs)
+
+        # find the device of the underlying module and move the buffers to it if they are meta
+        device = next(self.module.parameters()).device
```
kwen2501, line 134: if the `to(...)` method already passes in a device argument, would this check be redundant (or potentially anti-parallel)?

kwen2501: And there is a chance that `self.module` does not contain any parameter.

pippy/PipelineSchedule.py
```diff
@@ -131,11 +121,51 @@ def __init__(
...
+
+    def create_buffers(
+        self, inputs_meta: List[torch.tensor]
+    ) -> List[torch.Tensor]:
+        """
+        Creates buffers for a given input on a specified device.
+        This function takes as input a meta tensor or a list of meta tensors
+        and returns a flattened list of empty tensors of the same shape.
+        """
+        if isinstance(inputs_meta, torch.Tensor):
```
kwen2501, line 157: It seems this `if` branch would never be true based on the signature of this function.

pippy/PipelineSchedule.py
```diff
@@ -131,11 +121,51 @@ def __init__(
...
+            return [inputs_meta]
+        elif isinstance(inputs_meta, (list, tuple)):
+            return [
+                item
+                for sublist in [self.create_buffers(inp) for inp in inputs_meta]
+                for item in sublist
+            ]
```
kwen2501, line 164: If `inputs_meta` is a list of tensors, would it already be in "flattened" form?

pippy/PipelineSchedule.py
```diff
@@ -98,27 +81,34 @@ def __init__(
         num_stages: int,
         rank: int,
         world_size: int,
-        input_args: List[torch.tensor],
-        device: torch.device,
+        inputs_meta: List[torch.tensor],
+        outputs_meta: Optional[List[torch.tensor]] = None,
```
kwen2501, line 85: Curious, what's the reason for asking for `outputs_meta`?

H-Huang: In some cases, I found it easier to pass in the output manually rather than having the model run the forward itself. So there is an option to pass the shape of the output rather than running forward

pippy/PipelineSchedule.py
```diff
@@ -98,27 +81,34 @@ def __init__(
...
     ):
         super().__init__()
+        self.module = module
         self.rank = rank
         self.stage_id = stage_id
         self.is_first_stage = stage_id == 0
         self.is_last_stage = stage_id == num_stages - 1
         self.num_stages = num_stages
-        # When we materialize the model partition on cuda, we call reset_parameters() if it is available
```
wconstab, line 110: i'm confused about the state of the code before your PR- is the expectation that inputs are real inputs, and model comes on meta device, hence we used to move it to 'real device' and then run real inputs through it?

it must have changed back and forth since the old code `meta_output = self.module(*input_args)` looks half way between each world.  it claims to give meta outputs, but it uses module that's moved to 'device'.

i think in your new PR you are trying to solidify that we use meta inputs and meta model for the PP stage __init__ lifetime, and then 'real' happens sometime later.  If so i agree with that move.

PR #944, opened 2024-02-14
Set up unit tests for PipelineSchedules

H-Huang:
- Moved e2e tests to a different file
- Set up unit tests to test smaller pieces (currently testing `init` and `forward` for varying inputs/outputs)
- Fix bugs in PipelineStage which prevent e2e from running

### DIFF
--- a/pippy/PipelineSchedule.py
+++ b/pippy/PipelineSchedule.py
@@ -3,7 +3,7 @@
 import logging
 from abc import ABC, abstractmethod
 from collections import deque
-from typing import Deque, List, Optional, Tuple, Union
+from typing import Any, Deque, List, Optional, Tuple, Union
 
 import torch
 import torch.distributed as dist
@@ -98,7 +98,7 @@ class PipelineStageV2Impl(PipelineStage):
         num_stages: int,
         rank: int,
         world_size: int,
-        input_args: List[torch.tensor],
+        input_args: Union[torch.Tensor, List[torch.tensor]],
         device: torch.device,
     ):
         super().__init__()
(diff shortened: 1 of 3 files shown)


test/test_pipeline_schedule.py
```diff
@@ -79,225 +21,58 @@ def __init__(
...
-    for i in range(n_microbatches):
-        microbatches.append(
-            (torch.randn_like(x_cuda_empty), torch.ones((1, 1), device="cuda"))
+            model_meta,
+            stage_id=0,
+            num_stages=1,
+            rank=0,
+            world_size=1,
+            input_args=[input_meta],
+            device="cuda",
```
wconstab, line 62: thinking out loud, if we use .cpu we can run testing in CI today. if we have to do a lot of stuff to make cpu/gloo work nicely with pipelinestage, this might be less appealing but.. worth a try?

pippy/PipelineSchedule.py
```diff
@@ -179,19 +179,20 @@ def get_fwd_send_ops(self) -> List[dist.P2POp]:
...
-            for tensor in self.fwd_inputs:
-                tensor.requires_grad = True
-                tensor.retain_grad()
+        for tensor in self.fwd_inputs:
+            tensor.requires_grad = True
+            tensor.retain_grad()
 
         # perform forward pass on module
-        self.fwd_outputs = self.module(*self.fwd_inputs)
+        self.fwd_outputs = [self.module(*self.fwd_inputs)]
```
wconstab, line 195: wondering why the addition of the []

(is this gonna fix one case and break another? its risky if we do not have a design written down. maybe add comment somewhere that describes our assumptions about saving and later using fwd_outputs)

pippy/PipelineSchedule.py
```diff
@@ -179,19 +179,20 @@ def get_fwd_send_ops(self) -> List[dist.P2POp]:
...
+        logger.info(
+            f"[{self.rank} FORWARD {self.stage_id} {[fwd_input.shape for fwd_input in self.fwd_inputs]}"
+        )
 
         # this is needed when we access the gradients for this in backward()
-        if not self.is_first_stage:
-            for tensor in self.fwd_inputs:
-                tensor.requires_grad = True
-                tensor.retain_grad()
+        for tensor in self.fwd_inputs:
```
wconstab, line 190: add a TODO here to make this actually depend on the requires_gradness of the actual fwd_inputs during shape inference?

test/test_pipeline_schedule.py
H-Huang: yeah, that makes sense. That would make it a lot easier to test, everything should be able to "just work" with gloo

pippy/PipelineSchedule.py
H-Huang: i missed this when I added the support for multiple arguments, so i was fixing it then. Each input microbatch is a list of tensors [arg1, arg2, ..], likewise, each pipeline stage output also needs to be a list of tensors [output1, output2, ...]

H-Huang: let me write a comment for this

test/test_pipeline_schedule.py
wconstab: hmm. even batch_isend_irecv?

H-Huang: oh yeah thats a good point. in gloo, batch isend_irecv is implemented multiples send/recv operations rather than a single operations as with nccl. So schedules like gpipe / looped bfs should still work, looped DFS we may see that hang due to conflicting ops that we saw a while back.

wconstab: if the op is "literally" supported by gloo, then you can use cpu to do unit tests of pipelinestage, etc.  But maybe not worth trying to get integration/e2e testing working if unreliable. we should get gpu machines.

pippy/PipelineSchedule.py
H-Huang: Thanks, I fixed this in another way thats a bit more robust rather than adding the [], this will throw an error if the model forward() output is something not supported.

wconstab (approved):

PR #951, opened 2024-02-17
add get_stage_shapes() and validate_stage_shapes() helpers

### LINKED ISSUE #946: Add lazy shape inference with buffer shape validator
"""
manually computing input sizes for each stage is tedious. doing shape inference inside init is also tedious since the 'inputs_meta' you pass in has to match the meta-ness (or cpu/cuda-ness) of the model you pass in. Can we just do shape inference lazily on first forward call?
"""

For PipelineStage it would be nice to have a feature where all send and recv buffers are calculated through an initial run, thus removing the need to manually write the inputs for each model chunk.

Furthermore, at execution time, the shapes of the inputs and outputs given by the list of microbatches can be validated against the shapes of the buffers. This would prevent hangs and crashes from happening at runtime due size mismatches of the inputs shapes and the buffers.

H-Huang:
fixes #946

Update and simplifies some variable naming in `PipelineStage`

Add two helper functions:
- add `get_stage_shapes` which  passes a single microbatch through all the model_chunks and returns the inputs and outputs for each model_chunk.
- add `validate_stage_shapes ` which validates that the input and output for each pipeline stage can be communicated and the neighboring stages shapes are valid.

To implement these I needed to create a predefined format that ranks could communicate with as they transfer shapes to each other. The data is of format `[num_dims, dim1, dim2, ...]`, for example if the inputs are `[torch.ones((4,5)), torch.zeros(2,4,1)]`, then the metadata tensor would look like `[2, 4, 5, 3, 2, 4, 1, 0, 0, ...]` (rest are 0s). See `create_metadata_tensor` and `extract_metadata_from_tensor` implementations.

### DIFF
--- a/pippy/PipelineSchedule.py
+++ b/pippy/PipelineSchedule.py
@@ -3,7 +3,7 @@
 import logging
 from abc import ABC, abstractmethod
 from collections import deque
-from typing import Any, Deque, List, Optional, Tuple, Union
+from typing import Any, Deque, Dict, List, Optional, Tuple, Union
 
 import torch
 import torch.distributed as dist
@@ -12,9 +12,9 @@ from torch.profiler import record_function
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
-handler = logging.StreamHandler()
-handler.setLevel(logging.INFO)
-logger.addHandler(handler)
+# handler = logging.StreamHandler()
+# handler.setLevel(logging.INFO)
+# logger.addHandler(handler)
 
 
 class PipelineStage(ABC, nn.Module):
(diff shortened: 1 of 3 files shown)


pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+        f"Unsupported input type {type(tensor)} cannot create buffers"
     )
 
 
+METADATA_TENSOR_LEN = 50
+PLACEHOLDER_VAL = 0
+
+
+def create_metadata_tensor(
+    tensors: Optional[List[torch.Tensor]] = None,
```
kwen2501, line 98: nit: a comment for what `tensors` refer to?

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+        return [torch.empty_like(tensor, device=device)]
+    elif isinstance(tensor, (list, tuple)):
+        return [torch.empty_like(t, device=device) for t in tensor]
+    raise TypeError(
+        f"Unsupported input type {type(tensor)} cannot create buffers"
     )
 
 
+METADATA_TENSOR_LEN = 50
+PLACEHOLDER_VAL = 0
```
kwen2501, line 94: nit: shall we reserve value 0 for scalar tensor? (Just in case)
```
>>> a = torch.tensor(1)
>>> a.shape
torch.Size([])
>>> len(a.shape)
0
```

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+    if isinstance(tensor, torch.Tensor):
+        return [torch.empty_like(tensor, device=device)]
+    elif isinstance(tensor, (list, tuple)):
+        return [torch.empty_like(t, device=device) for t in tensor]
+    raise TypeError(
+        f"Unsupported input type {type(tensor)} cannot create buffers"
     )
 
 
+METADATA_TENSOR_LEN = 50
```
kwen2501, line 93: nit: make this even bigger just to be safe?

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+    while i < len(tensor) and tensor[i] != PLACEHOLDER_VAL:
+        num_dims = tensor[i].item()
+        shape = tensor[i + 1 : i + 1 + num_dims].tolist()
+        metadata.append(shape)
+        i += num_dims + 1
+    return metadata
+
+
+def dry_run(
+    models: List[nn.Module],
```
kwen2501, line 149: nit: does `models: List[nn.Module]` here refer to a full model or the nn.Module's assigned to this rank?

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+
+        # perform forward
+        # TODO: if forward fails raise a more descriptive error explaining which stage failed
+        fwd_outputs = model(*example_fwd_inputs)
+        fwd_outputs = create_buffers(fwd_outputs, device)
+        shapes["outputs"] = [fwd_output.shape for fwd_output in fwd_outputs]
+
+        # send shape dims
+        if stage_id != num_stages - 1:
+            placeholder_shape_tensor = create_metadata_tensor(
```
kwen2501, line 202: nit: use a different name than `placeholder_shape_tensor`? 1. because it has been used; 2. because here we refer to outputs not inputs (i.e. placeholders)

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+                fwd_outputs, device=device
+            )
+            dist.send(placeholder_shape_tensor, next_rank)
+        stages[stage_id] = shapes
```
kwen2501, line 206: nit: use a different name than `stages`?

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+    Only the first stage must pass in a microbatch.
+
+    Each rank must call dry_run or the program will hang.
+
+    Returns a dictionary containing the following keys:
+        "inputs": Shape of the inputs to the module
+        "outputs": Shape of the outputs of the module
+    """
+
+    stages = {}
```
kwen2501, line 169: nit: type it?

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+            example_fwd_inputs = [
+                torch.empty(shape_list, device=device)
+                for shape_list in metadata
+            ]
+        shapes["inputs"] = [fwd_input.shape for fwd_input in example_fwd_inputs]
+
+        # perform forward
+        # TODO: if forward fails raise a more descriptive error explaining which stage failed
+        fwd_outputs = model(*example_fwd_inputs)
+        fwd_outputs = create_buffers(fwd_outputs, device)
```
kwen2501, line 197: nit: why are we creating output buffers here (and later in `PipelineStageV2Impl`)?

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+        if stage_id != num_stages - 1:
+            placeholder_shape_tensor = create_metadata_tensor(
+                fwd_outputs, device=device
+            )
+            dist.send(placeholder_shape_tensor, next_rank)
+        stages[stage_id] = shapes
+    return stages
+
+
+def validate_stages(pipeline_stages: List[PipelineStage]):
```
kwen2501, line 210: nit: `validate_stage_shapes`?

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+        # first stage doesn't receive anything and uses a microbatch
+        if stage_id == 0:
+            if microbatch is None:
+                raise RuntimeError("Microbatch is required for first stage")
+            example_fwd_inputs = microbatch
+            if isinstance(example_fwd_inputs, torch.Tensor):
+                example_fwd_inputs = [example_fwd_inputs]
+        else:
+            # other stages must receive shape information
+            dist.recv(placeholder_shape_tensor, prev_rank)
```
kwen2501, line 186: TODO: `send` `recv` calls should take a group, rather than assuming the world == pipeline dim.

pippy/PipelineSchedule.py
kwen2501: Is this call optional?

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+    Check that the buffer shapes match between stages was expected by performing an all_gather between
+    all stages. Assumes that buffers have been initialized already such that get_fwd_recv_ops() and
+    get_fwd_send_ops() return valid lists of p2p ops.
+    """
+    virtual_pipeline_size = len(pipeline_stages)
+    all_inputs = []
+    all_outputs = []
+    # perform all gathers between all stages
+    for virtual_id, stage in enumerate(pipeline_stages):
+        world_size = stage.world_size
```
kwen2501, line 221: nit: not related to this PR, maybe we can find a better name that `stage.world_size`? It could be confusing.

pippy/PipelineSchedule.py
```diff
@@ -74,22 +74,225 @@ def compute_loss(self):
...
+        # all gather each ranks inputs
+        tensor_list = [
+            create_metadata_tensor(device=stage.device)
+            for _ in range(stage.world_size)
+        ]
+        expected_inputs = [op.tensor for op in stage.get_fwd_recv_ops()]
+        stage_input = create_metadata_tensor(
+            expected_inputs, device=stage.device
+        )
+        dist.all_gather(tensor_list, stage_input)
```
kwen2501, line 245: Same, take a group.

pippy/PipelineSchedule.py
```diff
@@ -98,58 +301,52 @@ def __init__(
...
-            not isinstance(output, torch.Tensor) for output in meta_output
-        ):
-            raise ValueError(
-                f"All elements in module output must be torch.Tensor instances, but got {[type(x) for x in meta_output]}"
-            )
+        if output_args is None:
+            self.outputs = self.module(*self.inputs)
+            self.outputs = create_buffers(self.outputs, device)
+        else:
+            self.outputs = create_buffers(output_args, device)
```
kwen2501, line 329: nit: a comment for why we create `outputs` buffers?

pippy/PipelineSchedule.py
```diff
@@ -181,86 +378,89 @@ def get_fwd_recv_ops(self) -> List[dist.P2POp]:
...
-                dist.P2POp(dist.isend, fwd_output, self.next_stage)
-                for fwd_output in self.fwd_outputs
-            ]
+        return [
+            dist.P2POp(dist.isend, out, self.next_stage) for out in self.outputs
+        ]
         raise AssertionError("invalid fwd_outputs type, cannot send")
 
+    def check_and_format_outputs(self, outputs: Any) -> List[torch.tensor]:
+        # validate the output of the module is a type supported
```
kwen2501, line 396: nit: can you state the `type supported` here?

test/test_pipeline_schedule.py
```diff
@@ -140,16 +166,99 @@ def test_pipeline_stage_fwd(self):
...
+            init_method=self.init_method,
+            backend="gloo",
+            rank=self.rank,
+            world_size=self.world_size,
+        )
+
+        model_chunk = MLP(dim=8, hidden_dim=4, out_dim=8)
+        microbatch = torch.rand((10, 8), device=device)
+        # test single model
+        stages = dry_run(
```
kwen2501, line 198: nit: `stages` --> `stage_shapes`

test/test_pipeline_schedule.py
```diff
@@ -3,11 +3,13 @@
 import unittest
 
 import torch
+import torch.distributed as dist
 import torch.nn as nn
-from pippy.PipelineSchedule import PipelineStageV2Impl
+from pippy.PipelineSchedule import dry_run, PipelineStageV2Impl, validate_stages
```
kwen2501, line 8: nit: if `dry_run` is going to be a user-facing util, maybe give it a more distinguishing name? (because users can have their own `dry_run` functions referring to other things). In fact, it may be even worth considering `deviating` from `dry_run` (implementation detail) and just name it based on purpose (e.g. `get_stage_shapes`, etc)

kwen2501 (approved):
This is a great achievement and would simplify our manual UI a lot!
I put in some minor comments. 
In addition, it may be worth hardening more:
in the `dry_run` function, see if the `send`/`recv` calling would deadlock or not.

kwen2501:
Also want to note that for the dry-run, stage 0 is using real example inputs, while stage 1 and above will be using random inputs (generated from the shape meta).
@H-Huang @wconstab how can we characterize implication of this?

pippy/PipelineSchedule.py
H-Huang: This is a good edge case, i hadnt considered it.

I changed the placeholder value to -1 and also added a new test which considers this shape as well. Thanks!

pippy/PipelineSchedule.py
H-Huang: nn.Modules assigned to this rank, so the length should be 1 for any non-interleaved schedules and >1 for any interleaved schedules

pippy/PipelineSchedule.py
H-Huang: good point, i updated both of the variable names

pippy/PipelineSchedule.py
H-Huang: renamed to `stage_id_to_shapes`

pippy/PipelineSchedule.py
H-Huang: I need to rename this, this function is just formatting the argument into a list. More specifically its just turning a single output into a list of length 1, otherwise if its already a tuple or list then leaves it.

pippy/PipelineSchedule.py
H-Huang: yes the call is optional

pippy/PipelineSchedule.py
H-Huang: agreed, i will update it in the future when pipeline stages also takes a group, added a comment for now

test/test_pipeline_schedule.py
H-Huang: makes sense, renamed it

H-Huang:
> Also want to note that for the dry-run, stage 0 is using real example inputs, while stage 1 and above will be using random inputs (generated from the shape meta).

Thats a good point. I can actually modify it so it performs an additional send/recv for the actual data which would make more sense.

wconstab:
I didn't review too carefully and overall lgtm. But a few thoughts maybe for follow up 

- wonder if we should consider pickling a dict of metadata into a byte tensor just for flexibility. This could allow sending other info like requires_grad, dtype of tensors as well as shape. Seems easy to slot this in in your design.
- it's nice to log all-shapes to rank0. But in the future if we have better logger control over ranks outputs, I wonder if we should just log all info to all ranks. Possible upsides - seeing a discrepancy if there was one, having the info on hand if you're looking at rank1 logs instead of rank0 for other reasons

PR #956, opened 2024-02-28
Retire PipeSplitWrapper and support UNet tracing

kwen2501:
`PipeSplitWrapper` is our current way of adding split annotation. 

However, it adds a `self.mod` hierarchy (due to wrapping), which:
- causes trouble for tracing certain model (e.g. UNet directly access `up_block0.resnet` from the root's forward; if `up_block0` is turned into a `PipeSplitWrapper`, then resnet's FQN becomes `up_block0.mod.resnet`, making `up_block0.resnet` an "attribute not found".
- makes it harder to maintain FQN invariance which distributed checkpointing prefers.

This PR replaces the wrapping method with one that swaps the forward function of the wrapped module, so that `self.mod = wrapped_mod` is not needed.

### DIFF
--- a/pippy/IR.py
+++ b/pippy/IR.py
@@ -1193,12 +1193,12 @@ class PipeSplitWrapper:
 
 def _split_before_forwad(self, *args, **kwargs):
     pipe_split()
-    return self.orig_forward(*args, **kwargs)
+    return self._orig_forward(*args, **kwargs)
 
 
 def _split_after_forwad(self, *args, **kwargs):
     try:
-        return self.orig_forward(*args, **kwargs)
+        return self._orig_forward(*args, **kwargs)
     finally:
         pipe_split()
 
@@ -1217,7 +1217,7 @@ def annotate_split_points(mod: torch.nn.Module, spec: Dict[str, SplitPoint]):
                 )
 
         mod_to_wrap = getattr(predecessor_module, atoms[-1])
-        mod_to_wrap.orig_forward = mod_to_wrap.forward
+        mod_to_wrap._orig_forward = mod_to_wrap.forward
         if split_type == SplitPoint.BEGINNING:
             mod_to_wrap.forward = MethodType(_split_before_forwad, mod_to_wrap)
         elif split_type == SplitPoint.END:
--- a/pippy/IR.py
+++ b/pippy/IR.py
@@ -1191,12 +1191,12 @@ class PipeSplitWrapper:
     SplitPoint = SplitPoint
 
 
-def _split_before_forwad(self, *args, **kwargs):
+def _split_before_forward(self, *args, **kwargs):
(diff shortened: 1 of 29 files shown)


pippy/IR.py
```diff
@@ -1178,34 +1179,31 @@ def _get_param_buffer_mapping(
...
+class PipeSplitWrapper:
+    # Create a class alias for BC
+    SplitPoint = SplitPoint
 
-            return self.mod(*args, **kwargs)
-        finally:
-            if self.split_point == self.SplitPoint.END:
-                pipe_split()
 
+def _split_before_forwad(self, *args, **kwargs):
```
lessw2020, line 1194: minor but helpful - can we update to proper spelling of 'forward' (currently is missing an r, 'forwad') esp since self._orig_forward right below is correct spelling. 
It can be confusing otherwise.  (might seem nitpicky but had an issue before that all boiled down to a common word mis-spelled in the api vs callers were spelling that word correctly).

pippy/IR.py
```diff
@@ -1178,34 +1179,31 @@ def _get_param_buffer_mapping(
...
+    pipe_split()
+    return self._orig_forward(*args, **kwargs)
 
-def annotate_split_points(
-    mod: torch.nn.Module, spec: Dict[str, PipeSplitWrapper.SplitPoint]
-):
+
+def _split_after_forwad(self, *args, **kwargs):
```
lessw2020, line 1199: same nit, forward spelling.

pippy/IR.py
```diff
@@ -1219,8 +1217,13 @@ def annotate_split_points(
                 )
 
         mod_to_wrap = getattr(predecessor_module, atoms[-1])
-        wrapped_mod = PipeSplitWrapper(mod_to_wrap, split_type)
-        setattr(predecessor_module, atoms[-1], wrapped_mod)
+        mod_to_wrap._orig_forward = mod_to_wrap.forward
+        if split_type == SplitPoint.BEGINNING:
+            mod_to_wrap.forward = MethodType(_split_before_forwad, mod_to_wrap)
```
lessw2020, line 1222: nit, forward spelling

pippy/IR.py
```diff
@@ -1219,8 +1217,13 @@ def annotate_split_points(
...
+        elif split_type == SplitPoint.END:
+            mod_to_wrap.forward = MethodType(_split_after_forwad, mod_to_wrap)
```
lessw2020, line 1224: nit, forward spelling

lessw2020 (approved):
looks great, I like the simplification to SplitPoint directly!
minor nit on forward spelling to avoid confusion.

verified back compat with the google/ddpm-cat-256 example: 
Pipeline stage 0 42M params
Pipeline stage 1 71M params

PR #961, opened 2024-03-07
Rename forward to forward_one_chunk

kwen2501:
* #964
* #963
* #962
* __->__ #961
* #960
* #959

Resolve method name conflicts between the two PipelineStage classes

### DIFF
--- a/pippy/PipelineSchedule.py
+++ b/pippy/PipelineSchedule.py
@@ -19,7 +19,7 @@ logger.setLevel(logging.INFO)
 
 class PipelineStage(ABC, nn.Module):
     @abstractmethod
-    def forward(self, args: List[torch.tensor]) -> torch.tensor:
+    def forward_one_chunk(self, args: List[torch.tensor]) -> torch.tensor:
         """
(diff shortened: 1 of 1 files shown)


pippy/PipelineSchedule.py
```diff
@@ -19,7 +19,7 @@
 
 class PipelineStage(ABC, nn.Module):
     @abstractmethod
-    def forward(self, args: List[torch.tensor]) -> torch.tensor:
+    def forward_one_chunk(self, args: List[torch.tensor]) -> torch.tensor:
```
H-Huang, line 22: i dont have a strong preference, but is `forward_one_microbatch` a clearer name?

H-Huang (approved):

pippy/PipelineSchedule.py
kwen2501: Yeah, `chunk` is more from the old pippy implementation, like `split_args_kwargs_into_chunks`, `merge_chunks`, etc. I don't have a strong pref either, lol

PR #963, opened 2024-03-07
Allow backward_one_chunk to take chunk id

kwen2501:
* #964
* __->__ #963
* #962
* #961
* #960
* #959

Chunk id is used by `backward_one_chunk` method of tracing path's `PipelineStage`


pippy/PipelineStage.py
```diff
@@ -599,19 +599,19 @@ def forward_one_chunk(
 
     def backward_one_chunk(
         self,
-        bwd_chunk: int,
+        chunk: int,
```
H-Huang, line 602: how come chunk is an `int`? Is it more like `chunk_id`?

H-Huang (approved):

PR #964, opened 2024-03-07
Allow step to take no microbatch (e.g. middle stages)

kwen2501:
* #967
* #966
* __->__ #964
* #963
* #962
* #961
* #960
* #959

Middle stages should be able to simply call:
`schedule.step()`
instead of 
`schedule.step(mb)`
because they don't have microbatch inputs.

### DIFF
--- a/pippy/PipelineSchedule.py
+++ b/pippy/PipelineSchedule.py
@@ -432,30 +432,28 @@ class PipelineStageV2Impl(PipelineStageBase):
     def forward_one_chunk(
         self, args: Union[torch.Tensor, List[torch.tensor]]
     ) -> Any:
-        if self.is_first_stage:
-            # we always expect to unpack an iterable of inputs, so if its a single tensor, wrap it in a list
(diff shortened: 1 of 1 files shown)


pippy/PipelineSchedule.py
```diff
@@ -517,8 +519,12 @@ def compute_loss(self):
 
 
 class PipelineSchedule(ABC):
+    def __init__(self, stage: PipelineStageBase, n_microbatches: int):
```
H-Huang, line 522: does `n_microbatches` need to be fixed at initialization? Should we allow `step()` to take microbatches of differing lengths?

kwen2501: We can. 
The caveat is that while rank 0 can figure out `n_microbatches` from `len(microbatches)`, the middle ranks do not take `microbatches` as input (as changed in this PR). We can let `step` to take `n_microbatches`, but then it also seems redundant for rank 0. 
What's your pref?

H-Huang (approved):

PR #967, opened 2024-03-12
Add get_fwd_recv_ops, get_fwd_send_ops

kwen2501:
* #972
* #971
* __->__ #967
* #966
* #964
* #963
* #962
* #961
* #960
* #959

### DIFF
--- a/pippy/PipelineStage.py
+++ b/pippy/PipelineStage.py
@@ -421,6 +421,55 @@ class PipelineStage(torch.nn.Module, QualnameMapMixin):
             # Return a list of empty tuples/dicts with matching length as chunks
             return [()] * self.chunks, [{}] * self.chunks
 
+    def get_fwd_recv_ops(self) -> List[dist.P2POp]:
+        """
+        Returns a list of ops that are needed to receive the input arguments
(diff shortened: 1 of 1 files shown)


pippy/PipelineStage.py
```diff
@@ -421,6 +421,55 @@ def split_inputs(
...
+        ops = []
+        recv_infos = self.args_recv_info[self.fwd_chunk_id]
+        for info in recv_infos:
+            if isinstance(info, RecvInfo):
+                peer_rank = self.stage_index_to_group_rank[info.source]
+                peer_global_rank = (
+                    peer_rank
+                    if self.group is None
+                    else dist.get_global_rank(self.group, peer_rank)
+                )  # TODO
```
H-Huang, line 438: What is the TODO for?

H-Huang (approved):

pippy/PipelineStage.py
kwen2501: It is for c10d I believe. 
```
dist.get_global_rank(self.group, peer_rank)
```
seems to error out when `self.group = None`.

PR #970, opened 2024-03-14
add pipeline schedule unittests to CI

H-Huang:
Needed to add `expecttest` as a dependency since `torch.testing._internal.common_distributed` imports it

removed `strict=True` from zip since that is only supported for python 3.10+ and tests (and pytorch) should support 3.9


kwen2501 (approved):
Thanks!

PR #978, opened 2024-03-18
Integration with DCP

LucasLLC:
## Description

Please read our [CONTRIBUTING.md](https://github.com/pytorch/PiPPy/blob/main/CONTRIBUTING.md) prior to creating your first pull request.

Please include a summary of the feature or issue being fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Testing out some Checkpointing code .

PR description is WIP 

Fixes #(issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

Please describe the Unit or Integration tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Have you added tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

### DIFF
--- a/test/test_transformer.py
+++ b/test/test_transformer.py
@@ -1,6 +1,8 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates
 import torch
 from pippy import annotate_split_points, Pipe, SplitPoint
+import torch.distributed.checkpoint as dcp
+import tempfile
 
 
 d_hid = 16
@@ -65,6 +67,49 @@ def get_layers(module):
     layers = [name for name, _ in module.layers.named_children()]
     return layers
 
+
+def pipe_to_sd(pipe):
+    sd = {}
+    for stage_idx in range(pipe.num_stages):
+        stage_mod = pipe.get_stage_module(stage_idx)
+        sd[f"stage_{stage_idx}"] = stage_mod
+    return sd
+
+with tempfile.TemporaryDirectory() as tmpdir:
+    #Simulate saving the pipe
+    # Option 1:
+    # for stage_idx in range(pipe.num_stages):
+    #     print(f"Saving pipeline stage {stage_idx}")
+    #     stage_mod = pipe.get_stage_module(stage_idx)
+    #     dcp.save(
+    #         {f"stage_{stage_idx}": stage_mod},
+    #         checkpoint_id=f"{tmpdir}_{stage_idx}"
+    #     )
+    # Option 2:
+    sd = pipe_to_sd(pipe)
+    dcp.save(state_dict, checkpoint_id=tmpdir)
+
+
+    #Simulate loading the pipe
+    # Option 1:
+    # for stage_idx in range(pipe.num_stages):
+    #     print(f"Loading pipeline stage {stage_idx}")
+    #     stage_mod = pipe.get_stage_module(stage_idx)
+    #     dcp.load(
+    #         {f"stage_{stage_idx}": stage_mod},
+    #         checkpoint_id=f"{tmpdir}_{stage_idx}"
+    #     )
+
+    #Option 2:
+    new_pipe = Pipe.from_tracing(
+        transformer,
+        1,
+        (x,),
+    )
+    sd = pipe_to_sd(new_pipe)
+    dcp.load(sd, checkpoint_id=tmpdir)
+
+pipe = new_pipe
 
 # Collect all layers in pipe
 layers = []


test/test_transformer.py
```diff
@@ -66,6 +68,49 @@ def get_layers(module):
     return layers
 
 
+def pipe_to_sd(pipe):
```
LucasLLC, line 71: @wz337 , might be interesting in dist state dict

kwen2501:
Thanks for making it work!
Quick comment: 
Do you mind creating a dedicated example for DCP + PP? You can copy the model out (we plan to build a "model hub" for tests, so that would solve the duplicated code problem).

test/test_transformer.py
```diff
@@ -66,6 +68,49 @@ def get_layers(module):
...
+    sd = {}
+    for stage_idx in range(pipe.num_stages):
+        stage_mod = pipe.get_stage_module(stage_idx)
+        sd[f"stage_{stage_idx}"] = stage_mod
+    return sd
+
+with tempfile.TemporaryDirectory() as tmpdir:
+    #Simulate saving the pipe
+    # Option 1:
```
kwen2501, line 80: I think Option 1 would be more likely used than Option 2 in realistic setting. Could you please uncomment this block of code?

test/test_transformer.py
```diff
@@ -66,6 +68,49 @@ def get_layers(module):
...
+    # for stage_idx in range(pipe.num_stages):
+    #     print(f"Saving pipeline stage {stage_idx}")
+    #     stage_mod = pipe.get_stage_module(stage_idx)
+    #     dcp.save(
+    #         {f"stage_{stage_idx}": stage_mod},
```
kwen2501, line 85: Curious, is the dict required by API of DCP? Can a user directly save `stage_mod`?

test/test_transformer.py
```diff
@@ -66,6 +68,49 @@ def get_layers(module):
     return layers
 
 
+def pipe_to_sd(pipe):
+    sd = {}
+    for stage_idx in range(pipe.num_stages):
```
wconstab, line 73: something a little fishy about this proposal (equally so for both option 1 and 2) is that it's not likely you'd want to iterate all the stages in the pipe and load/save them.

Example 1: simple pipeline with 4 gpus
rank0: save/load pipe.submod_0 only
...
Example 2: complex pipeline with 4 gpus, 2 stages per gpu
rank0: save/load pipe.submod_0 and pipe.submod_4
rank1: save/load pipe.submod_1 and pipe.submod_5
...

test/test_transformer.py
```diff
@@ -66,6 +68,49 @@ def get_layers(module):
...
+        stage_mod = pipe.get_stage_module(stage_idx)
+        sd[f"stage_{stage_idx}"] = stage_mod
```
wconstab, line 75: not really clear to me why we need to add a prefix at all.

orig model
Transformer
  embedding
  layers
    0 
    1

Split model

  submod0
     embedding
     layers
        0

  submod 1
    layers
      1

There should be no duplication of fqns between submods/stages.

what are we doing about the 'submod_0' part in the fqn?  when we do `stage_mod = pipe.get_stage_module(stage_idx)` does that return us a module that has top level keys like `embedding` and `layers` or a module that has a top level key of `submod_n`?  

If the former, can't we just save/load the keys as usual?  

If the latter, we can still save/load without a prefix of `stage_{idx}` i think, but we'll sadly be uncompatible to load into a non-PP model later on if we want to.

test/test_transformer.py
wconstab: why does this matter? i think the DCP api had reasons for interfacing with dict instead of model, adding a new variant that takes model and gets its dict should be possible, but i think it's clearer this way that the only part of the model that gets saved is the dict

test/test_transformer.py
kwen2501: Former. @wconstab

test/test_transformer.py
kwen2501: Just to be clear: I like saving the state dict too (instead of the module). That's more composable to me. 
My question above is: is `{f"stage_{stage_idx}": stage_mod}` necessary?

kwen2501:
What's our plan for this PR? @LucasLLC I think we are pretty close to the destination.
Would the following next steps be reasonable?
1. Move the example to `examples/checkpoint`, and name it `pippy_dcp.py`.
2. Focus on Option 1 (per-stage saving), and clean up the UI. (See comments)
3. Make the example runnable in a multi-process setting. Today it saves the stages in a for loop, would be nice if multiple ranks can do their saving job simultaneously.

kwen2501:
For code quality checks, please run:
```
./format.sh
./check.sh
```

PR #983, opened 2024-03-19
Fix T5 example

kwen2501:
Limit this example to 2 ranks only, i.e.

$ torchrun --nproc-per-node 2 pippy_t5.py

Reason:
(1) the need for `decoder_input_ids` -- we would like to provide both `input_ids` and `decoder_input_ids` at stage 0, so stage 0 must be big enough.
(2) the `embed_tokens` module is shared between encoder and decoder. In the 2-rank case, we cut the model carefully so that `embed_tokens` is only used on rank 0.


spupyrev (approved):
Thanks

PR #1014, opened 2024-03-27
allow manual stage to use group

H-Huang:
Update `ManualPipelineStage` to take in group and use it p2p comms

### DIFF
--- a/pippy/ManualPipelineStage.py
+++ b/pippy/ManualPipelineStage.py
@@ -274,10 +274,11 @@ class ManualPipelineStage(PipelineStageBase):
         stage_id: int,
         num_stages: int,
         device: torch.device,
+        group: dist.ProcessGroup = None,
         input_args: Optional[Union[torch.Tensor, List[torch.tensor]]] = None,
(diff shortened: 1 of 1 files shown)


pippy/ManualPipelineStage.py
```diff
@@ -306,9 +307,15 @@ def __init__(
...
-        self.next_stage = (self.rank + 1) % self.world_size
+        def stage_global_rank(peer_rank):
+            return (
+                peer_rank
+                if self.group is None
+                else dist.get_global_rank(self.group, peer_rank)
+            )
+
+        self.prev_stage = stage_global_rank((self.rank - 1) % self.world_size)
+        self.next_stage = stage_global_rank((self.rank + 1) % self.world_size)
```
kwen2501, line 318: nit: how will we deal with the Interleaved case?

kwen2501 (approved):
LGTM.

pippy/ManualPipelineStage.py
H-Huang: in manual pipeline stages there is no support for skip connections (multiple prev and next ranks) so this works

PR #1017, opened 2024-03-29
Add loss and step support to 1F1B

kwen2501:
- Add loss and step to 1F1B
- Extend backward tests to support different schedules
- Iterate over schedules in CI

```
$torchrun --nproc-per-node 4 test/test_bwd.py --schedule 1f1b
```
```
Using PipelineSchedule1F1B
Using PipelineSchedule1F1B
Using PipelineSchedule1F1B
Using PipelineSchedule1F1B
NCCL version 2.20.5+cuda12.4
Rank 3 completes
Rank 2 completes
Rank 1 completes
Rank 0 completes
```

### DIFF
--- a/pippy/PipelineSchedule.py
+++ b/pippy/PipelineSchedule.py
@@ -22,6 +22,7 @@ class PipelineSchedule(ABC):
         loss_fn: Optional[Callable] = None,
     ):
         self._stage = stage
+        self._num_stages = stage.num_stages
         self._n_microbatches = n_microbatches
         self._loss_fn = loss_fn
         self._has_backward = self._loss_fn is not None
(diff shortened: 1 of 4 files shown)


pippy/PipelineSchedule.py
```diff
@@ -282,25 +297,32 @@ def step_microbatches(
...
-            ):
+                if self._should_compute_loss:
+                    target = target_mbs[i]  # type: ignore[index]
+                    loss = self._compute_loss(output, target)
+                    internal_losses.append(loss)
+                    logger.debug(
+                        f"[{self._stage.stage_index}] Loss of microbatch {i}: {loss}"
+                    )
+
+            if i >= warmup_steps and self._has_backward:
```
wconstab, line 314: im kinda curious, is the expectation for all schedules to work for both inference and training?

it makes sense that during training folks would want to run eval; but it might be cleaner code if we just let the schedule be designed for FB or 'inference only'.  Inference-only might be implemented once at the base class schedule.

related question, are there actually different 'schedules' you'd want to use for inference, or are things a lot simpler there?  maybe just BFS vs DFS?

wconstab (approved):
lgtm!

pippy/PipelineSchedule.py
```diff
@@ -314,9 +336,31 @@ def step_microbatches(
         for work in bwd_sends_to_wait:
             work.wait()
 
-    def step(self, *args, **kwargs):
-        # TODO
-        pass
+        # Return losses if there is a container passed in
+        if losses is not None:
+            # Clean external container first
+            losses.clear()
```
wconstab, line 342: wondering if its better practice to request users clear or provide empty container, vs erasing it for them.  i could see this is convenient sometimes.

pippy/PipelineSchedule.py
kwen2501: True. For 1F1B, we should `assert self._has_backward`.
For inference only, today's provision is GPipe (degenerated).

pippy/PipelineSchedule.py
kwen2501: Here is more of a safety thinking. I don't want a case where `len(losses)` ends up not being equal to `n_microbatches`, as I am using `extend()` below.

pippy/PipelineSchedule.py
kwen2501: Actually a second thought, maybe we do wanna 1F1B to be able to do forward only? 
Because even in training, there are eval steps. But user probably don't want to create another `PipelineSchedule` to do inference only. So what would be nice would be:
```
schedule = 1F1B(stage, chunks, loss_fn=...)

# Train mode
schedule.train()
schedule.step(args, target=target, losses=losses)
optim.step()

# Eval mode
schedule.eval()
out = schedule.step(args)
```

wconstab: yea i think .eval() makes sense as a convenience.

is there a difference in how you'd implement it for different schedule classes?  or would it just fallthrough to the base class in all cases

kwen2501: > would it just fallthrough to the base class in all cases?

Depends on whether there is a concrete implementation in the base class, today there isn't.

An alternative would be: each schedule just reuses its own `step_microbatches` function, but with `self._has_backward` turned off -- this will make all the backward steps a no-op, and all the forward steps would still maintain a correct order.

wconstab: i was thinking that adding flags like 'has_backwards to train schedules was a higher cost in code/complexity than providing one simple inference function in the base.

is there a good reason to have more than one inference function?  (would you want to do inference differently depending which train schedule you picked?)  IMO the intricacy of the train schedules are entirely about overlapping between F and B, so its not clear to me that if you take out the B there is any value in using the same F logic.  But i haven't looked closely at the schedules with this in mind.

PR #1021, opened 2024-04-02
PP with modified DDP example

H-Huang:
This is the example that ads uses and converted it to a test

`python -m unittest test_pipeline_with_ddp.TestPipelineDDP.test_manual_pipeline_with_manual_allreduce`

Uses 4 ranks, 2 pipeline stages and 2 ddp groups:

        # 2 pipeline stages, 2 ddp groups
        #         PP0     PP1
        # DP0     0       2
        #         v       v
        # DP1     1       3

DDP is a modified version where allreduce is after the pipeline step

Only uses manual pipeline stages for now. will update

### DIFF
--- a/test/test_pipeline_with_ddp.py
+++ b/test/test_pipeline_with_ddp.py
@@ -1,0 +1,179 @@
+# (c) Meta Platforms, Inc. and affiliates. Confidential and proprietary.
+import unittest
+
+import torch
+import torch.distributed as dist
+import torch.nn as nn
+from pippy.ManualPipelineStage import ManualPipelineStage
+from pippy.PipelineSchedule import PipelineScheduleGPipe
+from torch.distributed.device_mesh import init_device_mesh
+
+# torch.testing._internal.common_distributed requies "expecttest"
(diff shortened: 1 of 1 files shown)


test/test_pipeline_with_ddp.py
```diff
@@ -0,0 +1,177 @@
...
+from pippy.PipelineSchedule import PipelineScheduleGPipe
+from torch.distributed.device_mesh import init_device_mesh
+from torch.nn.modules.loss import F
+
+# torch.testing._internal.common_distributed requies "expecttest"
+from torch.testing._internal.common_distributed import MultiProcessTestCase
+from torch.testing._internal.common_utils import FILE_SCHEMA
+
+
+class DDPAROnce(torch.nn.Module):
```
kwen2501, line 23: How representative is this helper?
Is there a preference between using this helper vs using `with ddp_model.no_sync()`?

kwen2501 (commented):
I think the addition of `group` to `ManualPipelineStage` is nice.
But I am not sure if we'd support testing with `DDPAROnce` for long term.

PR #1023, opened 2024-04-03
Remove qualname remap support

kwen2501:
Because unflattening already made the FQNs the same as the original FQNs.

Also modified test to directly assert FQN equivalence.

* #1030
* __->__ #1023

### DIFF
--- a/test/test_fwd.py
+++ b/test/test_fwd.py
@@ -86,13 +86,11 @@ def run_worker(args):
         )
 
     # Test qualname mapping
-    sd = stage.submod.state_dict()
-    print(f"Rank {args.rank} state dict keys: {sd.keys()}")
-    remapped_keys = [stage.remap_qualname(k) for k in sd.keys()]
-    print(f"Rank {args.rank} remapped keys: {remapped_keys}")
-    # Confirm remapped keys are consistent with original model
+    submod_keys = stage.submod.state_dict().keys()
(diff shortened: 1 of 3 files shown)


test/test_fwd.py
```diff
@@ -86,13 +86,11 @@ def run_worker(args):
...
-    print(f"Rank {args.rank} state dict keys: {sd.keys()}")
-    remapped_keys = [stage.remap_qualname(k) for k in sd.keys()]
-    print(f"Rank {args.rank} remapped keys: {remapped_keys}")
-    # Confirm remapped keys are consistent with original model
+    submod_keys = stage.submod.state_dict().keys()
+    print(f"Rank {args.rank} state dict keys: {submod_keys}")
+    # Confirm keys are consistent with original model
     old_keys = mod.state_dict().keys()
-    assert all(rk in old_keys for rk in remapped_keys)
+    assert all(k in old_keys for k in submod_keys)
```
H-Huang, line 93: curious, are `old_keys` supposed to be the same as `submod_keys`? How come we aren't doing the equivalence check that is being done in `test_pipe.py`

H-Huang (approved):

test/test_fwd.py
kwen2501: `old_keys` is a bigger set than `submod_keys` because it comes from the full model.

PR #1024, opened 2024-04-03
Unify pipeline stages fwd and bwd, add pipeline stage tests

H-Huang:
Fixes https://github.com/pytorch/PiPPy/issues/1005

Changes:

- Unify the `forward_one_chunk` and `backward_one_chunk` to be used by both `PipelineStage` and `ManualPipelineStage`
- Add `test_pipeline_stage` to test both `PipelineStage` and `ManualPipelineStage` can be used interchangeably and without error.

TODO:
- create test verify that the outputs for both manual pipeline stage and pipeline stage are equivalent

### DIFF
--- a/pippy/ManualPipelineStage.py
+++ b/pippy/ManualPipelineStage.py
@@ -8,6 +8,12 @@ import torch
 import torch.distributed as dist
 import torch.nn as nn
 
+from pippy.microbatch import (
+    merge_chunks,
+    split_args_kwargs_into_chunks,
+    TensorChunkSpec,
+)
+
 from pippy.PipelineStage import PipelineStageBase
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


pippy/ManualPipelineStage.py
```diff
@@ -359,20 +418,22 @@ def init_p2p_neighbors(self):
     def get_fwd_recv_ops(self) -> List[dist.P2POp]:
         if self.is_first_stage:
             return []
+        self.inputs_grad = [torch.empty_like(inp) for inp in self.inputs]
```
kwen2501, line 323: Is there a way to avoid creating new tensors at each iteration?

pippy/ManualPipelineStage.py
```diff
@@ -359,20 +418,22 @@ def init_p2p_neighbors(self):
...
         ]
 
     def get_fwd_send_ops(self) -> List[dist.P2POp]:
-        assert (
-            len(self.outputs) != 0
-        ), "forward() must be called before get_fwd_send_ops"
+        output_tuples, flatten_input_tensors = self.fwd_cache[
+            self.fwd_chunk_id - 1
+        ]
+        self.outputs_grad = [torch.empty_like(out) for out in output_tuples]
```
kwen2501, line 333: Same here.

kwen2501 (approved):
LGTM! Thanks for the consolidation!
A couple minor comments.

pippy/ManualPipelineStage.py
```diff
@@ -324,6 +330,59 @@ def stage_global_rank(peer_rank):
...
+        """
+        Splits a full-batch input into chunks (i.e. microbatches) and returns
+        the chunks
+        """
+        if args or kwargs:
+            # TODO: cannot split on another dimension other than 0
+            args_split, kwargs_split = split_args_kwargs_into_chunks(
+                args,
+                kwargs,
+                self.num_stages,
```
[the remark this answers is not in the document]
H-Huang: good catch, thanks!

PR #1025, opened 2024-04-03
Remove incorrect assertion on loss shapes

wconstab:
* #1028
* __->__ #1025

output and target shapes are expected to differ.

It might be possible to craft some shape-checking logic here, but it
would need to account for various loss functions and be general, not
sure its feasible. Removing for now to unblock.

(using cross-entropy loss in this case- see docs about expected shapes
https://pytorch.org/docs/stable/generated/torch.nn.functional.cross_entropy.html)


kwen2501 (approved):
LGTM! Sorry for my ignorance.

PR #1028, opened 2024-04-03
Unwrap output_vals in _create_grad_recv_info

wconstab:
* #1029
* __->__ #1028

Verified test case catches the error if the fix is commented.

Note: I'm not sure why the 'output' node is only _sometimes_ packaged in
a tuple.  I found that for example `torchrun --nproc-per-node 3
examples/basic/example_train.py` would fail if I unconditionally
unwrapped the output tuple.  The current solution is tenuous at best,
but seems to pass tests.

### DIFF
--- a/pippy/PipelineStage.py
+++ b/pippy/PipelineStage.py
@@ -415,6 +415,9 @@ class PipelineStage(PipelineStageBase, QualnameMapMixin):
         # The output node may take multiple args, meaning the submod having multiple output values.
         output_vals = output_node.args
 
+        # output vals is packaged in a wrapper tuple that must be shed before iterating its inner tuple's contents
+        output_vals = output_vals[0]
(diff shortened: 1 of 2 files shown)


kwen2501 (approved):
Thanks for catching the bug. LGTM.

kwen2501:
If it helps with code style, maybe we can do something like this:
https://github.com/pytorch/PiPPy/blob/ccb7c42cca4f3c4bdb3718c5e1613aca41bbdff1/pippy/PipelineStage.py#L729

kwen2501:
`flatten_args` is something like this:
https://github.com/pytorch/PiPPy/blob/ccb7c42cca4f3c4bdb3718c5e1613aca41bbdff1/pippy/utils.py#L53-L56

PR #1030, opened 2024-04-03
Add stage constructor for stage module

### LINKED ISSUE #1012: create constructor that accepts stage module (KW)

kwen2501:
User code example:
```
stage_module = pipe.get_stage_module(stage_index)
pipe_info = pipe.info()
stage = _PipelineStage(stage_module, stage_index, pipe_info, device, group)
```

- Added `_PipelineStage` constructor:
Temporarily underscoring it as I need to find ways to migrate the existing `PipelineStage(pipe, stage_index, ...)`

- Added class `PipeInfo` to convey pipe metadata

Resolves #1012

* __->__ #1030

### DIFF
--- a/pippy/IR.py
+++ b/pippy/IR.py
@@ -2,6 +2,7 @@
 import copy
 import logging
 import operator
+from dataclasses import dataclass
 from enum import Enum
 from inspect import Parameter, signature, Signature
 from types import MethodType
@@ -531,7 +532,7 @@ class Pipe(QualnameMapMixin, torch.nn.Module):
         self.split_gm: fx.GraphModule = split_gm
         self.executor: DetachExecutor = DetachExecutor(self.split_gm)
         self.num_stages: int = num_stages
-        self.has_loss_and_backwards = has_loss_and_backward
+        self.has_loss_and_backward = has_loss_and_backward
         self.loss_spec = loss_spec
 
         for node in split_gm.graph.nodes:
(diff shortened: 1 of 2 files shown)


pippy/IR.py
```diff
@@ -1207,6 +1208,27 @@ def __str__(self):
...
+    class PipeInfo:
+        graph: fx.Graph
+        num_stages: int
+        num_chunks: int
+        has_loss_and_backward: bool
+        args_chunk_spec: Optional[Tuple[Any, ...]] = None
+        kwargs_chunk_spec: Optional[Dict[str, Any]] = None
+        output_chunk_spec: Any = None  # TODO (kwen2501): type it better
+
+    def info(self) -> PipeInfo:
```
wconstab, line 1221: why do you need this method, that seems to take an existing PipeInfo and make it into a new one?

kwen2501: ? 
`self` here is a `Pipe`, not `PipeInfo`

pippy/PipelineStage.py
```diff
@@ -199,29 +206,27 @@ def __init__(
         # Caching chunk outputs for final output merge or reduction
         self.output_chunks: List[Any] = []
 
-        # Find my submodule
-        self.split_gm = self.pipe.split_gm
-        named_children = list(self.split_gm.named_children())
-        self.name, self.submod = named_children[stage_index]
+        # Find stage nodes in graph
+        submod_nodes = [
+            node for node in pipe_info.graph.nodes if node.op == "call_module"
```
wconstab, line 211: i think this PR is good on its own as a refactoring step that introduces a good API boundary, which is good.

but i'm curious, for stuff like this, where you still reach into pipe_info.graph, do you plan to eventually reduce this over time and replace with 'well defined fields' on pipe_info?

wconstab (approved):
lgtm.

nit: pipe_info sounds more like user-facing info.  maybe 'pipe_meta' is more accurate but not a big deal to me.

pippy/PipelineStage.py
kwen2501: what would be a better way to describe stage relationship than a graph?

kwen2501: note that this is an unflattened graph, so all it has is just:
submod0 --> submod1 --> submod2 ...

kwen2501: I could stick `pipe_info` into `stage_module`. But my concern is if any of DDP, FSDP, TP wraps `stage_module`, this attribute can't be found by `_PipelineStage`. Any good solution along this direction?

wconstab: i guess turning it from a graph (which is general) into specific fields like "my receivers" and "my senders" - which is specific and verbose but also more well constrained

i'm not too familiar with all the things you use the graph for though, so i dont know how much this will work

pippy/IR.py
wconstab: oh! missed that

PR #1032, opened 2024-04-04
Bring back stage_backward helper

kwen2501:
https://github.com/pytorch/PiPPy/pull/1029 reminds us of the need to extract outputs that requires grad. 

So I am bringing pippy's previous `stage_backward` helper back in action.

This helper has extraction functions for general cases.

Also brings back `backward_maybe_with_no_sync` to support DDP.

Cc: @wconstab

### DIFF
--- a/pippy/PipelineStage.py
+++ b/pippy/PipelineStage.py
@@ -450,6 +450,9 @@ class _PipelineStage(PipelineStageBase):
         grad_send_info: List[Optional[int]] = []
 
         def map_recv_to_send(a):
+            # Note: we send gradients back to previous stage as long as in
+            # forward it is a received input, regardless of whether it requires
+            # grad. It is up to the previous stage to disgard this gradient.
             if isinstance(a, RecvInfo):
                 grad_send_info.append(a.source)
                 return a.source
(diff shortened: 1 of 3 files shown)


pippy/PipelineStage.py
```diff
@@ -663,9 +666,13 @@ def forward_maybe_with_nosync(self, *args, **kwargs):
             out_val = self.submod(*args, **kwargs)
         return out_val
 
-    def backward_maybe_with_nosync(self, bwd_kwargs: Dict, is_last_chunk: bool):
+    def backward_maybe_with_nosync(self, bwd_kwargs: Dict, bwd_chunk_id: int):
+        # Add `stage_info` for debug purpose in `stage_backward`
+        bwd_kwargs["stage_info"] = f"{self.stage_index}"
```
H-Huang, line 671: how is the debug information used?

H-Huang (approved):
Nice! I was curious and going to ask why`backward_maybe_with_nosync` wasn't used but thanks for fixing

pippy/PipelineStage.py
kwen2501: Just like this, pretty dumb:
```
    except Exception as e:
        exc_msg = f"""
        Failed to run backward on stage {stage_info}
        Stage output: {map_debug_info(stage_output)}
        Output gradient: {map_debug_info(output_grads)}
        Input: {map_debug_info(input_values)}
        """
        raise RuntimeError(exc_msg) from e
```
I could remove this argument later.

kwen2501: Removed :)

PR #1035, opened 2024-04-05
Use context manager to replace spec arguments of pipeline API

kwen2501:
Removing these arguments from the `pipeline` API (which is used to create a `Pipe`):
```
def pipeline(
     module: torch.nn.Module,
     num_chunks: int,
     example_args: Tuple[Any, ...],
     example_kwargs: Optional[Dict[str, Any]] = None,
     split_policy: Optional[Callable[[fx.GraphModule], fx.GraphModule]] = None,
-    args_chunk_spec: Optional[Tuple[Any, ...]] = None,
-    kwargs_chunk_spec: Optional[Dict[str, Any]] = None,
-    output_chunk_spec=None,
-    constraints: Optional[List[Constraint]] = None,
) -> Pipe:
```

`args_chunk_spec` and `kwargs_chunk_spec` are now supported via a context manager:
```
# Numbers here are tensor chunking dimensions
with ArgsChunkSpec((0, 0, 1)), KwargsChunkSpec({"id": 0}):
    pipe = pipeline(model, num_chunks, example_args, example_kwargs)
```

For most cases, users don't need to provide the chunking dimensions, the default dimension of 0 (batch dimension) will be used. So the API would be as simple as:
```
pipe = pipeline(model, num_chunks, example_args, example_kwargs)
```

`output_chunk_spec` is gone because I plan to move it to PipelineSchedule -- user tells us there how they want the output to be presented. Also because IR does not need output_chunk_spec.

`constraint` is gone because it is gone from torch.export's API. (Replaced by something else). And we'd never truly have support for constraint in pippy. We can add support for the something else in future.

* __->__ #1035


pippy/IR.py
```diff
@@ -1326,54 +1338,80 @@ def annotate_split_points(mod: torch.nn.Module, spec: Dict[str, SplitPoint]):
...
+        Pipe.args_chunk_spec = None
+
+
+# Context manager for setting `kwargs_chunk_spec` during creation of Pipe
+class KwargsChunkSpec:
+    def __init__(
+        self,
+        chunk_dims: Dict[str, int],
+    ):
+        self.kwargs_chunk_spec = map_aggregate(
```
H-Huang, line 1405: Wouldn't this apply to all kwargs in the dict? In the summary description this is used like `KwargsChunkSpec({"id": 0})` so wouldn't we want to just apply to "id" key?

test/test_pipe_bwd.py
```diff
@@ -84,16 +82,10 @@ def run_worker(args, model_class):
     x = torch.randn(batch_size, d_hid)
     y = torch.randn(batch_size, d_hid)
 
-    output_chunk_spec = (
-        TensorChunkSpec(0),  # logits
-        sum_reducer,  # loss
-    )
-
     pipe = pipeline(
```
H-Huang, line 85: perhaps add an example/test using the context manager?

H-Huang (approved):
Nice! `pipeline` looks a lot cleaner now

wconstab:
why is output and constraints totally gone?

pippy/IR.py
kwen2501: So a more general example would be like:
`KwargsChunkSpec({"id": 0, "mask": 1})`
then this helper will cut the id tensor on dim 0, and cut the mask tensor on dim 1.

So yes, the intention is to apply to all kwargs, based on their respective chunk dim.

In future, we may have to support "replicate" too -- instead of chunking.

kwen2501:
@wconstab sorry failed to mention that.

`output_chunk_spec` is gone because I plan to move it to PipelineSchedule -- user tells us there how they want the output to be presented. Also because IR does not need `output_chunk_spec`.

`constraint` is gone because it is gone from torch.export's API. (Replaced by something else). And we'd never truly have support for `constraint` in pippy. We can add support for the something else in future.

kwen2501:
@H-Huang test added
```
    def forward(self, x, y, z=torch.zeros(batch_size, d_hid)):
        ...

    x = torch.randn(batch_size, d_hid)
    y = torch.randn(batch_size, d_hid)
    z = torch.randn(batch_size, d_hid)

    with ArgsChunkSpec((0, 0)), KwargsChunkSpec({"z": 0}):
        pipe = pipeline(
            mod,
            chunks,
            example_args=(x, y),
            example_kwargs={"z": z},
        )
```

PR #1045, opened 2024-04-10
Add FSDP support for GPipe schedule

wconstab:
* __->__ #1045

Adds a helper that can be integrated to other schedules too.

Set up an integration test and attach it to CI runner

### DIFF
--- a/.github/workflows/gpu_tests.yaml
+++ b/.github/workflows/gpu_tests.yaml
@@ -43,7 +43,7 @@ jobs:
         run: conda activate test
       - name: Install dependencies
         run: |
-          pip install numpy
+          pip install numpy expecttest
           pip install --pre -r requirements.txt --find-links https://download.pytorch.org/whl/nightly/cu121/torch_nightly.html
       - name: Install pippy
         run: python setup.py install
--- a/.github/workflows/gpu_tests.yaml
+++ b/.github/workflows/gpu_tests.yaml
@@ -53,6 +53,8 @@ jobs:
         run: torchrun --nproc-per-node 4 test/test_autosplit.py
       - name: Test skip connection support
         run: torchrun --nproc-per-node 4 test/test_skip_conn.py
+      - name: Test Composability
+        run: python test/test_composability.py
       - name: Run example
         run: torchrun --nproc-per-node 3 examples/basic/example.py
       - name: Run training example
--- a/pippy/PipelineSchedule.py
+++ b/pippy/PipelineSchedule.py
@@ -8,6 +8,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 import torch
 import torch.distributed as dist
 from torch.profiler import record_function
+from torch.distributed._composable.fsdp.fully_shard import FSDP
 
 from pippy.microbatch import merge_chunks, split_args_kwargs_into_chunks
 
@@ -41,6 +42,21 @@ class PipelineSchedule(ABC):
             self._stage.pipe_info if hasattr(self._stage, "pipe_info") else None  # type: ignore[attr-defined]
         )
         self._output_merge_spec = output_merge_spec
+
+    def _configure_data_parallel_mode(self, last_backward: bool):
+        """
+        Whether using PP with FSDP or DDP, there are some runtime differences between the last backward step and the
+        other steps.  Namely, we need to accumulate gradients on previous steps and reduce them on the last step, but
+        there are additional state-variables and performance considerations depending on the data parallelism used.
+
+        This helper should adapt any pipeline parallel schedule to work with common/supported data parallel libraries.
+        """
+        module = self._stage.submod
+        if isinstance(module, FSDP):
+            module.set_is_last_backward(last_backward)
+            module.set_requires_gradient_sync(last_backward)
+        else:
+            raise NotImplementedError("Only FSDP is supported now, please add support for your data parallel library")
 
     @abstractmethod
     def step_microbatches(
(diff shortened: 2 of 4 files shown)


test/test_composability.py
```diff
@@ -174,6 +184,79 @@ def test_manual_pipeline_with_manual_allreduce(self):
...
+            pp_group,
+            input1,
+            num_microbatches,
+        )
+
+        pipeline_schedule = ScheduleGPipe(
+            pipeline_stage,
+            n_microbatches=num_microbatches,
+            # dummy loss needed just to force backwards to run in schedule step
+            loss_fn=lambda y, t: y.sum(dim=1),
```
wconstab, line 253: im doing something wrong here? rank0,2 is complaining about shapes.  But loss should be running on rank1,3. so i'm confused.

RuntimeError:       
        Failed to run stage backward:
        Stage output: ('Tensor(torch.Size([10]), grad=True)',)
        Output gradient: ['Tensor(torch.Size([1, 10]), grad=False)']
        Input: ['Tensor(torch.Size([1, 10]), grad=False)']

pippy/PipelineSchedule.py
```diff
@@ -42,6 +43,21 @@ def __init__(
...
+
+        This helper should adapt any pipeline parallel schedule to work with common/supported data parallel libraries.
+        """
+        module = self._stage.submod
+        if isinstance(module, FSDP):
+            module.set_is_last_backward(last_backward)
+            module.set_requires_gradient_sync(last_backward)
+        else:
+            raise NotImplementedError("Only FSDP is supported now, please add support for your data parallel library")
+
```
kwen2501, line 60: This looks simple and clean.
nit: do you mind moving this method to `PipelineStage` level? 
Two reasons:
1. we need to access `submod` here;
2. I am refactoring self._stage -- in case of multiple stages per rank, `self._stage` will be gone (and you probably don't want two implementations here).

kwen2501: In `PipelineStage`, you can get `last_backward: bool` by checking `bwd_chunk_id`, like here:
https://github.com/pytorch/PiPPy/blob/9cd29d17324a53b1342c2b6fc5eeef5f5405ec49/pippy/PipelineStage.py#L149-L152

pippy/PipelineSchedule.py
```diff
@@ -236,6 +252,10 @@ def step_microbatches(
         # Delay send waits
         bwd_sends_to_wait: List[dist.Work] = []
         for i in range(self._n_microbatches):
+
+            # set library-specific data-parallel config flags to ensure gradient accumulation across microbatches
+            self._configure_data_parallel_mode(i == self._n_microbatches - 1)
+
```
kwen2501, line 258: Following the above comment -- if we move `_configure_data_parallel_mode` to `PipelineStage` level, then `PipelineSchedule` can remain agnostic to submod type.

test/test_composability.py
```diff
@@ -174,6 +184,79 @@ def test_manual_pipeline_with_manual_allreduce(self):
...
+        for layer_name, layer in partial_model.named_children():
+
+            # As an optimization, do not reshard after forward for the last
+            # transformer block since FSDP would prefetch it immediately
+            # reshard_after_forward = layer_id < len(model.layers) - 1
+            # TODO(whc) need to fix correctly handle layer-ids on pp-split module
+            fully_shard(
+                layer,
+                **fsdp_config,
+                reshard_after_forward=True,
```
kwen2501, line 231: nit: should we keep `reshard_after_forward` false all the time in view of multiple microbatches?

test/test_composability.py
```diff
@@ -174,6 +184,79 @@ def test_manual_pipeline_with_manual_allreduce(self):
...
+            )
+            setattr(partial_model, layer_name, layer)
```
kwen2501, line 233: nit: this line is necessary for AC only according to @awgu

test/test_composability.py
kwen2501: @H-Huang can you help please?

kwen2501 (approved):

test/test_composability.py
kwen2501: @awgu ? wdyt?

test/test_composability.py
awgu: Yep! We do not need this call if we are not using `CheckpointWrapper` (or some other `nn.Module` wrapper). If we are only using `fully_shard`, then there is no need.

pippy/PipelineSchedule.py
wconstab: yep, that sounds good!

test/test_composability.py
wconstab: https://github.com/pytorch/PiPPy/issues/1054
made it an issue.  and found a workaround but need to discuss what the right fix is.

pippy/PipelineSchedule.py
wconstab: @H-Huang can you TAL? 

I just realized that there is already DDP support in there in PipelineStageBase in the form of the _nosync methods.

Do you think it would be easy to put the ddp nosync into the 'configure_data_parallel_mode` helper and then remove the 'maybe_with_nosync' helpers, or should we put FSDP into those helpers instead?

H-Huang: I think we should put the ddp nosync logic into the `'configure_data_parallel_mode`  method you created. I think it is clearer and more explicit that the forward of the module is being modified. I'm happy to do this in a follow up issue or PR as well!

.github/workflows/gpu_tests.yaml
```diff
@@ -53,6 +53,8 @@ jobs:
         run: torchrun --nproc-per-node 4 test/test_autosplit.py
       - name: Test skip connection support
         run: torchrun --nproc-per-node 4 test/test_skip_conn.py
+      - name: Test Composability
+        run: python test/test_composability.py
```
H-Huang, line 57: looks like the job also needs to `pip install expecttest` since the pt distributed tests use it

H-Huang (approved):
Tests and changes look good! Agree with Ke that any module manipulation should happen at the `PipelineStage` level

pippy/PipelineSchedule.py
wconstab: ok. in this PR i just moved the `_configure_data_parallel_mode` helper into stage for now.

we may need to do several changes to migrate DDP, so lets keep that in a separate PR and also add DDP to the unit test for that one.  I'll let you do that if you are offering.

PR #1048, opened 2024-04-11
Add base class for multi-stage schedules

kwen2501:
Added class `PipelineScheduleMulti` as Base class for multi-stage schedules. And implement a common `step()` function for multi-stage schedules.

* #1055
* #1052
* #1051
* __->__ #1048

### DIFF
--- a/pippy/ManualPipelineStage.py
+++ b/pippy/ManualPipelineStage.py
@@ -301,7 +301,6 @@ class ManualPipelineStage(PipelineStageBase):
             dist.P2POp(dist.isend, out, self.next_stage, self.group)
             for out in output_tuples
         ]
-        raise AssertionError("invalid fwd_outputs type, cannot send")
 
     def check_and_format_outputs(self, outputs: Any) -> List[torch.tensor]:
         # validate the output of the module is a type supported
--- a/pippy/PipelineSchedule.py
+++ b/pippy/PipelineSchedule.py
@@ -218,7 +218,9 @@ class PipelineScheduleSingle(PipelineSchedule):
 
         # Split target into microbatches
         if target is not None:
-            targets_split = torch.tensor_split(target, self._n_microbatches)
+            targets_split = list(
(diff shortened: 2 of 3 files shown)


pippy/ManualPipelineStage.py
```diff
@@ -301,7 +301,6 @@ def get_fwd_send_ops(self) -> List[dist.P2POp]:
             dist.P2POp(dist.isend, out, self.next_stage, self.group)
             for out in output_tuples
         ]
-        raise AssertionError("invalid fwd_outputs type, cannot send")
```
kwen2501, line 304: @H-Huang We will never hit here, right?

pippy/PipelineSchedule.py
```diff
@@ -464,27 +529,32 @@ def step_microbatches(
...
+        output_merge_spec: Optional[Union[Dict[str, Any], Tuple[Any]]] = None,
+    ):
+        self.pp_group_size = stages[0].group_size
+        # TODO: is this limitation a must?
+        if n_microbatches % self.pp_group_size != 0:
             raise ValueError(
-                "Looped DFS schedule requires at least two stages to be used."
+                f"Interleaved 1F1B schedule requires the number of microbatches ({self._n_microbatches}) \
+                to be a multiple of the number of pipeline ranks ({self.pp_group_size})."
             )
```
kwen2501, line 547: @H-Huang wondering if this is a real limitation. If it is, can you put some words here to describe the reason?

pippy/ManualPipelineStage.py
H-Huang: yeah this is not necessary, thanks for removing

pippy/PipelineSchedule.py
H-Huang: i dont think this is a hard requirement, but i haven't seen the literature describe otherwise. I have to do some tests and think through to see if different sized microbatches also works

H-Huang (approved):

PR #1049, opened 2024-04-11
Use stage index in PipelineStage's log prefix

kwen2501:
Instead of using rank as the log prefix, clearer in case of Interleaved 1F1B.

### DIFF
--- a/pippy/PipelineStage.py
+++ b/pippy/PipelineStage.py
@@ -191,7 +191,7 @@ class PipelineStageBase(ABC):
 
         except Exception as e:
             exc_msg = f"""s
-            Rank {self.group_rank} failed to run forward stage
+            Stage {self.stage_index} failed to run forward:
             args: {map_debug_info(composite_args)}
             kwargs: {map_debug_info(composite_kwargs)}
             """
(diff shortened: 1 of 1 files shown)


pippy/PipelineStage.py
```diff
@@ -218,7 +218,7 @@ def forward_one_chunk(
         )
 
         logger.debug(
-            f"[{self.group_rank}] Forwarded chunk {self.fwd_chunk_id}, outputs: {map_debug_info(output)}"
+            f"[{self.stage_index}] Forwarded chunk {self.fwd_chunk_id}, outputs: {map_debug_info(output)}"
```
wconstab, line 221: maybe its better to specify `Stage {stage_idx}` everywhere we log.  I could see it being still a bit confusing if you get logs saying 

"3 forwarded chunk 2 ... "

wconstab (approved):
love the change. suggesting one more change though

H-Huang (approved):
looks good!

pippy/PipelineStage.py
kwen2501: Thanks! added.

PR #1052, opened 2024-04-12
Fix backward microbatch index in 1F1B

kwen2501:
* #1055
* __->__ #1052


H-Huang (approved):
yep this indeed looks like a mistake. Thanks for fixing!

PR #1058, opened 2024-04-15
Pipeline stages share fwd/bwd send/recv ops

H-Huang:
After changes in https://github.com/pytorch/PiPPy/pull/1024, `PipelineStage` and `ManualPipelineStage` shared the same data structures for managing the buffers and fwd cache. These changes build on top of that so that also the `get_fwd_recv_ops`, `get_fwd_send_ops`, `get_bwd_recv_ops`, `get_bwd_send_ops` are also shared.

* #1063
* #1060
* #1059
* __->__ #1058

### DIFF
--- a/pippy/ManualPipelineStage.py
+++ b/pippy/ManualPipelineStage.py
@@ -8,7 +8,12 @@ import torch
 import torch.distributed as dist
 import torch.nn as nn
 
-from pippy.PipelineStage import PipelineStageBase
+from pippy.PipelineStage import (
+    _make_tensor_from_meta,
+    PipelineStageBase,
+    RecvInfo,
+    RootArgPlaceholder,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -181,9 +186,7 @@ class ManualPipelineStage(PipelineStageBase):
         super().__init__(
             module, stage_id, num_stages, device, num_microbatches, group
         )
-        self.module = module.to(device)
-        self.is_first_stage = stage_id == 0
-        self.is_last_stage = stage_id == num_stages - 1
+        self.submod.to(self.device)
         # When we materialize the model partition on cuda, we call reset_parameters() if it is available
         # logger.info(f"input args {input_args=}")
         self.inputs: List[torch.tensor] = []
@@ -193,7 +196,7 @@ class ManualPipelineStage(PipelineStageBase):
             self.inputs = create_buffers(input_args, device)
 
         if output_args is None:
-            self.outputs = self.module(*self.inputs)
+            self.outputs = self.submod(*self.inputs)
             # create buffers for the output so that the data is in the correct
             # shape in order to use in p2p op (send)
             self.outputs = create_buffers(self.outputs, device)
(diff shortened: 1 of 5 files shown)


pippy/ManualPipelineStage.py
```diff
@@ -220,36 +223,63 @@ def stage_global_rank(peer_rank):
...
+        # Receive info during forward
+        # TODO: create args_recv_info lazily? (same needed for PipelineStage)
+        for chunk_id in range(self.chunks):
+            self.set_requires_grad[chunk_id] = False
+            if not self.is_first:
+                self.args_recv_info[chunk_id] = tuple(
+                    [
+                        RecvInfo(
+                            f"recv_for_{self.stage_index}_from_{self.stage_index - 1}",
+                            self.stage_index - 1,
```
kwen2501, line 235: nit: Can you added a comment for this line stating the assumption of receiving from -1 stage?

pippy/ManualPipelineStage.py
```diff
@@ -220,36 +223,63 @@ def stage_global_rank(peer_rank):
...
+                self.args_recv_info[chunk_id] = tuple(
+                    [RootArgPlaceholder() for _ in self.inputs]
+                )  # type: ignore
+
+        # Send info during forward for each activation
+        # only need the rank that is being sent to
+        self.act_send_info: Dict[int, List] = {}
+        for idx in range(len(self.outputs)):
+            if not self.is_last:
+                self.act_send_info[idx] = [self.stage_index + 1]
```
kwen2501, line 251: Same here. Nice to state an assumption.

pippy/ManualPipelineStage.py
```diff
@@ -220,36 +223,63 @@ def stage_global_rank(peer_rank):
...
-        return self.outputs_grad
+        act_send_info: Dict,
+    ) -> Tuple[RecvInfo, ...]:
+        grad_recv_info: Tuple[RecvInfo, ...] = ()
+        if not self.is_last:
+            grad_recv_info = tuple(
+                [
+                    RecvInfo(
+                        f"recv_grad_for_{self.stage_index}_from_{dst_list[0]}",
+                        dst_list[0],
```
kwen2501, line 274: Same here. Nice to state an assumption

pippy/ManualPipelineStage.py
```diff
@@ -220,36 +223,63 @@ def stage_global_rank(peer_rank):
...
+                        _make_tensor_from_meta(self.outputs[idx], self.device),
+                    )
+                    for idx, dst_list in act_send_info.items()
+                ]
+            )
+        else:
+            grad_recv_info = tuple()
```
kwen2501, line 281: nit: this else branch is not needed given the initialization at the top.

kwen2501 (approved):
Thanks for the consolidation! Looks good to me!

pippy/ManualPipelineStage.py
H-Huang: yep, good point. thanks!

PR #1059, opened 2024-04-15
add loss to looped_bfs

H-Huang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #1063
* #1060
* __->__ #1059
* #1058

PR #1060, opened 2024-04-16
Refactor loss logic in schedules

H-Huang:
Refactor the logic used to calculate and update loss into the base `PipelineSchedule` class methods.

Will update Interleaved1F1B in the following PR

* #1063
* __->__ #1060
* #1059
* #1058


kwen2501 (approved):
Looks good to me!

PR #1066, opened 2024-04-17
Add gradient equivalence test

kwen2501:
Each stage will check its gradients against a reference model. 
The ref model uses a full-batch to run backward.

### DIFF
--- a/test/test_grad.py
+++ b/test/test_grad.py
@@ -1,0 +1,190 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates
+import argparse
+import copy
+import os
+import unittest
+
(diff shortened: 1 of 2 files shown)


test/test_grad.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+
+    stage = PipelineStage(
+        pipe,
+        args.rank,
+        device=args.device,
+    )
+
+    # Attach to a schedule
+    ScheduleClass = schedule_map[args.schedule]
```
H-Huang, line 92: cool trick

H-Huang (approved):
Nice!

PR #1070, opened 2024-04-18
Zero grad of input buffers

kwen2501:
## Description

Zero grad of input buffers between schedule steps. 
This is because `torch.autograd.backward()` will accumulate gradients into leaf tensors by default. For gradients to pass back to previous stages, we don't want such accumulation.

## Testing

Modified `test_grad.py` to step twice than once.

- Without the PR
Rank 3 Gradient test passed, while Ranks 0-2 all see wrong gradients.

- With the PR
```
Rank 1 Gradient test passed
Rank 2 Gradient test passed
Rank 0 Gradient test passed
Rank 3 Gradient test passed
```

### DIFF
--- a/pippy/PipelineStage.py
+++ b/pippy/PipelineStage.py
@@ -310,6 +310,19 @@ class PipelineStageBase(ABC):
         # Caching chunk outputs for final output merge or reduction
         self.output_chunks.clear()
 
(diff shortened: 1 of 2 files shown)


H-Huang (approved):
nice catch and fix!

PR #1071, opened 2024-04-18
Adjust logging level

kwen2501:
## Description

- "Upgrade" some logs in `PipelineStage` from info to debug;
- Print schedule type in `PipelineSchedule`.


H-Huang (approved):
👍

PR #1077, opened 2024-04-22
update loss utilities to take stage

H-Huang:
Hit some errors where the non-last stage is trying to get the loss or updating internal losses, so these changes fix that.

* #1079
* __->__ #1077

### DIFF
--- a/pippy/PipelineSchedule.py
+++ b/pippy/PipelineSchedule.py
@@ -44,9 +44,9 @@ class PipelineSchedule(ABC):
                 f"[{stage.stage_index}] Loss of microbatch {mb_index}: {loss}"
             )
 
-    def _maybe_get_loss(self, mb_index):
+    def _maybe_get_loss(self, stage, mb_index):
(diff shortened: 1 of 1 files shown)


pippy/PipelineSchedule.py
```diff
@@ -56,12 +56,17 @@ def _maybe_get_loss(self, mb_index):
...
-    def _update_losses(self, losses):
+    def _update_losses(self, stages, losses):
         """
         Update the losses to those in the internal state
         """
+        # if stages not a list turn into a list
+        if not isinstance(stages, list):
+            stages = [stages]
+        contains_last_stage = any([stage.is_last for stage in stages])
+
```
kwen2501, line 67: Because of this, maybe at least consider having two versions of this utility? One for single-stage schedules, the other for multi.

kwen2501 (approved):
Thanks for fixing the error.
The requirement for quite a few arguments is why I didn't pack those lines into a function :) 
But LGTM in general. Left one comment.

PR #1078, opened 2024-04-22
Test interleaved schedules

kwen2501:
## Description

- Added test_interleave.py
- Added to GPU CI.

## Status
- LoopBFS passes the test.
- Interleaved 1F1B fails with key error re micro-chunk id. (See CI result)

Cc: @H-Huang

### DIFF
--- a/test/test_interleave.py
+++ b/test/test_interleave.py
@@ -1,0 +1,168 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates
+import argparse
+import os
+import unittest
+
+import torch
+import torch.distributed as dist
+
+from pippy.IR import annotate_split_points, pipeline, SplitPoint
+from pippy.PipelineSchedule import ScheduleInterleaved1F1B, ScheduleLoopedBFS
+from pippy.PipelineStage import PipelineStage
+
+# Using same key words as single-stage tests for convenience in CI.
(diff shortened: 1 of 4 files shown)


test/test_interleave.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+    parser.add_argument(
+        "--chunks",
+        type=int,
+        default=4,
+    )
+    parser.add_argument(
+        "--schedule",
+        type=str,
+        default="dfs",
```
H-Huang, line 133: update default value

kwen2501: Good catch :)

test/test_interleave.py
```diff
@@ -0,0 +1,168 @@
...
+    parser.add_argument(
+        "--master_port", type=str, default=os.getenv("MASTER_PORT", "29500")
+    )
+    parser.add_argument(
+        "--cuda", type=int, default=int(torch.cuda.is_available())
+    )
+    parser.add_argument(
+        "--chunks",
+        type=int,
+        default=4,
```
H-Huang, line 128: added a fix to support this: https://github.com/pytorch/PiPPy/pull/1081

H-Huang (approved):
Thanks for adding the test!

PR #1079, opened 2024-04-22
re-enable interleaved 1f1b test

H-Huang:
* (to be filled)

### DIFF
--- a/test/test_pipeline_schedule.py
+++ b/test/test_pipeline_schedule.py
@@ -299,9 +299,6 @@ class TestPipelineSchedule(MultiProcessTestCase):
 
     @skip_if_lt_x_gpu(4)
     def test_interleaved_1f1b(self):
-        # TODO: not working
-        return
-
         device = torch.device(f"cuda:{self.rank}")
         dist.init_process_group(
             init_method=self.init_method,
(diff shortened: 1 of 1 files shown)


test/test_pipeline_schedule.py
```diff
@@ -320,38 +317,33 @@ def test_interleaved_1f1b(self):
...
+        ]
 
+        loss_fn = torch.nn.MSELoss()
         schedule = ScheduleInterleaved1F1B(
             stages,
             num_microbatches,
+            loss_fn=loss_fn,
         )
-        schedule.step_microbatches(microbatches)
+        schedule.step_microbatches(microbatches, target_mbs=target_mbs)
```
kwen2501, line 330: Should we feed target to every rank?

test/test_pipeline_schedule.py
```diff
@@ -320,38 +317,33 @@ def test_interleaved_1f1b(self):
...
         )
-        schedule.step_microbatches(microbatches)
-
-        # differing microbatch size
-        num_microbatches = 64
-        microbatches = [
-            torch.randn_like(microbatch) for _ in range(num_microbatches)
-        ]
-        schedule.step_microbatches(microbatches)
+        schedule.step_microbatches(microbatches, target_mbs=target_mbs)
```
kwen2501, line 346: Same here.

kwen2501 (commented):
Thanks for enabling the CI. Left one comment.

test/test_pipeline_schedule.py
H-Huang: we don't have to, but if we do that is also allright too since if it is not the last stage then it wont calculate the loss using the target_mbs.

kwen2501: I'd rather not rely on assumptions :) 
We should test the real case (strictly speaking even `microbatches` should not be fed to every rank).

PR #1081, opened 2024-04-22
fix interleaved 1f1b edge case

H-Huang:
There are some cases when either `num_microbatches` is small or `num_local_stages_per_rank` is large which could lead to the number of warmup steps >= `num_microbatches * num_local_stages_per_rank`. In this case the algorithm doesn't perform 1f1b and should just do warmups then cooldowns. 

This fix adds a `min` so it doesn't error. Perhaps we should also add a warning?

* __->__ #1081


kwen2501 (approved):
Thanks for the fix. LGTM.

PR #1082, opened 2024-04-22
Tidy up __init__

kwen2501:
## Description

- Remove stale APIs.
- Add in new APIs.
- Change all tests to directly import from pippy root.

### DIFF
--- a/test/test_bwd.py
+++ b/test/test_bwd.py
@@ -6,9 +6,13 @@ import unittest
 import torch
 import torch.distributed as dist
 
-from pippy.IR import pipe_split, pipeline
-from pippy.PipelineSchedule import Schedule1F1B, ScheduleGPipe
-from pippy.PipelineStage import PipelineStage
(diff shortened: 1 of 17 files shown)


test/test_bwd.py
```diff
@@ -6,9 +6,13 @@
 import torch
 import torch.distributed as dist
 
-from pippy.IR import pipe_split, pipeline
-from pippy.PipelineSchedule import Schedule1F1B, ScheduleGPipe
-from pippy.PipelineStage import PipelineStage
+from pippy import (
```
wconstab, line 9: does this cleanup align with how things will work after you move the tests into torch? will it just be s/pippy/torch.something.something/?

wconstab (approved):
lgtm, idk if howard has any particular preferences

test/test_bwd.py
kwen2501: Yes.

PR #1088, opened 2024-04-24
Fix interleaved 1f1b race

H-Huang:
Current fixes:
- We need to delay the send in the warmup phase **right before** 1f1b phase so that it is sent along with the recvs in the next step. This is to align the communication between ranks to prevent hang
- `Sorted_batch_isend_irecv()` is not working for interleaved 1f1b since we need to ensure that recv / send across different peers are all part of the same coalesced op. So we remove its usage from interleaved 1f1b.

------
Test added:

When I add sleep for certain ranks then there is a failure on rank 0 `[Rank 0] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data.`

Run with:
`PIPPY_VERBOSITY=DEBUG pytest test/test_pipeline_schedule.py -vsk test_interleaved_1f1b_with_model_sleep`

* __->__ #1088


kwen2501:
Thanks for the fix! I haven't run through the entire change but one small comment:
maybe try not to reuse the name `ops` too much. It could lead to bugs if other devs did not use it carefully.

pippy/PipelineSchedule.py
```diff
@@ -603,27 +603,22 @@ def step_microbatches(
...
-        # n_loop = n_stage / n_pp
-        # run microbatches in sequences of NPp
+        Operate on the microbatches for interleaved 1f1b schedule (https://arxiv.org/pdf/2104.04473.pdf).
 
-        schedule operates at the rank level
-
-        highest rank has a warmup (F only) count of [len(stages) - 1] * seq_size
-        each hop away from highest rank adds 2 warmup stages
+        Highest rank has a warmup (F only) count of [len(stages) - 1] * group_size
+        and each hop away from highest rank adds 2 warmup stages due to:
```
kwen2501, line 609: What does `group_size` mean here? Number of PP ranks? If so, is `[len(stages) - 1] * group_size` a square of # PP ranks?
And what does "hop" mean?

pippy/PipelineSchedule.py
```diff
@@ -603,27 +603,22 @@ def step_microbatches(
...
-        dist_from_highest = (worldsize - 1) - rank
-
-        total_steps = warmup_steps + (num_stages * num_microbatch)
 
-        Rank 0: 0F 0F 0F 0F 2F 2F 2F 2F
-        Rank 1:    1F 1F 1F 1F 3F3B 3F 3F 3F
+        TODO: Interleaved 1F1B does not support using sorted_batch_isend_irecv()
+        because it requires recvs and sends from different peers
+        to execute in the same coalesced operation. As a result, this schedule does
+        not support models with skip connections.
```
kwen2501, line 616: I like this comment. To actuate it, we can add a `has_skip_conn` filed in `pipe_info`, so that the schedule can check this field and error out here. (Rather than silently hang)

pippy/PipelineSchedule.py
```diff
@@ -669,42 +664,47 @@ def backward_stage_local_index(step):
         # Delay send waits
         sends_to_wait: List[dist.Work] = []
 
+        # Store ops (potentially across steps)
+        ops: List[dist.P2POp] = []
+
         for step in range(self.total_steps):
-            # warmup, forward only
+            # Warmup Phase (forward only)
             if step < warmup_steps:
```
kwen2501, line 672: nit: `for step in range(warmup_steps)`

pippy/PipelineSchedule.py
```diff
@@ -669,42 +664,47 @@ def backward_stage_local_index(step):
...
+                        sends_to_wait.append(work)
+                        ops.clear()
 
                     self._maybe_compute_loss(
                         fwd_stage, output, target_mbs, mb_index
                     )
 
-            # 1f1b
+            # 1F1B Phase (forward and backward)
             elif warmup_steps <= step < warmup_steps + fwd_bwd_steps:
```
kwen2501, line 707: nit: `for step in range(warmup_steps, warmup_steps + fwd_bwd_steps)`
nit: rename `fwd_bwd_steps` to `1f1b_steps`

pippy/PipelineSchedule.py
```diff
@@ -719,31 +719,30 @@ def backward_stage_local_index(step):
...
                     )
 
-                    # bwd
+                    # Backward
                     loss = self._maybe_get_loss(bwd_stage, bwd_mb_index)
                     bwd_stage.backward_one_chunk(loss=loss)
                     ops.extend(bwd_stage.get_bwd_send_ops())
 
-                    works = sorted_batch_isend_irecv(ops)
-                    sends_to_wait.extend(works.values())
```
kwen2501, line 744: I assume this firing get merged with the firing right above?

pippy/PipelineSchedule.py
```diff
@@ -719,31 +719,30 @@ def backward_stage_local_index(step):
...
                 with record_function(f"1F1B {step}"):
-                    ops = fwd_stage.get_fwd_recv_ops()
+                    ops.extend(fwd_stage.get_fwd_recv_ops())
                     ops.extend(bwd_stage.get_bwd_recv_ops())
-                    works = sorted_batch_isend_irecv(ops)
-                    for work in works.values():
+                    if ops:
+                        work = dist.batch_isend_irecv(ops).pop()
                         work.wait()
+                        sends_to_wait.append(work)
```
kwen2501, line 730: nit: if we are batching sends *and recvs* together, then this is no need to push `work` to `sends_to_wait`. The latter is for sends only.

pippy/PipelineSchedule.py
```diff
@@ -719,31 +719,30 @@ def backward_stage_local_index(step):
...
                     loss = self._maybe_get_loss(bwd_stage, bwd_mb_index)
                     bwd_stage.backward_one_chunk(loss=loss)
                     ops.extend(bwd_stage.get_bwd_send_ops())
 
-                    works = sorted_batch_isend_irecv(ops)
-                    sends_to_wait.extend(works.values())
-
-            # cooldown
+            # Cooldown Phase (backward only)
             else:
```
kwen2501, line 746: nit: `for step in range(...)`

test/test_pipeline_schedule.py
```diff
@@ -400,6 +433,41 @@ def test_interleaved_1f1b_negative(self):
...
+        )
+
+        model = ModelWithSleep(dim=4, hidden_dim=8, out_dim=4, rank=self.rank)
+        stages_per_rank = 2
+        num_microbatches_list = [4, 8, 16]
+        for num_microbatches in num_microbatches_list:
+            batch = torch.rand((num_microbatches, 4), device=device)
+            stages = self._create_virtual_pipeline_stages(
+                model,
+                torch.rand((1, 4)).to("meta"),
```
kwen2501, line 453: nit: try to use variable than number.

test/test_pipeline_schedule.py
```diff
@@ -400,6 +433,41 @@ def test_interleaved_1f1b_negative(self):
...
+                num_microbatches=num_microbatches,
+            )
+
+            schedule = ScheduleInterleaved1F1B(
+                stages, num_microbatches, loss_fn=nn.MSELoss()
+            )
+            if self.rank == 0:
+                schedule.step(batch)
+            elif self.rank == self.world_size - 1:
+                target = torch.rand((num_microbatches,), device=device)
```
kwen2501, line 465: Not sure why the 2nd dim of `target` is 1 rather than 4.

kwen2501 (approved):
Nice work.

pippy/PipelineSchedule.py
H-Huang: Yep `group_size` is pp ranks. And the warmup steps is going to be a multiple of the pp ranks. Hop means for every rank less than the highest. The comment is a bit old so i will update it.

pippy/PipelineSchedule.py
H-Huang: yes it gets executed on the next step

pippy/PipelineSchedule.py
H-Huang: good point!

pippy/PipelineSchedule.py
kwen2501: oh, so `len(stages)` means the number of **local** stages?

PR #1091, opened 2024-04-25
Clean up microbatch.py and add unit test

kwen2501:
- Added comment to each function
- Added typing
- Moved default chunking setting
- Added test


H-Huang (approved):
LGTM

PR #1093, opened 2024-04-25
Make some modules private

kwen2501:
This is a PyTorch requirement. Otherwise, the module must be documented in a .rst file.

Making modules private by prefixing an `_`, as suggested by PyTorch doc checker.

Privatized modules: 
`_utils`, `_debug`, `_backward`, `_unflatten`


H-Huang (approved):
cool cool

PR #1095, opened 2024-04-30
pipeline() API accepting split_spec

kwen2501:
The PR allows the pipeline() API to directly accept a `split_spec` (a dict) rather than asking user to `annotate_split_point()` first.

Reason:
1. Making it a one-stop experience.
2. "Downplaying" the concept of _annotation_ by hiding it from the user.
3. The `pipeline()` API today accepts `split_policy`, so it would be best to put `split_spec` at the same place, making it an "either-or".

For examples of the new UI, please see the code change to example.py

### DIFF
--- a/examples/basic/example.py
+++ b/examples/basic/example.py
@@ -4,7 +4,7 @@
 
 import os
 import torch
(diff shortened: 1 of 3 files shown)


wconstab (approved):
Like it.

H-Huang (approved):
Nice!

PR #1099, opened 2024-05-01
Make IR private

kwen2501:
User-facing APIs in IR.py have been lifted into `__init__.py`. Thus it seems to make sense to make IR.py private.


test/local_test_c10d_ddp.py
```diff
@@ -7,7 +7,7 @@
 
 import torch
 import torch.distributed as dist
-from pippy.IR import pipe_split
+from pippy._IR import pipe_split
```
H-Huang, line 10: nit: just pippy?

H-Huang (approved):
Looks good! I assume this will break any users using pippy.IR right now, so should there be an error message to switch? Or does it not matter since we are migrating.

wconstab:
fine with me. although, depending on if we care about tip of PipPy breaking we could just do the rename in the pytorch side.  i dont think this matters though. once we land it then there is no point of keeping PipPy repo/pkg maintained

PR #1101, opened 2024-05-02
Use non-strict mode by default

kwen2501:
`torch.export` has strict mode and non-strict mode. 
For difference, please read [Non-Strict Export](https://pytorch.org/docs/stable/export.html#non-strict-export).

This PR switches to non-strict mode by default. Improving tracing success rate (no Dynamo graph break).

### DIFF
--- a/pippy/_IR.py
+++ b/pippy/_IR.py
@@ -1142,6 +1142,7 @@ class Pipe(QualnameMapMixin, torch.nn.Module):
             mod,
             example_args,
             example_kwargs,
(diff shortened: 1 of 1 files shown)


H-Huang (approved):

wconstab:
im not convinced.

i should be easy to convince since at the moment i'm fighting with tracer issue after tracer issue.

however, if the answer is "turn off safety so it doesn't throw errors" i'm not sure i want to use the tracer at all.

PR #1109, opened 2024-05-03
refactor manual stage, include docs and example

H-Huang:
Add doc string for manual stage and example under `basic/`

Made input_args a required argument

* __->__ #1109

### DIFF
--- a/examples/basic/example_manual_stage.py
+++ b/examples/basic/example_manual_stage.py
@@ -1,0 +1,119 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates
+# Minimal effort to run this code:
+# $ torchrun --nproc-per-node 3 example_manual_stage.py
+
+import os
+import torch
+from pippy import ScheduleGPipe, ManualPipelineStage
+
+in_dim = 512
+layer_dims = [512, 1024, 256]
+out_dim = 10
+
+# Single layer definition
+class MyNetworkBlock(torch.nn.Module):
+    def __init__(self, in_dim, out_dim):
+        super().__init__()
+        self.lin = torch.nn.Linear(in_dim, out_dim)
+
+    def forward(self, x):
+        x = self.lin(x)
+        x = torch.relu(x)
+        return x
+
+
+# Model chunk definition
+class ModelChunk0(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.layer0 = MyNetworkBlock(in_dim, layer_dims[0])
+
+    def forward(self, x):
+        return self.layer0(x)
+
+class ModelChunk1(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.layer1 = MyNetworkBlock(layer_dims[0], layer_dims[1])
+
+    def forward(self, x):
+        return self.layer1(x)
+
+class ModelChunk2(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.layer2 = MyNetworkBlock(layer_dims[1], layer_dims[2])
+        # Final output layer (with OUT_DIM projection classes)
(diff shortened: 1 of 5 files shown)


pippy/ManualPipelineStage.py
```diff
@@ -172,36 +176,55 @@ def get_stage_shapes(
...
+    This class extends the `PipelineStageBase` class and can similarly be used
+    in `PipelineScheule`.
+    Args:
+        submodule (nn.Module): The PyTorch module wrapped by this stage.
+        stage_index (int): The ID of this stage.
+        num_stages (int): The total number of stages.
+        device (torch.device): The device where this stage is located.
+        num_microbatches (int): The number of microbatches to use.
+        input_args (Union[torch.Tensor, List[torch.tensor]], optional): The input arguments for the module.
+        output_args (Union[torch.Tensor, List[torch.tensor]], optional): The output arguments for the module.
```
kwen2501, line 192: > The input arguments for the module

What does "the module" refer to? 

Earlier, the doc uses a term "submodule":
> submodule (nn.Module): The PyTorch module wrapped by this stage.

examples/basic/example_manual_stage.py
```diff
@@ -0,0 +1,119 @@
...
+# the rank of this process, and the device.
+if rank in rank_model_and_input:
+    model, example_input = rank_model_and_input[rank]
+    stage = ManualPipelineStage(
+        model,
+        rank,
+        world_size,
+        device,
+        chunks,
+        example_input,
```
wconstab, line 100: i'd like to see a test case where rank1 does not pass input shapes, and we use the logic in `get_stage_shapes` to propagate this information from rank0 (including dtype).  IMO its too hard to manually specify shape+dtype for all stages.

pippy/ManualPipelineStage.py
```diff
@@ -172,36 +176,55 @@ def get_stage_shapes(
...
+        group (dist.ProcessGroup, optional): The process group for distributed training. If None, default group.
```
wconstab, line 193: i think its tricky to let group be optional, and have the num_stages vs group_size be ambigious.

(in my case I was passing num_stages =2 and i was forgetting to pass my pp_mesh.group, so ManualStage was complaining that the default group's size wasn't compatible with my num_stages.  Should we just always require group, or is it useful to not require it?

pippy/ManualPipelineStage.py
H-Huang: oops, meant "submodule," thanks!

examples/basic/example_manual_stage.py
wconstab: Not a question for this PR, but where did we land on shape inference on first forward?

i think the only argument against this is that it wont be easy to make the UX the same for manual/tracer stages if tracer needs a shape at trace time while manual stage does not.

but it is definitely nicer UX to not require a shape up front, and could even be more flexible if the lazy thing allowed reallocating buffers for larger shapes (in case of dynamic shape)

question RE tracer ux- @kwen2501- how mature is export+dynamicshape? can we mark the input shapes for tracing as dynamic somehow and at least ensure the trace is capable of being used for different number of batches or seqlen?

pippy/ManualPipelineStage.py
wconstab: we should support devicemesh too.

examples/basic/example_manual_stage.py
H-Huang: I think we still want shape inference. Eventually we can make input_args an optional argument and perform shape inference for the user, but for now we require it for initialization.

kwen2501 (approved):
LGTM. Thanks for the improvement!

PR #1139, opened 2024-08-20
Update all hf examples to have dist.barrier

muellerzr:
Without having `dist.barrier()`, all of the HF examples wind up hanging since we're destroying the pg before all comms have completed in these small examples, leading to a hang. This PR adds `dist.barrier()` just before `dist.destroy_process_group()` to fix this.

### DIFF
--- a/examples/huggingface/pippy_bert.py
+++ b/examples/huggingface/pippy_bert.py
@@ -73,6 +73,7 @@ def run(args):
     else:
         out = schedule.step()
 
+    dist.barrier()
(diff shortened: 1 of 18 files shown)


kwen2501 (approved):
LGTM. Thanks!

Pasting our triage here for future reference:
If the batch is small, and the two ranks have a gap in terms of when they are launched, then rank 0 could have finished sending the activations before rank 1 bootstraps.
(In NCCL, sending a small message does not need handshake between the two ranks)

